PR #33, opened 2016-09-09
Add in-game plugin settings

jonpas:
- Add in-game settings (using CBA Settings) for what used to be settings in TeamSpeak plugin.
- Remove unused `acre_sys_core_ACRE_USE_DEBUG_EXTENSIONS` variable.
- Change `USE_DEBUG_EXTENSIONS` to be simply defined or not defined (no value).
- Log successful loading of `acre2.ini` plugin configuration file (many times it simply reported failed on first try but then still loaded it).
- Fix `acre2.ini` location (no idea how or why this broke, might just be something wrong on my end but it kept generating `acreacre2.ini` outside of the folder, folder was created empty and plugin failed to load it).

The only caveat with this is that the settings can be forced on server and in mission, I'll speak to CBA guys about adding a condition to that. It shouldn't be a problem either, no reason anyone would ever force those settings.


NouberNou:
Wait, isn't this the same as #36?

NouberNou:
OH Nevermind, this is for the plugins.

NouberNou:
This should also be tested in a dev build ASAP.

Sniperhid (approved):
LGTM, agree with Nou!

PR #35, opened 2016-09-09
API addon function headers

Sniperhid:
- Provide accurate function headers for all sqf functions in the API addon.


jonpas:
:+1:

PR #36, opened 2016-09-11
Convert all suitable settings to CBA Settings

jonpas:
- Convert difficulty and basic mission settings to CBA settings.
- Add `ACRE_DEPRECATED` macro.
- Deprecate and hide "Difficulty Settings" and "Basic Mission Setup" modules (still usable, just not visible if not already placed, to be removed in 2.4.0).
- Update "Basic Mission Setup" module to call new helper functions (for seamless "run once" handling).
- Add `acre_api_fnc_getAllRadios` function (with caching) that returns all radios in `CfgAcreRadios` in format `[ ["class1", "class2",], ["name1", "name"] ]` (can easily be expanded for more information.
- Move CBA settings to API component.
- Remove unused `disableSignalLoss` function.

**Requires #33** due to CBA settings preparation being there!

"Name Channels" module was not converted as it's designed so multiple of them can be placed to configure each side. That can be moved to Eden attributes one day when we have a nice framework for that maybe.


NouberNou:
This sounds like it is moving along. Keep me posted on testing.

jonpas:
This has nothing to do with C++ side, the commits show up because it's based on #33. Not sure if this is viable yet, will decide once this is fully tested.

addons/api/CfgModules.hpp
```diff
@@ -9,9 +9,9 @@ class CfgVehicles
 {
     class Logic;
     class Module_F: Logic { class ArgumentsBaseUnits { class Units; }; class ModuleDescription { class AnyBrain; }; };
-    
+
     class GVAR(basicMissionSetup) : Module_F {
-        scope = 2;
+        scope = 1;
```
Sniperhid, line 14: I would disagree with hiding the modules for now. As they are pretty intuitive for most mission makers

addons/api/CfgModules.hpp
```diff
@@ -69,9 +69,9 @@ class CfgVehicles
             sync[] = {};
         };
     };
-    
+
     class GVAR(DifficultySettings) : Module_F {
-        scope = 2;
+        scope = 1;
```
Sniperhid, line 74: Same thing again with hiding the module

Sniperhid (changes requested):
Please don't scope away the modules for now.

NouberNou:
What is the decision here moving forward with modules? Should we get a dev build out soon and testing these things?

Also backwards compatibility is big, so we gotta make sure that this isn't going to break old missions that use or rely on modules for core functionality.

PabstMirror:
CBA 3.1 will have some new events that were designed to help with backwards compatibility for ace modules/settings which will be going through the same transition problems.

Ref https://github.com/CBATeam/CBA_A3/pull/437

Modules should be able to add event handler for `CBA_beforeSettingsInitialized`
And inside that call event `CBA_settings_setSettingMission` with `[_setting, _value, _true]`
Actual 3den cba settings will be loaded first and take priority

Sniperhid:
From what I understand Jonpas is planning to do some further experimentation here, Modules that effect the mission settings would a nice method.

jonpas:
Yes, no point discussing anything until I do more work on that.

jonpas:
I will extract all fixes and improvements from this PR into a separate one.

jonpas:
Replaced by #53, we won't be moving those things to CBA settings, but rather Eden attributes when the time comes.

PR #38, opened 2016-09-14
Fix: Prevent 117F Squelch menu from looping

Sniperhid:
- If 'TONE' or 'NONE' squelch settings were selected the AN/PRC-117F interface would enter a loop that you could not exit unless you did not use these settings.
- Minor macro optimization


jonpas (approved):
Macro changes look good.

PR #39, opened 2016-09-16
Optimise: getGear for large item counts

Sniperhid:
- Improve acre_lib_fnc_getGear. Only returns ACRE radios. This improves performance for many of the frequently run gear checks. Particularly for high item counts. This should prove gains particularly when running ACE due to all the new items.
- Local check for uninitialised radio worst case performance significantly lowered (38 to 2 milliseconds for 400 items). This was causing noticable stutter for medics carrying large amounts of medical supplies
- Server radio ID monitor runs faster particularly when doing a full search. I created a test mission with ~10 equipment crates, ~10 vehicles and several squads of AI. With the getGear improvement it ran twice as fast (50ms v 24ms) with 150 vanilla units (running ACE). I later implemented the optimisation for ItemCargo. Further reducing runtime from 24 ms to 20ms.
- Downside all acre radios should begin with acre*. Several other solutions were tested but this led to the best performance.

### DIFF
--- a/addons/game/XEH_pre_init.sqf
+++ b/addons/game/XEH_pre_init.sqf
@@ -240,6 +240,9 @@ FUNC(getGear) = {
     if (isNull _unit) exitWith {[]};
     // diag_log text format["Assigned Items: %1", (assignedItems _unit)];
     private _gear = (weapons _unit) + (items _unit) + (assignedItems _unit);
+    _gear = _gear select {(_x select [0, 4]) == "ACRE" || _x == "ItemRadio" || _x == "ItemRadioAcreFlagged"}; // We are only interested in ACRE gear.
+    // The below is really slow and tends to worsen performance.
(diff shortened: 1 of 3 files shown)


addons/game/XEH_pre_init.sqf
```diff
@@ -240,6 +240,9 @@ FUNC(getGear) = {
     if (isNull _unit) exitWith {[]};
     // diag_log text format["Assigned Items: %1", (assignedItems _unit)];
     private _gear = (weapons _unit) + (items _unit) + (assignedItems _unit);
+    _gear = _gear select {(_x select [0, 4]) == "ACRE" || _x == "ItemRadio" || _x == "ItemRadioAcreFlagged"}; // We are only interested in ACRE gear.
```
NouberNou, line 243: A substring check is faster than a config check?

Sniperhid: Indeed, even without lazy eval here. I didn't lazy eval to improve worst case performance (lazy eval adds some overhead). Though the monitorRadios function (client check for uninitalized radios and a few other things) would do 3 config lookups for each equipment return. But filtering in getGear seems to speed it up. Each in the server masterIDTracker there is a speed up and that's only 1 config look up.

jonpas (approved):
LGTM

PR #40, opened 2016-09-17
Improve replaceGear/addGear to work with equipment overflow

Sniperhid:
- Prevent ACRE from running out of Radio IDs in cases were mission makers overflow containers.
- This helps makes ACRE more robust and essentially means ACRE will always be able to give radio regardless of circumstance.
- Ensure the desync check is run even if there is no room for the test desync item.


NouberNou (commented):
This looks good. Has it been tested much?

PabstMirror:
I believe this will fix the minor issue I saw when placing a vanilla CSAT UAV operator (`O_soldier_UAV_F`)
Uniform and vest are so full it can't add converted radio and would throw

```
ACRE Warning: Radio ID acre_prc343_id_1 was returned for a non-existent baseclass (ACRE_PRC343) in inventory, possibly removed by a gear script while requesting ID: ["arifle_Katiba_ACO_pointer_F","hgun_Rook40_F","FirstAidKit","ItemMap","ItemCompass","ItemWatch","O_UavTerminal","NVGoggles_OPFOR"]
```

Sniperhid:
Tested a bit in SP and with some of the commands in MP. Should be good IMO.

PR #43, opened 2016-09-18
Remove VCRedist extras files copying

jonpas:
- Title

Has been decided we won't package this anymore, used to be some sort of forward compatibility due to using pre-released VC.

Wiki already updated as well.


Sniperhid (approved):
LGTM

PR #44, opened 2016-09-18
Show missing files as errors and don't copy unwanted DLLs

jonpas:
- Show missing files as errors during building and sum of those missing files at the end.
- Don't auto-copy all `*.dll` files in root, list them as important files rather and copy plugins same way as extras.
- Remove unnecessary `amountOfBuildsFailed` variable, read length of the names and shorten `namesOfBuildsFailed`.
- Fix new lines at the end.


Sniperhid (approved):
Lgtm

PR #45, opened 2016-09-18
Increment build number automatically

jonpas:
- Add arguments to `make.py` for incrementing any part of the version.
- Fix double call to `get_project_version` (was diplaying "ACRE2 VERSION set to <version>" twice).

`increment_build`, `increment_patch`, `increment_minor`, `increment_major`
Will add to wiki once merged.

Perfect for CI as we can simply designate that based on branch and our plan for workflow.

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -604,7 +604,7 @@ def addon_restore(modulePath):
     return True
 
 
-def get_project_version(version_update):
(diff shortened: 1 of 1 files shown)


NouberNou (commented):
Looks good.

NouberNou (approved):

PR #48, opened 2016-09-19
STATE Macro harmonization

Soldia1138:
- Harmonize STATE Macros for radios
- Put STATE Macros to main component


addons/sys_prc117f/menus/fnc_changeMenu.sqf
```diff
@@ -22,11 +22,11 @@ params["_newMenu"];
 
 
 // clear menu data
-SET_STATE("menuSelection" , 0);
+SET_STATE(menuSelection , 0);
```
Sniperhid: Due to the way the macros works this becomes some thing like `["menuSelection "`

addons/sys_prc152/menus/fnc_changeMenu.sqf
```diff
@@ -22,15 +22,15 @@ params["_newMenu"];
 
 /*
 if(!([_newMenu] call FUNC(verifyIsMenu)) ) then {
-    _newMenu = GET_STATE_DEF("currentHome", GVAR(VULOSHOME));
+    _newMenu = GET_STATE_DEF(currentHome, GVAR(VULOSHOME));
 };
 */
 // clear menu data
-SET_STATE("menuSelection" , 0);
+SET_STATE(menuSelection , 0);
```
Sniperhid: Same issue again -> `"menuSelection "`

Sniperhid (changes requested):

addons/sys_prc152/menus/fnc_changeMenu.sqf
Soldia1138: Should be fixed by https://github.com/IDI-Systems/acre2/pull/48/commits/50c0de446a85dfa0d8d5fc99457ec22a2b752e47

Sniperhid (changes requested):
You have fixed the issues however I discored a new issue with the 152/117F menus

```
private _saveName = MENU_SELECTION_VARIABLE(_menu);
SET_STATE(_saveName, _value);
```

is commonly used previously the value of `_saveName` would be used as the variable name e.g. `pgm_preset_number` now instead they are soted on `"_saveName"` instead

Soldia1138:
According to my tests, this is working now.

Sniperhid (approved):

Sniperhid:
Seconded my testing seems good as well! GJ JP

PR #53, opened 2016-09-29
Cleanup modules

jonpas:
- Implement things from #36
- Support debug mode recompilation in `CfgFunctions` of `api` component.
- Add `acre_api_fnc_getAllRadios` function (with caching) that returns all radios in `CfgAcreRadios` in format `[ ["class1", "class2",], ["name1", "name"] ]` (can easily be expanded for more information).
- Remove unused `disableSignalLoss` function.
- Fix basic mission setup not exiting if already ran.
- Cleanup all module function code (remove unneeded checks, `waitUntil`, `spawn`, use direct array definition rather than slow `set`, use `QGVAR()` instead of `QUOTE(GVAR())`).
- Use `PATHTO_FNC` and `RECOMPILE` from CBA 3.1.0 instead of `DEFINE_FUNCTION`.
- Move module functions to root (so we don't redefine `PATHTO_FNC` for special place again).
- De-hard-code(?) radios for preset definitions (only 152, 148 and 117F were supported before).
- Add `isNil` check to `acre_sys_data_fnc_getPresetData` in case of radios without presets.
- Remove lots of redundant white space apparently.


addons/api/fnc_getAllRadios.sqf
```diff
@@ -0,0 +1,39 @@
...
+{
+    private _isAcre = getNumber (_x >> "isAcre"); // All radio classes have that
+    private _name = getText (_x >> "name"); // Only non-base classes have that
+
+    // Has name and isAcre, assume valid radio class
+    if (_isAcre == 1 && {_name != ""}) then {
+        _classes pushBack (configName _x);
+        _names pushBack _name;
+    };
+} forEach ("true" configClasses (configFile >> "CfgAcreComponents"));
```
Sniperhid, line 37: @NouberNou - Is this a valid approach with CfgAcreComponents to find radio base classnames?
Just wondering if with future plans for the component system this is acceptable.

Sniperhid (approved):
LGTM - Would be nice to wait for Nou to confirm that the CfgAcreComponents approach for finding base radios is appropriate though.

PR #63, opened 2016-10-04
API: prevent setRevealToAI erroring on non-dedicated.

Sniperhid:
- Fix the setRevealToAI from erroring on dedicated clients.


addons/api/fnc_setRevealToAI.sqf
```diff
@@ -19,6 +20,8 @@ INFO_2("%1 called with: %2",QFUNC(setRevealToAI),_this);
 
 params ["_var"];
 
+if (!hasInterface) exitWith {false};
```
jonpas, line 23: I'd say print an warning, but I am not sure what's our standard about it.

jonpas (approved):
Looks good.

addons/api/fnc_setRevealToAI.sqf
Sniperhid: So far it's mostly been doing the same thing just silently exit if run on server/headless client.

jonpas: Sounds good then, add that to coding guidelines? :) (something along the lines of "in case of API calls on wrong machines, silently exit")

PR #68, opened 2016-10-08
Fixes the too agressive muting on briefing stage

Soldia1138:
- The muting on briefing screen happened as soon as a player clicked on continue. This PR fixes that, so players are still able to talk globally


Sniperhid (approved):
LGTM

PR #69, opened 2016-10-08
Always log synchronously

jonpas:
- Make log always be logged synchronously


Soldia1138 (approved):
Looks fine for me

PR #72, opened 2016-10-10
Fix API Macros

Sniperhid:
- Fix the MACROs in the API functions


jonpas (approved):
It was not me, Git blame/history lies if it tells you it was me.

PR #75, opened 2016-10-11
Fix mismatched TS plugin and mod versions error formatting

jonpas:
- Use correct macro that supports new-lines for on-screen error message.

### DIFF
--- a/addons/main/script_mod.hpp
+++ b/addons/main/script_mod.hpp
@@ -11,7 +11,7 @@
 
 // MINIMAL required version for the Mod. Components can specify others..
 #define REQUIRED_VERSION 1.64
-#define REQUIRED_CBA_VERSION {3,1,0}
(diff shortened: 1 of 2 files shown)


Sniperhid:
Is there a way for a mod to specify a required CBA version? Might be nice to throw that in too if it exists.

Sniperhid (approved):
LGTM, see my comment though

thojkooi:
@Sniperhid https://github.com/IDI-Systems/acre2/blob/master/addons/main/script_mod.hpp#L14

and it's used here: https://github.com/IDI-Systems/acre2/blob/master/addons/main/CfgSettings.hpp

jonpas:
Yeah, I already answered him on Slack. I should start getting used to this GitHub being more public. 😆

jonpas:
CBA 3.1.1 is actually not needed as we don't have anything else we want to display in big orange box at the moment.

PR #80, opened 2016-10-19
Improve mismatched versions text to be clearer

jonpas:
- "Mod" is not very descriptive as per @commy2
- It's actually written as "Client" in the info log when no mismatch is found, must have overlooked it.


Sniperhid:
As this is shown to the user it should be really clear imo. Replacing the term client with your should make it understandable to everyone imo.

```
Plugin -> Teamspeak plugin version:
Mod/Client -> Your mod version:
Server -> The server version:
```

jonpas:
```
TeamSpeak plugin version: %1
Your version: %2
Server version: %3
```

I like this as it shows only on clients anyways, 👍 @Sniperhid .

Sniperhid (approved):
Lgtm

PR #81, opened 2016-10-21
Reset sub-version counters with increment argument in make.py

jonpas:
- Make version increment arguments reset version parts below that, except build
- Add another way of retrieving Git commit ID if no Git client could be found (from ACE3's version)
- Fix possible undefined variable error

Wiki updated accordingly.


Sniperhid (approved):
LGTM!

PR #90, opened 2016-11-04
sys_components update

Sniperhid:
- Improve file headers - Not all are complete but some documentation is better than none!
- Fixes getAllAvailableConnectors - In case of a detached component


Sniperhid:
Superseeded by #95

PR #96, opened 2016-11-13
Add CBA setting for desync hint display

Soldia1138:
- Allow the player to disable the GearDesync Hint
- Most communities are aware of the issues with gearDesyncs
- Some players want to capture videos and don't care about gear desync
- If hint is disabled, it is written to rpt. So we can still debug it on upcoming issues

### DIFF
--- a/addons/sys_core/initSettings.sqf
+++ b/addons/sys_core/initSettings.sqf
@@ -29,7 +29,7 @@
 ] call CBA_Settings_fnc_init;
 
 [
-    QGVAR(disableGearDesync),
+    QGVAR(disableDesyncHint),
     "CHECKBOX",
     "Disable Gear Desync Hint",
     "ACRE2",
--- a/addons/sys_core/initSettings.sqf
+++ b/addons/sys_core/initSettings.sqf
@@ -33,7 +33,7 @@
     "CHECKBOX",
     "Disable Gear Desync Hint",
     "ACRE2",
-    false,
+    true,
     false,
     {EGVAR(sys_radio,disableDesyncHint) = _this}
 ] call CBA_Settings_fnc_init;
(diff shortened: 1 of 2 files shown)


addons/sys_core/initSettings.sqf
```diff
@@ -27,3 +27,13 @@
     false,
     {["disableUnmuteClients", _this] call FUNC(setPluginSetting)}
 ] call CBA_Settings_fnc_init;
+
+[
+    QGVAR(disableGearDesync),
```
jonpas: I'd keep that as `QGVAR(disableDesyncHint)`, don't want people thinking it could in any way disable actual desync.

addons/sys_radio/fnc_monitorRadios.sqf
```diff
@@ -195,7 +195,12 @@ DFUNC(hasGearDesync) = {
         {
             _message = _message + format ["%1\n", name _x];
         } forEach ACRE_SERVER_DESYNCED_PLAYERS;
-        hintSilent _message;
+        if(GVAR(disableDesyncHint)) then {
+            diag_log _message;
```
jonpas: As said on Slack, this should always be printed to RPT, even if it's hinted. And also using a macro. So this entire thing would become:

```
WARNING(_message);
if (!GVAR(disableDesyncHint)) then {
    hintSilent _message;
};
```

jonpas (changes requested):

jonpas:
I honestly thing we should just get rid of the hint entirely. Logging it to RPT is enough, as it's not something we should worry about if nothing actually goes wrong.

Sniperhid:
IIRC, when this occurs new radios will no longer be initialised? This typically effects JIP/respawns but in the case of a one-life mission where everyone is spawned at the beginning this isn't a big deal. I might be wrong as I only encountered this once in the last 18 months.

jonpas (approved):

Sniperhid (approved):
I am willing to accept this as a temporary solution.

PR #97, opened 2016-11-16
Remove double log in sys_io function ping

jonpas:
- Title

I blame @Sniperhid but it was probably merged at the same time.


Sniperhid (commented):
LGTM

Sniperhid (approved):

PR #103, opened 2016-11-25
Fix edit me in documentation results in 404

thojkooi:
URL uses wrong branch


jonpas:
Already done in #102

I feel bad for you @Glowbal !

PR #106, opened 2016-11-28
Moving XEH_preInit functions to separate files

TheMagnetar:
- Move functions defined in XEH_preInit to separate function files. This is still WIP (see coment below regarding edge cases)

- [x] sys_components
- [x] sys_core
- [x] sys_data
- [x] sys_gui
- [x] sys_server
- [x] sys_signalmap


TheMagnetar:
Edge cases. What to do with them?
- EFUNC functions in main/XEH_preInit.sqf: *EFUNC(lib,getGear)*, *EFUNC(lib,replaceGear)*, *EFUNC(lib,removeGear)*, *EFUNC(lib,addGear)*, *EFUNC(lib,getCompartment)*, *EFUNC(lib,fastHashCreate)*, *EFUNC(lib,fastHashCopyArray)*, *EFUNC(lib,fastHashCopy)* and *EFUNC(lib,fastHashKeys)*
- *_monitorFnc* in sys_core/XEH_preInit.sqf
- *DFUNC(_hashSerialize)*, *DFUNC(_arraySerialize)*, *DFUNC(_hashDeserialize)* and *DFUNC(_arrayDeserialize)* in sys_data/XEH_preInit.sqf
- Empty functions in sys_prc152 *DFUNC(onKnobMouseEnter)* and *DFUNC(onKnobMouseExit)*.
- *FUNC(handleEndTransmission)* in sys_signal/XEH_preInit.sqf

Soldia1138:
imho we have three options here:
- Leave it as it is
- Move them all in one .hpp and simply include that
- Move them to different files and alternate the component for those functions

I'll vote for the second option. Third one would be nice but is somehow hacky.

Edit:
Fourth option for the lib functions:
We could introduce a new PREP macro like `PREPLIB` like `PREPMAIN` but resulting in lib functions.

jonpas:
I believe we decided to move them and simply rename them to not be `lib`, but whatever component they should be in. Something that is required by all other, `sys_core` maybe?

jonpas:
Contains merge conflicts due to code formatting merges.

TheMagnetar:
Resolved merge conflicts. Before removing WIP, should I also add in here this corner cases we discussed before or do you prefer having them in another PR?

jonpas:
Separate PR would be better, easier to review.

addons/sys_data/fnc_deserialize.sqf
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Example:
+ * [ARGUMENTS] call acre_sys_data_fnc_serialize;
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private ["_ret"];
+if (IS_SERIALIZEDHASH(_this)) then {
```
Sniperhid: This macro is undefined. The definition was previously in `sys_data/XEH_preInit.sqf`
Definition should be
`#define IS_SERIALIZEDHASH(array) (IS_ARRAY(array) && {(count array) > 0} && {IS_STRING((array select 0))} && {(array select 0) == "ACRE_HASH"})`

Alternative is to move into the `sys_data/script_component.hpp` for use throughout the addon.

addons/sys_data/script_component.hpp
```diff
@@ -14,4 +14,6 @@
     #define DEBUG_SETTINGS DEBUG_SETTINGS_SYS_DATA
 #endif
 
+#define IS_SERIALIZEDHASH(array) (IS_ARRAY(array) && {(count array) > 0} && {IS_STRING((array select 0))} && {(array select 0) == "ACRE_HASH"})
```
jonpas, line 17: This whole thing could be improved like:
```
#define IS_SERIALIZEDHASH(array) (!(array isEqualTo []) && {(array select 0) isEqualTo "ACRE_HASH"}) 
```

But that can be a separate PR.

Sniperhid: That still doesn't guarantee the variable  `array` is of the type array?
Also `isEqualTo` will be case-sensitive (unsure if it matters in this case)
So there would certainly be cases where you suggested optimisation is not a good idea.

jonpas: `arrayIsEqualTo []` will return `false` if it's not empty and if it's not array, so it guarantees it is array. Yes it will be case-sensitive, unsure if it matters, can just add `toLower` if it does.

I am not sure where all this is used, that's why I said separate PR. It's very minor optimization if any at all anyways.

Sniperhid (approved):
Tested locally - Looks good

Issue #83, opened 2016-10-25
Teamspeak 3.1 Support


Sniperhid:
Teamspeak 3.1 will make some interesting changes to the way plugins are handled. They will now be stored in the local user's appdata folder. One benefit is that copying of the plugins will no longer require admin priviledges.
- [x] Update teamspeak plugin API version -> #170 
- [x] Update the steam/auto-copy dll to use the new plugin locations -> #114 

The teamspeak plugin location can no longer be read from the registry. There is also now a feature for syncing teamspeak plugins across various machines but I don't think this will be useful for ACRE2. For example (win 10) - `C:\Users\<USER>\AppData\Roaming\TS3Client\plugins`

PR #114, opened 2016-11-30
Plugin auto-copy overhaul

Sniperhid:
- Removes the config file and instead directly compares the dlls
- Adds Roaming application data folder to scan list - In a manner compatible with the TS 3.1 usage of the registry (Partially addressing #83)
- Skips searching non-existent folders - Should registry/users not be updated. It will however create the plugins folder if it does not exist.
- Prefer AppData over Program Files due to its lack of requiring admin privileges. As teamspeak prefers program files over AppData in plugin load order precedence. This will remove the plugin from Program Files if copying to AppData

### DIFF
--- a/extensions/src/ACRE2Steam/CallExt_DllMain.cpp
+++ b/extensions/src/ACRE2Steam/CallExt_DllMain.cpp
@@ -227,7 +227,7 @@ std::string ReadRegValue64(HKEY root, std::string key, std::string name)
     return value;
 }
 
-void write_config(std::string version, int skip) {
+/*void write_config(std::string version, int skip) {
     std::string mod_folder = find_mod_folder();
     std::string config_path = mod_folder + "acresteamconfig";
     std::ofstream config_file(config_path);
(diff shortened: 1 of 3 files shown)


extensions/src/ACRE2Steam/CallExt_DllMain.cpp
```diff
@@ -238,8 +238,39 @@ void write_config(std::string version, int skip) {
...
+    sizeB = streamB.seekg(0, std::ifstream::end).tellg();
+    streamB.seekg(0, std::ifstream::beg);
 
+    if (sizeA != sizeB) return false;
+
+    static const size_t BLOCKSIZE = 4096*512; // 2MB good expected size for the teamspeak dlls
+    size_t remaining = sizeA;
+
+    while (remaining)
+    {
```
jonpas: Opening brace on same line as per coding style guidelines.

jonpas (changes requested):
I know it's WIP, but [snake formatting](http://acre2.idi-systems.com/wiki/development/coding-guidelines-cpp#variable-names) as per coding style guidelines.

Sniperhid:
@jonpas - addressed seeking updated review. Tested this locally myself and seems to work well

jonpas (approved):

jonpas:
Looks good code wise, however it's still not using snake formatting (I clicked approve before checking that). I think we should also get rid of the old code.

jonpas (approved):
Awesome work @Sniperhid !

PR #116, opened 2016-11-30
Code formatting c++ files

TheMagnetar:
- Change the following entries (done using grep command) according to coding guidelines:
  - `if(` to `if (`
  - `for(` to `for (`
  - `while(` to `while (`
  - `#if(` to `#if (`
  - `#elif(` to `#elif (`
  - `#if(` to `#if (`

This does not include functions that end with `*_for(` like `std::this_thread::sleep_for(` or `*_if`. If requested those can also be added.


jonpas (approved):

jonpas:
Tested in MP, works fine.

Sniperhid (approved):
Tested with Jonpas

PR #127, opened 2016-12-05
Correct underwater check for spectator - use camera position rather than unit position

Sniperhid:
- Correct underwater check for spectator - use camera position rather than `acre_player` position


Sniperhid:
Regardless this is proper place to do the check as this is where the actual position for the listener is calculated. The check itself is relatively inexpensive.

Soldia1138 (approved):

Issue #126, opened 2016-12-04
SEM52SL - Invisible on ultra-wide setups.


Sniperhid:
**ACRE2 Version:** `2.0.3.926` (stable / dev + commit hash)

**Description:**
- SEM52SL background does not display on ultra-wide resolutions

**Steps to reproduce:**
- Use resolution 5860x1080 and bring up the SEM52SL interface

```
5860x1080 - SafeZone stuff
safezonew = 7.27078
safezoneh = 1.81818
safezoney = -0.409091

// Extract from config
y=SafeZoneY + SafeZoneH - NEW_SCALE * SafeZoneH + 2/8 * safeZoneW;

// Below sum is greater than the screen height (safezoneh)....
2/8 * safeZoneW = 1.817695
```

PR #131, opened 2016-12-10
SEM52SL - Update all dialog positions to be based off screen height

Sniperhid:
- Fixes the Dialog to work for all aspect ratios not just 16:9
- Addresses #126


jonpas (approved):
Changes numbers, must work.

PR #133, opened 2016-12-11
Add CBA setting for spectator volume

PabstMirror:
- Adds CBA setting to change sepectator volume
- also added stringtable support for other cba settings

Players in spectator talking can overpower in-game/alive players.
Adds a variable settings instead of only being able to mute.


addons/sys_core/initSettings.sqf
```diff
@@ -11,17 +11,27 @@
...
 ] call CBA_Settings_fnc_init;
 
 [
+    "ACRE_SPECTATOR_VOLUME",
+    "SLIDER",
+    localize LSTRING(ACRE_SPECTATOR_VOLUME_displayName),
+    "ACRE2",
+    [0, 1, 1, 2],
+    false,
+    {["premixGlobalVolume", _this] call FUNC(setPluginSetting)}
```
jonpas: This will overwrite the actual premix volume and will clash with the  above settings when sent to TS.

jonpas (changes requested):
That won't work with the TS plugin. Not sure if forgot to change the line or what?

addons/sys_core/initSettings.sqf
Sniperhid: In this case the fix should be just replacing this line with `{}`. The function `FUNC(setPluginSetting)` is for sending values to the teamspeak plugin for settings relevant to the plugin. But in the case of `ACRE_SPECTATOR_VOLUME` it only needs setting in SQF land and will also have almost immediate effect on spectator volume 👍

Sniperhid (changes requested):

jonpas (approved):

Sniperhid (approved):
Awesome thanks for the PR @PabstMirror !

PR #153, opened 2016-12-19
German localization

Dorbedo:
- Add Localization into German


Soldia1138 (approved):
Danke für's übersetzen
Thx for translating

PR #155, opened 2016-12-19
Cleanup unacknowledged IDs after a time-out period

Sniperhid:
- When a unique radio ID is created it reverses an ID this becomes an unacknowledged ID until the client that requested the unique ID acknowledges it. One quirk was that if a client disconnects mid-process the ID would be reserved forever. This PR will now free-up the unacknowledged IDs after 60 seconds.

### DIFF
--- a/addons/sys_server/fnc_acknowledgeId.sqf
+++ b/addons/sys_server/fnc_acknowledgeId.sqf
@@ -1,22 +1,23 @@
 /*
  * Author: ACRE2Team
- * SHORT DESCRIPTION
+ * Acknowledges that a radio of with the ID has been created.
  *
  * Arguments:
- * 0: ARGUMENT ONE <TYPE>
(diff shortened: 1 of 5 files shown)


addons/sys_server/fnc_acknowledgeId.sqf
```diff
@@ -1,34 +1,37 @@
...
+ * 0: Radio class name <STRING>
+ * 1: Player <OBJECT>
  *
  * Return Value:
- * RETURN VALUE <TYPE>
+ * None
  *
  * Example:
- * [ARGUMENTS] call acre_COMPONENT_fnc_FUNCTIONNAME
+ * ["acre_prc343_id_1", acre_player] call acre_sys_server_fnc_acknowledgeId;
```
jonpas: No `;`

jonpas (approved):
Looks good apart from the comment! 👍

PR #159, opened 2016-12-20
ace_interact - german translation

Dorbedo:
- adds German to the ace_interact component
- modifies sys_core for consistency

### DIFF
--- a/addons/ace_interact/stringtable.xml
+++ b/addons/ace_interact/stringtable.xml
@@ -3,45 +3,59 @@
     <Package name="ace_interact">
         <Key ID="STR_ACRE_ace_interact_radios">
             <English>Radios</English>
+            <German>Funkgeräte</German>
         </Key>
         <Key ID="STR_ACRE_ace_interact_channelShort">
             <English>%1 Chn: %2</English>
+            <German>%1 Kanal: %2</German>
         </Key>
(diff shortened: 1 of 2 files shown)


addons/ace_interact/stringtable.xml
```diff
@@ -3,45 +3,59 @@
...
             <English>Set To Right Ear</English>
+            <German>Auf rechten Ohrhörer legen</German>
         </Key>
         <Key ID="STR_ACRE_ace_interact_setAsActive">
             <English>Set as Active</English>
+            <German>Als aktiv setzen</German>
         </Key>
         <Key ID="STR_ACRE_ace_interact_bindMultiPushToTalk">
             <English>Bind Multi Push To Talk</English>
+            <German>Sprechtasten festlegen</German>
```
Soldia1138, line 50: In the translation for the control adjustment menu, (M)PTT is translated as "Funkgerätetaste". I think we should be consistent here and I'd prefer "Sprechtaste". (#153)

It is your call if you want to change it yourself or let me do it.

Soldia1138: PS: It is ok for me if you add the changes to this PR. We can modify the PR commit text then.

Soldia1138 (changes requested):
This one looking good. Just waiting for a harmonization with #153

Soldia1138 (approved):

PR #160, opened 2016-12-21
CBA settings ACRE 2 difficulty

TheMagnetar:
- Add the following CBA settings:
  - Interference
  - Signal loss
  - Ignore antena direction
  - Full duplex
  - Reveal to AI
- CBA settings will be ignored if the module is placed and no settings are changed at the briefing screen. Cancel and ok buttons do nothing as long as no values have been changed.
- Changing other ACRE 2 settings except difficulty settings will result in module difficulty settings taking preference.
- CBA settings will be applied if changes are made to the difficulty settings during the briefing screen.
- Settings during init.sqf take precedence over CBA settings set during the briefing screen and module settings.
- Added localisation for the introduced difficulty settings.
- Show a warning that the difficulty module is deprecated in favor of CBA settings
- Hide difficulty module

**Note**: Not tested on dedicated.

**Question**: I tried to make it work through initPost eventhandlers but to no avail. Any reason why?


addons/sys_core/XEH_postInit.sqf
```diff
@@ -151,4 +150,37 @@ if (getClientStateNumber < 10) then { // Check before game has started (in brief
...
 };
 
+///////////////////////////////////
+//
+// CBA SETTINGS
+// Applies the difficulty module settings over CBA settings. This allows the mission editor to bypass cba server settings.
+// If the module is not present, this function has no effect.
+//////////////////////////////////
+{
+    private _ignoreAntennaDirection = ((allMissionObjects "acre_api_DifficultySettings") select 0) getVariable ["IgnoreAntennaDirection", nil];
```
Sniperhid: I would suggest storing the result of -> `(allMissionObjects "acre_api_DifficultySettings")` as it's a relatively expensive command, so doing it just once would be great. Also would be good to handle the case where there aren't any modules!

addons/sys_core/stringtable.xml
```diff
@@ -126,5 +126,25 @@
...
+        <Key ID="STR_ACRE_sys_core_difficultyInterference_displayName">
+            <English>Model Interference?</English>
+            <Spanish>Modelar interferencias?</Spanish>
+        </Key>
+        <Key ID="STR_ACRE_sys_core_difficultyFullDuplex_displayName">
+            <English>Full Duplex radios?</English>
+            <Spanish>Radios Dúplex?</Spanish>
+        </Key>
+        <Key ID="STR_ACRE_sys_core_difficultyAntennaDirection_displayName">
+            <English>Ignore Antena Direction?</English>
```
Sniperhid: `antena` -> `antenna`

addons/sys_core/stringtable.xml
```diff
@@ -126,5 +126,25 @@
...
+            <Spanish>Ignorar dirección de antena?</Spanish>
+        </Key>
+        <Key ID="STR_ACRE_sys_core_difficultyTerrainLoss_displayName">
+            <English>Terrain Loss Coefficient</English>
+            <Spanish>Coefficiente de pérdida de señal por terreno</Spanish>
+        </Key>
+        <Key ID="STR_ACRE_sys_core_difficultyRevealToAI_displayName">
+            <English>Reveal to AI?</English>
```
Sniperhid: Perhaps we should re-word this to `Speaking reveals players to AI?`

Sniperhid (changes requested):

TheMagnetar:
Changes done.

> Also would be good to handle the case where there aren't any modules!

I will update this to handle lack of module more gracefully.

addons/sys_core/stringtable.xml
kymckay: Suggestion: `AI Voice Detection` (or equivalent)

There's already context that this setting is related to ACRE due to the `"ACRE2"` category and there's no need to phrase the title as a question (thought maybe that's just my personal preference) since it's self evident that it's a setting - if more context was needed then I'd suggest using the optional inclusion of a tooltip.

addons/sys_core/stringtable.xml
```diff
@@ -126,5 +126,25 @@
             <Japanese>ヘッドセットのトグル/スペクテイターのトグル</Japanese>
             <Italian>Attiva/Disattiva Cuffia - Attiva/Disattiva Spettatori</Italian>
         </Key>
+        <Key ID="STR_ACRE_sys_core_difficultyInterference_displayName">
+            <English>Model Interference?</English>
```
jonpas: No question marks in setting names.

jonpas (changes requested):
I am not sure I really like this. I like my [old PR](https://github.com/IDI-Systems/acre2/pull/36) better, however not everything from there is suitable.

Either way, we should also deprecate the modules as done in that PR.

addons/sys_core/stringtable.xml
TheMagnetar: About deprecating the modules. I would be against it since most/some mission makers use them to quickly configure things, personalise the mission even further and bypass CBA settings.
I know the same can be done from init.sqf but, well, not everyone may do that. Just my 2 cents.

jonpas: CBA settings are accessible from within Eden and are saved in the mission, there should be no down-side to it at all.

TheMagnetar: true 👍

TheMagnetar:
Changes with:
- Translation updates (Spanish, English).
- Hidden difficulty module.
- Deprecated difficulty module.
- Optimisation when handling lack of difficulty module and all _allMissionObjects_

addons/sys_core/XEH_postInit.sqf
```diff
@@ -151,4 +150,31 @@ if (getClientStateNumber < 10) then { // Check before game has started (in brief
...
+
+    if (count _missionModules == 0) exitWith {};
+
+    private _ignoreAntennaDirection = (_missionModules select 0) getVariable ["IgnoreAntennaDirection", false];
+    private _fullDuplex = (_missionModules select 0) getVariable ["FullDuplex", false];
+    private _interference = (_missionModules select 0) getVariable ["Interference", true];
+    private _signalLoss = (_missionModules select 0) getVariable ["SignalLoss", true];
+
+    if (_signalLoss) then {
+        [1.0] call acre_api_fnc_setLossModelScale;
```
jonpas, line 170: Can just pass `1` and `0` for easier readability, but doesn't matter in SQF.

addons/sys_core/initSettings.sqf
```diff
@@ -47,3 +47,59 @@
     false,
     {EGVAR(sys_radio,disableDesyncHint) = _this}
 ] call CBA_Settings_fnc_init;
+
+// Difficulty settings
+// Interference
+[
+    "ACRE_INTERFERENCE",
```
jonpas: Those variables should probably be macroized to be consistent with others, they are not public API. Probably best if they stick to being `acre_sys_core_<name>`.

addons/sys_core/stringtable.xml
```diff
@@ -126,5 +126,25 @@
             <Japanese>ヘッドセットのトグル/スペクテイターのトグル</Japanese>
             <Italian>Attiva/Disattiva Cuffia - Attiva/Disattiva Spettatori</Italian>
         </Key>
+        <Key ID="STR_ACRE_sys_core_difficultyInterference_displayName">
```
jonpas: I think `difficulty` is not needed here.

TheMagnetar:
Changed to QGVAR, including ACRE_SPECTATOR_VOLUME to QGVAR(spectatorVolume)

addons/sys_core/stringtable.xml
jonpas: Yeah, I meant for all of them. :)

addons/sys_core/initSettings.sqf
```diff
@@ -19,9 +19,9 @@
 ] call CBA_Settings_fnc_init;
 
 [
-    "ACRE_SPECTATOR_VOLUME",
+    "QGVAR(spectatorVolume)",
```
Sniperhid: `"QGVAR(spectatorVolume)"` should be `QGVAR(spectatorVolume)`

Sniperhid (changes requested):

addons/sys_core/initSettings.sqf
TheMagnetar: (facepalm)  I did not upload all changes.

addons/sys_core/initSettings.sqf
```diff
@@ -47,3 +47,59 @@
     false,
     {EGVAR(sys_radio,disableDesyncHint) = _this}
 ] call CBA_Settings_fnc_init;
+
+// Difficulty settings
+// Interference
+[
+    QGVAR(interference),
+    "CHECKBOX",
+    localize LSTRING(Interference_displayName),
```
jonpas: Lowercase first letter inside `LSTRING`, all from here to bottom.

jonpas (changes requested):

Sniperhid (approved):

jonpas (approved):

PR #165, opened 2017-01-02
Change example documentation for company signals setup

Jsmuk:
As per a brief discussion on Slack, the call given in the `mission_setupCompanySignals.sqf` example doesn't seem to actually work. Adding `preprocessFileLineNumbers` seems to fix it. I honestly have no idea why it makes a difference but @jonpas suggested it, and then hinted I should submit a PR after it worked.. 

- Change the example call in `mission_setupCompanySignals.sqf` to use `preprocessFileLineNumbers`

### DIFF
--- a/extras/examples/mission_setupCompanySignals.sqf
+++ b/extras/examples/mission_setupCompanySignals.sqf
@@ -1,7 +1,7 @@
 /*
  * Sample of setting up RTO's for a company signals setup.
  *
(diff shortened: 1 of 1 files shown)


Sniperhid (approved):
good point!

jonpas:
`preprocessFileLineNumbers` preprocesses the file into a string that can be used with `compile`.

PR #166, opened 2017-01-02
Add CBA settings on dedicated server

PabstMirror:
Should have been done before, 
but now it's critical for the new difficulty settings which are global


jonpas (approved):

jonpas:
Nice catch!

PR #176, opened 2017-01-21
Fix renamed settings variables

jonpas:
- Fix regression from #160 
- Improve API functions in question

Somewhat my fault for not noticing that, as I requested the rename of those internal variables.


Sniperhid (approved):
Nice one!

PR #179, opened 2017-01-26
More optimization

Sniperhid:
- Parameter formatters for calling extensions should be faster.
- `HASH_KEYS` macro is now faster
- Various improvements using `apply` and `select`. Including several serializers and deserializers are now faster.
- Adds functions for caching frequently used config lookups `acre_sys_radio_fnc_isBaseClassRadio` and `acre_sys_radio_fnc_isUniqueRadio`.

### DIFF
--- a/addons/sys_core/fnc_getSpokenLanguages.sqf
+++ b/addons/sys_core/fnc_getSpokenLanguages.sqf
@@ -15,4 +15,4 @@
  */
 #include "script_component.hpp"
 
-ACRE_SPOKEN_LANGUAGES apply {((GVAR(languages) select _x) select 0);};
+ACRE_SPOKEN_LANGUAGES apply {(GVAR(languages) select _x) select 0};
--- a/addons/sys_core/fnc_getSpokenLanguages.sqf
+++ b/addons/sys_core/fnc_getSpokenLanguages.sqf
@@ -15,9 +15,4 @@
  */
 #include "script_component.hpp"
 
-private _list = [];
-
-{
-    _list pushBack ((GVAR(languages) select _x) select 0);
-} forEach ACRE_SPOKEN_LANGUAGES;
-_list
+ACRE_SPOKEN_LANGUAGES apply {((GVAR(languages) select _x) select 0);};
--- a/addons/sys_core/fnc_setSpokenLanguages.sqf
+++ b/addons/sys_core/fnc_setSpokenLanguages.sqf
@@ -16,11 +16,8 @@
 #include "script_component.hpp"
 
 private _list = _this;
-private _languageIds = [];
+private _languageIds = _list apply {([_x] call FUNC(getLanguageId))};
 
-{
-    _languageIds pushBack ([_x] call FUNC(getLanguageId));
-} forEach _list;
 ACRE_SPOKEN_LANGUAGES = _languageIds;
 [_list select 0] call FUNC(setSpeakingLanguage);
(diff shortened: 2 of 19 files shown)


addons/sys_core/fnc_getSpokenLanguages.sqf
```diff
@@ -15,9 +15,4 @@
  */
 #include "script_component.hpp"
 
-private _list = [];
-
-{
-    _list pushBack ((GVAR(languages) select _x) select 0);
-} forEach ACRE_SPOKEN_LANGUAGES;
-_list
+ACRE_SPOKEN_LANGUAGES apply {((GVAR(languages) select _x) select 0);};
```
jonpas: No need for extra parenthesis and semi-colon.
`ACRE_SPOKEN_LANGUAGES apply {(GVAR(languages) select _x) select 0};`

addons/sys_core/fnc_setSpokenLanguages.sqf
```diff
@@ -16,11 +16,8 @@
 #include "script_component.hpp"
 
 private _list = _this;
-private _languageIds = [];
+private _languageIds = _list apply {([_x] call FUNC(getLanguageId))};
```
jonpas: What do you have with parenthesis! :P
`private _languageIds = _list apply {[_x] call FUNC(getLanguageId)};`

addons/sys_data/XEH_preInit.sqf
```diff
@@ -38,39 +38,37 @@ GVAR(validStates) = HASH_CREATE;
...
         if (!isNil "_val") then {
-            _keys pushBack _x;
             if (IS_ARRAY(_val)) then {
                 _val = _val call FUNC(_arraySerialize);
             };
             if (IS_HASH(_val)) then {
                 _val = _val call FUNC(_hashSerialize);
             };
             _vals pushBack _val;
+            true;
```
jonpas: Return value? No semi-colon.

addons/sys_radio/fnc_isBaseClassRadio.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * Example:
+ * ["acre_prc152_id_1"] call acre_sys_radio_fnc_isBaseClassRadio
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _isBaseClassRadio = false;
+
+if (HASH_HASKEY(GVAR(radioBaseClassCache),_this)) then {
```
jonpas: `if-exitwith` is faster than `if-then-else` in this case, as it will only cache it on first call and then just needs the value.

addons/sys_server/fnc_onGetRadioId.sqf
```diff
@@ -18,7 +18,7 @@
 
 params ["_player", "_class", "_callback", ["_replacementId",""]];
 
-if (getNumber (configFile >> "CfgWeapons" >> _class >> "acre_hasUnique") == 0) then {
+if (!(_class call EFUNC(sys_radio,isBaseClassRadio))) then {
```
jonpas: `if !(_class call EFUNC(sys_radio,isBaseClassRadio)) then {` makes it easier to read IMO.

jonpas (changes requested):
Some possible improvements.

jonpas (approved):

PR #191, opened 2017-01-28
Removed all  traces of NO_CLIENT macro

TheMagnetar:
- Removes all traces of existence of the NO_CLIENT macro.
- No RPT errors detected.

### DIFF
--- a/addons/sys_data/XEH_postInitServer.sqf
+++ b/addons/sys_data/XEH_postInitServer.sqf
@@ -1,7 +1,5 @@
 #include "script_component.hpp"
 
-if (!isServer) exitWith {};
-
 LOG("ENTER");
 
 ACRE_DATA_SYNCED = true;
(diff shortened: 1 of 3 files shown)


addons/sys_server/XEH_preInitServer.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 
-NO_CLIENT;
+if (!isServer) exitWith {};
```
jonpas: This is redundant, this preInit is already called via `serverInit` in XEH.

addons/sys_data/XEH_postInitServer.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 
-NO_CLIENT;
+if (!isServer) exitWith {};
```
jonpas: This is redundant, this postInit is already called via `serverInit` in XEH.

Mind, that is not the same for `clientInit` combined with `!hasInterface` due to headless clients which are clients without an interface and most of the time don't need either. However `serverInit` excludes headless clients as well.

jonpas (changes requested):
Those are both redundant, I guess I missed them when I converted things to `clientInit`/`serverInit`.

jonpas (approved):

Issue #206, opened 2017-02-06
Proper indentation and spacing in all config files


Soldia1138:
- [x] Opening brace on same line
- [x] `QUOTE(DFUNC( ))` -> `QFUNC( )`
- [x] Correct spaces (_something = something_)
- [x] Correct indentation
- [x] `CLASS:` without space between class and semi-colon.
- [x] `QUOTE(PATHTOF( ))` -> `QPATHTOF( )`

TheMagnetar:
The `QFUNC()` change I did it without noticing in the #209 alongside different macro conversions, including some changes to use `EFUNC()`, `QEFUNC()`, `GVAR()`, `EGVAR()` and `QEGVAR()`.

Soldia1138:
Think this could be closed. > 99% should be fixed in master branch now

PR #209, opened 2017-02-07
Conversion to macro

TheMagnetar:
- Conversion form `call acre_*` to `FUNC`/`EFUNC`
- Conversion from "acre_* variables to `GVAR`, `EGVAR` and `QEGVAR`.
- Conversion from `QUOTE(DFUNC())`  to `QFUNC/QEFUNC` (#206).
- Minor code formatting.
- Tested in local server with ACE3 for interaction menu changes.


jonpas (approved):
Looks good.

We should probably get rid of those macros https://github.com/IDI-Systems/acre2/blob/master/addons/main/script_macros.hpp#L81-L82 if this PR fully removes the use of them.
Or make an issue to remove them, as some other PR probably still uses them.
-> #215

Sniperhid (approved):

PR #212, opened 2017-02-13
Added missing Czech translation of sys_core

Kllrt:
Just some Czech translation which Fanatic missed :^)

### DIFF
--- a/addons/sys_core/stringtable.xml
+++ b/addons/sys_core/stringtable.xml
@@ -187,6 +187,7 @@
             <Japanese>モデル干渉</Japanese>
             <French>Modèle d'interférence</French>
             <German>Interferenz Modul</German>
+            <Czech>Interference radiových vln</Czech>
         </Key>
         <Key ID="STR_ACRE_sys_core_fullDuplex_displayName">
             <English>Full Duplex radios</English>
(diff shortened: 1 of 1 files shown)


Sniperhid (approved):

addons/sys_core/stringtable.xml
```diff
@@ -187,34 +187,39 @@
...
             <English>AI Voice Detection</English>
             <Spanish>Detección del habla por parte de la IA</Spanish>
             <Japanese>AI による声の検知</Japanese>
             <French>Détection de la voix par l'IA</French>
             <German>Sprachwahrnehmung durch KI</German>
+            <Czech>Detekce hlasu hráčů AIčky</Czech>
         </Key>
     </Package>
-</Project>
+</Project>
```
jonpas: Missing newline.

jonpas (changes requested):

addons/sys_core/stringtable.xml
Kllrt: That's weird, I didn't change that. Could it be Tabler issue?

jonpas: Right.

jonpas (approved):

PR #223, opened 2017-02-21
Channel switching functionality

tbeswick96:
- Add the ability for clients to be automatically moved to a channel in teamspeak containing the arma server name
- If no channel containing the server name can be found, it will look for a default channel "ACRE" instead
- If the default channel does not exist, it will not move the client
- Moving is configured via a CBA setting, so can be forced per mission or server

So far I've tested this on local MP with a teamspeak I have full admin control over. We have a channel called "ACRE" and I added one called "ACRE - Test". I launched a server via eden, and named it "Test". The plugin successfully moved me to "ACRE - Test" after pressing "continue" in the map screen. 
Then I did the same test as above but called the arma server "Test this should not work", whereby I was moved to the "ACRE" channel.
Disabling the plugin in teamspeak moved me back to my original channel. Re-enabling it moved me back to correct ACRE channel. Likewise closing teamspeak and re-launching it moved me back to the correct ACRE channel almost straight after connecting. Stopping the arma server caused me to move back to the original channel, as did hard exiting the game (Alt-F4).

What needs doing:
- [ ] I will build this version and push to my unit's repo on Sunday, where we can then fully test the functionality and resolve any issues. I would recommend waiting until after then before doing any merges to master.
- [ ] Translation for cba setting text

Current issues:
- Currently this does not work with channels that have passwords (in the majority case where you don't have permission to bypass passwords). I think this could be resolved when/if CBA adds text as an option for settings. Until then though I can't think of a good workaround. (It is impossible to retrieve a channel password as a client in a teamspeak extension)
- It searches for the entire server name in a channel. This could be quite ugly in teamspeak. Would like some input here as to what might be the best solution to avoid having the entire server name in the channel. Potentially just a keyword followed by an id in the server name? Not sure...


Bosenator:
Awesome stuff, really appreciated. This is going to help with our public server and moving 40+ individuals each week on our operation server.

I have a question in regards to using the server name, does it need to be an exact match? I know some server names are usually quite long to advertise their unit website, as well as their usage of mods. I'm unsure as if there's a character limit on TeamSpeak Channel names, or a discrepancy in allowed character sets. 

Is there not a way this can be set via the description.ext?

tbeswick96:
@Bosenator yeah the server -> channel name needs a bit of thought for the best way to do it that works with what acre already has in it. With a move to cba settings over modules and description/userconfig (I believe this is the case, someone correct me if I'm wrong), I'm hesitant to add modules etc in case they get removed or whatever.
I think the preferred approach would be simply a cba setting for channel name and password that can be set per mission, however cba settings does not currently support text, so that's not possible right now.
I need some input from the acre devs/maintainers as to what they think the best approach for this would be.

addons/sys_core/stringtable.xml
```diff
@@ -41,6 +41,16 @@
             <Italian>Disattiva l'Audio dei Clients</Italian>
             <Korean>클라이언트 음소거 해제</Korean>
         </Key>
+        <Key ID="STR_ACRE_sys_core_channelSwitch_displayName">
+            <English>Channel Switch</English>
+            <Spanish>HELP</Spanish>
```
jonpas, line 46: Just remove those translations. (Same principle as ACE3.)

addons/sys_server/fnc_channelCheck.sqf
```diff
@@ -0,0 +1,23 @@
+/*
+ * Author: Tim Beswick
+ * Checks if main display is visible and sets server name, triggering channel move
```
jonpas, line 3: Missing final dot.
Add a note about it being a PFH check.

addons/sys_server/fnc_channelCheck.sqf
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Arguments:
+ * None
```
jonpas, line 6: That's not true ^^.

jonpas (changes requested):
C++ indentation should be 4 spaces as well (tab is currently used).

extensions/src/ACRE2Shared/_CONSTANTS.h
```diff
@@ -15,7 +15,8 @@
 #undef _DEBUG_DIALOG
 
 
-#define PIPE_TIMEOUT 10000
+//#define PIPE_TIMEOUT 10000
+#define PIPE_TIMEOUT 5000
```
jonpas, line 19: Reason?

jonpas:
As for settings, I'd say keep it to as it currently is. If CBA Settings add different options some day we can always add that functionality. A current nice addition would probably be to make it work with partial server names as channel names as well (which it already does if I read PR description correctly). There is also the default "ACRE" channel you can use as stated in the PR description.

tbeswick96:
I am quite retarded so didn't reset the master I branched this off and I'm not skilled enough to fix it, so I've create a new pull request with the same details. Sorry :3

jonpas:
Replaced by #225

PR #224, opened 2017-02-21
Infantry Phone

TheMagnetar:
- Allows a unit to use the vehicle intercom externally.
- A unit with the "Infantry Telephone" can give it to another player.
- The crew members  (not passengers) of a vehicle with infantry telephone can make it "beep" to request the phone being picked.
- Allowed (default) distance is 10m, hull distance (if custom position is set) is 1.5m.
- Currently this ~is available for vehicles of the class Wheeled_APC_F,  tanks and MRAP_02_base_F.~ can be enabled for classes (and its children) `Tank` and `Car_F`.
- **Move intercom config values to root vehicle config** (backwards compatibility until 2.7.0)
- Requires ACE interaction.
- Needs testing in dedicated.

Todo:
- [x] Disable debug (including `systemChat` in `sys_core`)
- [x] Set custom position for all vanilla vehicles that have an infantry phone
- [x] Revert default custom position distance to 1.5m (~currently investigating interaction menu clipping issue~ was just checking wrong positions)
- [x] Move to `sys_intercom`
- [x] Move config values to root vehicle config
- [x] Use #239


addons/sys_core/fnc_vehicleCrewPFH.sqf
```diff
@@ -16,15 +17,51 @@
 #include "script_component.hpp"
 
 private _vehicle = vehicle acre_player;
+private "_vehicleIntercom";
```
Sniperhid: This line shouldn't be neccesary as `_vehicleIntercom` is never used in this scope.

addons/sys_core/fnc_updateExternalIntercomStatus.sqf
```diff
@@ -0,0 +1,45 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_vehicle", "_intercomUnit", "_action", "_givingUnit"];
+
+switch (_action) do {
+    case 0: {
+        // Stop using the intercom externally
+        [_vehicle, ["intercomUnit", nil, true]] remoteExecCall ["setVariable", _vehicle];
+        _intercomUnit setVariable ["vehicleIntercom", nil, true];
```
Sniperhid: `_intercomUnit setVariable ["vehicleIntercom", nil, true];` is already global so there is no need to use `remoteExecCall` I think using `setVariable` with the global flag (3rd parameter as true) instead of `remoteExecCall` is probably better. Do this for all instances in this function.

addons/ace_interact/stringtable.xml
```diff
@@ -141,5 +141,25 @@
...
+            <Spanish>Intercom</Spanish>
+            <German>Intercom</German>
+        </Key>
+        <Key ID="STR_ACRE_ace_interact_giveIntercom">
+            <English>Give intercom</English>
+            <Spanish>Dar intercom</Spanish>
+            <German>Intercom geben</German>
+        </Key>
+        <Key ID="STR_ACRE_ace_interact_takeIntercom">
+            <English>Take intercom</English>
```
Sniperhid: Perhaps `Access vehicle intercom` is better?

addons/ace_interact/stringtable.xml
```diff
@@ -141,5 +141,25 @@
...
+            <Spanish>Coger intercom</Spanish>
+            <German>Intercom nehmen</German>
+        </Key>
+        <Key ID="STR_ACRE_ace_interact_returnIntercom">
+            <English>Return intercom</English>
```
Sniperhid: Perhaps `leave vehicle intercom` is better?

Sniperhid (dismissed):
Would like to see this feature called the "Infantry Phone" and as such actions and variables renamed to match for clarity.

Also a new config definition for it would be great and by default only enable for Tanks and IFVs, something like:
``getNumber (configFile >> "CfgVehicles" >> typeOf (_vehicle) >> "ACRE" >> "CVC" >> "hasInfantryPhone");``

docs/wiki/user/direct-speech.md
```diff
@@ -29,3 +29,7 @@ Voice occlusion will occur if there is direct path for another player's voice to
 Vehicle attenuation is the system where people outside and inside of a vehicle may hear each other at reduced volumes. This is particularly noticeable in armoured vehicles but generally isn't in effect in open seats. This is can be different for each seat in the vehicle. Turning out of the vehicle will also reduce this effect. Generally firing from vehicles positions, helicopter door gunners) will have this effect weakened too. This should generally work well with most vehicles from mods.
 
 For technical details for addon makers see [vehicle attenuation configuration](api_vehicle-attenuation).
+
+## Vehicle intercom and infantry telephones
```
jonpas: This doesn't feel right in this document. 🤔

TheMagnetar:
Ready for merging. All comments from @Sniperhid were taken care off.

jonpas (approved):

PR #225, opened 2017-02-21
Teamspeak channel switching functionality

tbeswick96:
- Add the ability for clients to be automatically moved to a channel in teamspeak containing the arma server name
- If no channel containing the server name can be found, it will look for a default channel "ACRE" instead
- If the default channel does not exist, it will not move the client
- Moving is configured via a CBA setting, so can be forced per mission or server

So far I've tested this on local MP with a teamspeak I have full admin control over. We have a channel called "ACRE" and I added one called "ACRE - Test". I launched a server via eden, and named it "Test". The plugin successfully moved me to "ACRE - Test" after pressing "continue" in the map screen. 
Then I did the same test as above but called the arma server "Test this should not work", whereby I was moved to the "ACRE" channel.
Disabling the plugin in teamspeak moved me back to my original channel. Re-enabling it moved me back to correct ACRE channel. Likewise closing teamspeak and re-launching it moved me back to the correct ACRE channel almost straight after connecting. Stopping the arma server caused me to move back to the original channel, as did hard exiting the game (Alt-F4).

I will build this version and push to my unit's repo on Sunday, where we can then fully test the functionality and resolve any issues. I would recommend waiting until after then before doing any merges to master.

Current issues:
- Currently this does not work with channels that have passwords (in the majority case where you don't have permission to bypass passwords). I think this could be resolved when/if CBA adds text as an option for settings. Until then though I can't think of a good workaround. (It is impossible to retrieve a channel password as a client in a teamspeak extension)
~- It searches for the entire server name in a channel. This could be quite ugly in teamspeak. Would like some input here as to what might be the best solution to avoid having the entire server name in the channel. Potentially just a keyword followed by an id in the server name? Not sure...~


tbeswick96:
Comment from @Bosenator:
>Awesome stuff, really appreciated. This is going to help with our public server and moving 40+ individuals each week on our operation server.
I have a question in regards to using the server name, does it need to be an exact match? I know some server names are usually quite long to advertise their unit website, as well as their usage of mods. I'm unsure as if there's a character limit on TeamSpeak Channel names, or a discrepancy in allowed character sets.
Is there not a way this can be set via the description.ext?

tbeswick96:
@Bosenator yeah the server -> channel name needs a bit of thought for the best way to do it that works with what acre already has in it. With a move to cba settings over modules and description/userconfig (I believe this is the case, someone correct me if I'm wrong), I'm hesitant to add modules etc in case they get removed or whatever.
I think the preferred approach would be simply a cba setting for channel name and password that can be set per mission, however cba settings does not currently support text, so that's not possible right now.
I need some input from the acre devs/maintainers as to what they think the best approach for this would be.

tbeswick96:
Comment from @jonpas 
>As for settings, I'd say keep it to as it currently is. If CBA Settings add different options some day we can always add that functionality. A current nice addition would probably be to make it work with partial server names as channel names as well (which it already does if I read PR description correctly). There is also the default "ACRE" channel you can use as stated in the PR description.

Sniperhid (changes requested):
This largely looks well coded and like it should work.

- My main criticism with the code is using the term `channelCheck` this should be `teamspeakChannelCheck` to be clear (and avoid confusion with radio channels). There a large amount of instances where this should be corrected. Particularly in what an end-user reads "Channel Switch" in CBA settings it doesn't provide enough information.
- The `channelCheck` function should be moved to `sys_io`. As the `sys_io` component handles communication with TS not `sys_server`.

jonpas (changes requested):
Still using tabs instead of spaces in C++ code.

tbeswick96:
Changes:
- Renamed instances of channel predominantly with TeamspeakChannel or TSChannel, to avoid confusion (both in code and for end-user)
- Moved the TS channel check function to sys_io
- Improved the searching for channels matching server name. It will now look for a channel containing the server name, and if none found, it will look for a channel where the server name contains that channel name. As before, if no channel is found, it will look for the default channel "ACRE", but only as a direct match (So a channel called "ACRE1" would not be classed as default).
- To avoid the case where it repeatedly looks for a valid channel but none can be found, it will only try to switch to a dedicated channel once. If then a valid channel is created/re-named (due to user error in naming etc), it will trigger another search.
- In the odd case where TS admins might rename several channels each of which is considered valid one at a time causing a switch to occur, therefor causing the previous channel to be stored each time, the previous channel will only get stored once, until a move back to the previous channel (on server/game/pipe close) has occurred.
- Replaced tabs in C++ with spaces...damn VS not taking my settings properly >:(

extensions/src/ACRE2Shared/_CONSTANTS.h
```diff
@@ -15,7 +15,8 @@
 #undef _DEBUG_DIALOG
 
 
-#define PIPE_TIMEOUT 10000
+//#define PIPE_TIMEOUT 10000
+#define PIPE_TIMEOUT 5000 //Lowered this as a 10 second wait before being moved is very noticable. This is however my opinion/preference, so is not required.
```
jonpas: I am not sure what kind of impact this has on performance. @Sniperhid 

Should remove the second part of comment and the previous value that is commented out above.

extensions/src/ACRE2Core/setSetting.h
```diff
@@ -26,6 +26,11 @@ RPC_FUNCTION(setSetting) {
         if (CAcreSettings::getInstance()->getDisableUnmuteClients() != (value != 1)) {
             CAcreSettings::getInstance()->setDisableUnmuteClients(value != 1);
         }
+    } 
+    else if(name == "disableTeamspeakChannelSwitch") {
```
jonpas: Space after `if`.

extensions/src/ACRE2Core/setSetting.h
```diff
@@ -26,6 +26,11 @@ RPC_FUNCTION(setSetting) {
...
+        if(CAcreSettings::getInstance()->getDisableTeamspeakChannelSwitch() != (value != 1)) {
```
jonpas: Space after `if`.

extensions/src/ACRE2Core/NamedPipeServer.cpp
```diff
@@ -271,6 +272,12 @@ ACRE_RESULT CNamedPipeServer::readLoop() {
                 this->setConnectedRead(FALSE);
                 break;
             }
+
+            //Run channel switch to server channel
+            if(CEngine::getInstance()->getClient()->shouldSwitchTSChannel()) {
```
jonpas: Space after `if`.

addons/sys_io/fnc_teamspeakChannelCheck.sqf
```diff
@@ -0,0 +1,25 @@
+/*
+ * Author: Tim Beswick
+ * Checks if main display is visible and sets server name, triggering teamspeak channel move.
```
jonpas: `TeamSpeak 3`

jonpas (changes requested):

jonpas:
I think using `ts3` instead of `teamspeak` would more consistent with other TeamSpeak functions and variables we have in SQF. Also shorter.

jonpas (approved):

jonpas:
Will wait for this to be tested by @tbeswick96 's community. :)

Sniperhid (approved):

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -18,6 +21,8 @@
 #define DEFAULT_TS3_CHANNEL "ACRE"
 
 extern TS3Functions ts3Functions;
+using namespace std;
```
jonpas: Nope nope nope.

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -488,51 +493,76 @@ ACRE_RESULT CTS3Client::moveToPreviousTS3Channel() {
...
             } else {
                 //LOG("Can't get channel name");
             }
-            ts3Functions.freeMemory(channelName);
         }
-        if (name != DEFAULT_TS3_CHANNEL) {
-            //LOG("Can't find server channel, looking for default");
-            return findChannelByName(DEFAULT_TS3_CHANNEL);
+
+        for(auto& element : channelMap) {
```
jonpas: Space after `for`.

jonpas (changes requested):

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -434,7 +433,7 @@ ACRE_RESULT CTS3Client::moveToServerTS3Channel() {
     //Only switch channel if enabled in settings
     if (!CAcreSettings::getInstance()->getDisableTS3ChannelSwitch()) {
         anyID clientId;
-        string serverName = getServerName();
+		std::string serverName = getServerName();
```
jonpas: Fix your Visual Studio! 😆

tbeswick96: Goddammit! Sorry XD

jonpas:
@tbeswick96 on Slack:
> Just ran first test with teamspeak channel switching...all worked fine (after taking ages for people to update teamspeak and install it)
server was named "UKSF | Private Operations" and channels in teamspeak were "ACRE" and "ACRE - Secondary". As expected, all moved to "ACRE"

This PR looks good, I'd just like to know about that `PIPE_TIMEOUT` reducing @Sniperhid .

addons/sys_io/fnc_ts3ChannelCheck.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * Checks if main display is visible and sets server name, triggering ts3 channel move.
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call acre_sys_server_fnc_teamspeakChannelCheck
```
jonpas: `_sys_io_`

addons/sys_io/fnc_ts3ChannelCheck.sqf
```diff
@@ -0,0 +1,25 @@
+/*
+ * Author: Tim Beswick
+ * Checks if main display is visible and sets server name, triggering ts3 channel move.
```
jonpas: `TeamSpeak 3`

addons/sys_io/fnc_server.sqf
```diff
@@ -49,6 +49,7 @@ DFUNC(connectionFnc) = {
                 } else {
                     hint "ACRE CONNECTED";
                 };
+                call FUNC(ts3ChannelCheck);
```
jonpas: Why not call it after `GVAR(serverStarted)` is set to `true` on L55? Then you don't need that extra check in the function itself.

addons/sys_io/fnc_ts3ChannelCheck.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * [] call acre_sys_server_fnc_teamspeakChannelCheck
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+[{
+    params ["", "_idPFH"];
+
+    if (GVAR(serverStarted) && {!(isNull (findDisplay 46))}) exitWith {
```
jonpas: Why must this wait for main display to initialize?

jonpas (changes requested):

extensions/src/ACRE2Shared/_CONSTANTS.h
Sniperhid: Should be okay - just lowers the timeout from the TeamSpeak 3 plugin waiting for a message from the game.

jonpas: 👍 If we can get rid of the commented old timeout and the comment that would be great @tbeswick96

addons/sys_io/fnc_ts3ChannelCheck.sqf
tbeswick96: This ensures it only moves people once the briefing is complete. On the server when you press OK in the lobby, you come to the briefing screen, then there is the Continue button which starts the game. Personally I think being able to chat normally in the briefing screen is better than being moved at that stage.

addons/sys_io/fnc_server.sqf
tbeswick96: Excellent point :)

addons/sys_io/fnc_ts3ChannelCheck.sqf
jonpas: Ah, I see. In that case we should use the CBA Display EH:
```
class Extended_DisplayLoad_EventHandlers {
    class RscDisplayMission {
        ADDON = QUOTE(_this call COMPILE_FILE(XEH_missionDisplayLoad.sqf));
    };
};
```
And obviously move that code to the `XEH_missionDisplayLoad.sqf` file. No need for a PFH then.

jonpas (approved):

jonpas:
Ready to go from my side. @tbeswick96 may still do some improvements regarding compiler warnings.

lubomirblazekcz:
is this working with latest TS version? I'm having problems to get it work

jonpas:
Yes it is, make sure to read the [documentation](http://acre2.idi-systems.com/wiki/user/ts-channel-switching) on it if you haven't yet.

lubomirblazekcz:
I did, but doesn't work for me. When I connect on the server nothing happens.. I have two rooms on TS3 server, WAITING ROOM nad ACRE. I've looked for the TeamSpeak Channel Switch setting, but did not find. How can I set this for dedicated server, did I miss something?

jonpas:
Via [CBA Settings System](https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System), you have to enable it there, it's a global setting, so you just have to set it on the server or in a mission.

Also, I suggest you join the Slack or make a new issue in the future, so we don't revive old PRs. :)

PR #227, opened 2017-02-22
acre_api_fnc_setCustomSignalFunc - reset fix

Dorbedo:
- fixes the possibility to reset the function to default behavior
- changes the reset value from an emtpy array (nil) into an emtpy code({})


addons/api/fnc_setCustomSignalFunc.sqf
```diff
@@ -14,14 +14,17 @@
...
  * [mySignalFunc] call acre_api_fnc_setCustomSignalFunc
+ * [{}] call acre_api_fnc_setCustomSignalFunc
  *
  */
 #include "script_component.hpp"
 
-params ["_code"];
-
-if (!(_code isEqualType {})) exitWith {
+If !(_this params [["_code",{},[{}]]]) exitWith {
```
jonpas: Lower-case if.

addons/api/fnc_setCustomSignalFunc.sqf
```diff
@@ -14,14 +14,17 @@
...
     ERROR("acre_api_fnc_setCustomSignalFunc called with invalid argument.");
 };
 
-EGVAR(sys_signal,customSignalFunc) = _code;
+If (_code isEqualTo {}) then {
```
jonpas: Lower-case if.

jonpas (changes requested):

jonpas (dismissed):

Sniperhid:
What is the advantage of this?

`[] call acre_api_fnc_setCustomSignalFunc;` - Presently works fine to reset the custom signal function.

Dorbedo:
Did you try it in sheduled environment?

Sniperhid (approved):

Sniperhid:
Good point! This does indeed fix it in that case.

Issue #222, opened 2017-02-21
Feature Request: Modify the TS Mute Range


Mezilsa:
Good evening,

would it be possible to make the TS mute range editable for a mission creator?

Sincerly
Mezilsa

jonpas:
What would be the use case? The mute range is there only to improve performance, you can't hear anyone in direct that far anyways, but it does not impact radio communications.

Mezilsa:
The use case would be that you can't estimate the distance to a unmuted player (which would be actual less than 300 meters). If the range would be modular or just realy big it wouldn't be as interesting to try to do this.

Soldia1138:
Added in #228

PR #228, opened 2017-02-22
Introduce Variable to deactivated distance-based muting

Soldia1138:
- Add feature requested in #222


addons/sys_core/fnc_speaking.sqf
```diff
@@ -230,7 +230,7 @@ DFUNC(speakingLoop) = {
         private _unit = _x;
         if (!isNull _unit) then {
             if (!IS_MUTED(_unit) && _unit != acre_player) then {
-                if ((getPosASL _unit) distance ACRE_LISTENER_POS < 300) then {
+                if ((getPosASL _unit) distance ACRE_LISTENER_POS < 300 && GVAR(enableDistanceMuting)) then {
```
jonpas: This would profit from lazy evaluation.

jonpas (approved):

PR #233, opened 2017-02-24
Fix the antenna direction calculation

Soldia1138:
- Fixes the antenna direction calculation. Previously the rotation speed of the antenna was twice the rotation speed of the unit


jonpas (approved):
Nice!

PR #240, opened 2017-02-24
Fix Race condition

### LINKED ISSUE #82: ACRE fails to initialise - Can't hear anyone (but everyone else can hear you on direct)
This is an issue in the current versions that has a low probability of occurring (~1 in 15). The temporary solution is to restart the game/teamspeak.

In my tinkering it seems that the client variable `ACRE_MAP_LOADED` remains `false` and that setting it to `true` tends to fix the effected client.

Sniperhid:
- Should fix #82 - Assuming my prediction of a race condition is correct - will need the test in the ticket to be carried out before merging

### DIFF
--- a/extensions/src/ACRE2Arma/common/dispatch.hpp
+++ b/extensions/src/ACRE2Arma/common/dispatch.hpp
@@ -132,7 +132,6 @@ namespace acre {
 
     protected:
         void monitor() {
-            _ready = false;
             while (!_stop) {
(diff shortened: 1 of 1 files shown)


jonpas (approved):
@Sniperhid
> According to sparfell my fix for the race condition should work :smiley:
> As he got the output to confirm what I suspected was occuring (edited)

PR #252, opened 2017-03-07
Use CBA Namespaces

jonpas:
- Title
- Fix `Wrong location draw style - "acre"`
- Privatize some newly created namespaces on fast hash pool creation
- Fix UI performance issue (performance/profiling exe)


Soldia1138 (approved):
Tested.
- no more log message
- no more UI lag

PR #257, opened 2017-03-11
Fix SEM70

Soldia1138:
Hotfix:

- Frequency default (stupid)
- PBOPrefix was ooold


jonpas (approved):
GitHub decided to show all commits, such a looser.

PR #273, opened 2017-03-28
Fix Radio Signal Loss documentation links

jonpas:
- Title


NouberNou:
Still on going?

jonpas:
Nope, done.

docs/wiki/user/radio-signal-loss.md
```diff
@@ -51,7 +51,7 @@ API function example:
 
 `true` enables the full duplex mode.
 
-###  Ignore Antenna direction (New setting to be introduced in v2.2.0)
+###  Ignore Antenna direction (introduced in v2.2.0)
```
Soldia1138, line 54: Not sure about the upper/lowercases. But it's only a minor thing.

Soldia1138 (approved):

docs/wiki/user/radio-signal-loss.md
jonpas: I thought of streamlining it in this PR, but it has to be streamlined across all files, so that'll be in another PR (if I'll make one).

PR #275, opened 2017-03-30
Remove DFUNC and DEFUNC macros

### LINKED ISSUE #215: Document and reimplement D(E)FUNC macros
~~https://github.com/IDI-Systems/acre2/blob/fb6ce402358ede72d11484ef752d0fd93a63a6e6/addons/main/script_macros.hpp#L79-L80
are just copies of `(E)FUNC`, once all PRs convert we should just remove them.~~

Document `D(E)FUNC` properly with the "definite function" use case.

TheMagnetar:
- Remove D(E)FUNC macros since they are duplicates of (E)FUNC.
- Closes #215.

- [x] Remove D(E)FUNC macro copies.
- [ ] Update documentation accordingly.

### DIFF
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -76,9 +76,6 @@ Antenna Defines
 
 #define LOAD_SOUND(className) [QUOTE(className)] call EFUNC(sys_sounds,loadSound);
 
-#define DFUNC(var1) TRIPLES(ADDON,fnc,var1)
-#define DEFUNC(var1,var2) TRIPLES(DOUBLES(PREFIX,var1),fnc,var2)
-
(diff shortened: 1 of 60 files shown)


NouberNou:
There was a purpose to having these, in that when used they will register the functions without wrapping them in possible debugging wrappers used when stack tracing is turned on... If we plan to never use the stack tracing functionality in CBA macros then this is fine to remove, but otherwise there was a purpose for this.

jonpas:
Which is why I didn't do that yet, but forgot to comment on the issue.

jonpas:
Going to leave those and rather document them properly, I edited the issue.

Hope you used search and replace anyways. 😆

TheMagnetar:
No, I did it manually over the whole night! :(

PR #280, opened 2017-04-02
Fix distortion test plugin and map gen compilation

jonpas:
- Fix compile errors in `DistortionTestPlugin` and `mapgen`
- Update `DistortionTestPlugin` with TeamSpeak3 Channel Switching functionalities (copy)
- Add omni-directional support to `mapgen`

### DIFF
--- a/extensions/src/ACRE2Arma/signal/signal.hpp
+++ b/extensions/src/ACRE2Arma/signal/signal.hpp
@@ -356,7 +356,7 @@ namespace acre {
             }
 
             void signal_map_chunk(std::vector<signal_map_result> *results, float start_y, float start_x, uint32_t y_offset, uint32_t length, uint32_t x_size, float sample_size,
-                float rx_antenna_height, glm::vec3 tx_pos_, glm::vec3 tx_dir_, antenna_p &tx_antenna_, antenna_p &rx_antenna_, float frequency_, float power_, float sinad_, float omnidirectional_
+                float rx_antenna_height, glm::vec3 tx_pos_, glm::vec3 tx_dir_, antenna_p &tx_antenna_, antenna_p &rx_antenna_, float frequency_, float power_, float sinad_, bool omnidirectional_
                 ) {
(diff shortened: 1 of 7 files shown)


extensions/src/ACRE2Arma/signal/signal.hpp
```diff
@@ -239,6 +239,7 @@ namespace acre {
 
                 float lower_sensitivity = args_.as_float();
                 float upper_sensitivity = args_.as_float();
+                bool omnidirectional = args_.as_int();
```
jonpas, line 242: Need to use `bool omnidirectional = args_.as_int() == 1;`.

TheMagnetar (changes requested):
I do not think that this  is necessary in c++. 1 is already "true". There is no difference.

jonpas:
It is done the same way on L108, I think doing it explicitly might be cleaner for consistency as it is already used like that. Either way, something for another PR if we want improvement on it.

extensions/src/ACRE2Arma/signal/signal.hpp
```diff
@@ -355,7 +356,7 @@ namespace acre {
             }
 
             void signal_map_chunk(std::vector<signal_map_result> *results, float start_y, float start_x, uint32_t y_offset, uint32_t length, uint32_t x_size, float sample_size,
-                float rx_antenna_height, glm::vec3 tx_pos_, glm::vec3 tx_dir_, antenna_p &tx_antenna_, antenna_p &rx_antenna_, float frequency_, float power_, float sinad_
+                float rx_antenna_height, glm::vec3 tx_pos_, glm::vec3 tx_dir_, antenna_p &tx_antenna_, antenna_p &rx_antenna_, float frequency_, float power_, float sinad_, bool omnidirectional_
```
jonpas, line 359: Would actually be good to use that new variable on L373...

jonpas:
Tested and works, good to go.

TheMagnetar (approved):

PR #293, opened 2017-04-07
Vehicleracks fixes

### LINKED ISSUE #283: Vehicle racks issues
**Description:**
- After the testing session the following issues must be addressed.

- [x] Prevent the use of the radio if it is in use by other players. If possible limit this when radio is transmitting. Unmounting should always be possible and the transmission should end.
- [x] Assign proper roles for the radio racks. For the moment some racks are only available to the driver (replace by commander).
- [x] Apply the same system as in sys_intercom for configuring rack access.
- [x] Enhance the system for `turnedout` units and `external access`.
- [x] Fix duplicate IDs caused by lower case treatment of radios.
- [x] Investigate the following RPT code snippet (Related to removing a radio while transmitting?)
- [x] Integrate with intercom.

`17:40:54 [ACRE] (sys_data) WARNING: System event handler function attachComponent is not defined! idi\acre\addons\sys_data\fnc_noApiSystemFunction.sqf:21` <<< This is to remind us to expand the attachComponent code.

TheMagnetar:
- Closes #283.

- [x] Applies the same system as in sys_intercom for configuring rack access, while at the same time enhancing it with the options `inside`, `external`, `copilot` and `turnedout`.
- [x] Configures rack access to intercom networks.
- [x] Prevent the use of the radio if it is in use by other players.
- [x] When transmission is not active, direct speach is not heard.
- [x] Assign proper roles for the radio racks.
- [x] Fix duplicate IDs caused by lower case treatment of radios.
- [x] Fix `Recieved new rack ID (xxxx) for vehicle (zzzzz) but no entry in queue rack.` message.
- [x] ~~Fix ` System event handler function attachComponent is not defined! idi\acre\addons\sys_data\fnc_noApiSystemFunction.sqf`~~ <--- This is to remind us to expand the component system.


addons/sys_rack/CfgEventHandlers.hpp
```diff
@@ -15,3 +15,21 @@ class Extended_PostInit_EventHandlers {
         clientInit = QUOTE(call COMPILE_FILE(XEH_postInit));
     };
 };
+
+class Extended_InitPost_EventHandlers {
```
jonpas: Should use `CBA_fnc_addClassEventHandler` as in `sys_intercom`.

jonpas (changes requested):

addons/sys_rack/CfgEventHandlers.hpp
TheMagnetar: I did not change this since this is what snippers coded. Once all this is ready and working as expected (no warnings in the rpt as the ones we observed) I will put it up to date. Besides, the `ace_interaction` code is  going to suffer heavy changes after what we discussed...

jonpas: Ok.

PR #298, opened 2017-04-09
Prevent switch channel fast to change between groups

TheMagnetar:
- Prevents the switch channel  keybind to exceed the maximum number of channels within a group.

**Issue**
- Using the Knob on the 148 one can change channels within a group only.
- PRC152 and PRC343 switch functionality only allows to change channels as turning the knob would, but not on the PRC148.

### DIFF
--- a/addons/sys_core/fnc_switchChannelFast.sqf
+++ b/addons/sys_core/fnc_switchChannelFast.sqf
@@ -35,6 +35,10 @@ if (_isManpack == 0) then {
         case ("ACRE_PRC152"): {
             _channel = (_channel + _dir) min 5;
         };
+        case ("ACRE_PRC148"): {
+            private _currentGroup = floor((_channel - 1) / 16);
+            _channel = ((_channel + _dir) max (_currentGroup*16 + 1)) min ((_currentGroup + 1)*16);
+        };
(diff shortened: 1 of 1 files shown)


addons/sys_core/fnc_switchChannelFast.sqf
```diff
@@ -35,6 +35,10 @@ if (_isManpack == 0) then {
         case ("ACRE_PRC152"): {
             _channel = (_channel + _dir) min 5;
         };
+        case ("ACRE_PRC148"): {
```
jonpas: This can be combined with the first case like this (or vice-versa):
```
case "ACRE_PRC343";
case "ACRE_PRC148": {
    // magic
};
```
As per BI wiki:
> If a case has no code block, the code of the next case will automatically be executed. This makes it possible to formulate a logical "or" for cases which otherwise would contain the exact same code.

TheMagnetar: Thanks for the tip

Soldia1138 (approved):
For a moment I was thinking about having this one as a config entry for the RadioClasses...but that would result in more lines of code, so yeah. Good to go

PR #299, opened 2017-04-09
Fix collect function

TheMagnetar:
- Fix warning due to ToLower treatment of radio base class.


TheMagnetar:
This should be ready.

PabstMirror:
Not sure if fixed elsewhere, but fnc_collect.sqf is not using `toLower` on `find _baseRadio;`

TheMagnetar:
@PabstMirror It was something I forgot from vehicleracks branch. Thanks for pointing it out.

Soldia1138 (approved):

PR #305, opened 2017-04-16
Reverse order of racks

Soldia1138:
- Somehow the order of racks in ACE interaction was reversed (e.g MRAP had "Dashboard Lower" above "Dashboard Upper"). This corrects it.


TheMagnetar:
I think it depends on how the racks are defined in CfgVehicles. Does it make sense we modify it in ACE interaction?

Soldia1138:
As jonpas stated in Slack, ACE adds the action in a counter-intuitive way (Counter-Clock Wise/at top). This way addon makers don't make to do it manually.

TheMagnetar:
But it is in the right order when in list form...

Soldia1138:
Damn you ACE...need to find another solution

PR #306, opened 2017-04-18
Rack intercom Receiving/Transmitting functionality

TheMagnetar:
- Rack radios can now be configured to have the following functionality if they are connected to an intercom:
  - No monitoring (radio is not used):
  - Receiving only (radio can only be used for receiving incoming transmissions)
  - Transmitting only (radio can only be used for transmitting)
  - Receive and transmit functionality.
  - Radio status persists on a seat level. If a player enters a seat with a configured rack radio, the previous functionality will automatically load.
  - The same happens when unmounting/mounting a radio or when substituting it.
- All players in the intercom will have this functionality.
- Requires ACE3


addons/sys_intercom/fnc_setRxTxCapabilities.sqf
```diff
@@ -18,11 +18,14 @@
...
 
-params ["_rackId", "_vehicle", "_unit", "_functionality"];
+params ["_radioId", "_vehicle", "_unit", "_functionality", ["_rackId", objNull]];
 
 private _found = false;
 private _vehicleConfiguration = _vehicle getVariable [QGVAR(rackTxRxConfig), []];
-_rackId = toLower _rackId;
+
+if (isNull _rackId) then {
+    private _rackId = toLower ([toLower _radioId] call EFUNC(sys_rack,getRackFromRadio));
```
Cuel: Should not be private here?

TheMagnetar: Yes :) indeed. This code is still WIP however and it lacks revision from my side.

PR #310, opened 2017-04-19
Fix spatial functions

Soldia1138:
- ACE interaction spatial functions did not respect the selected radio but used active radio at all times. This fixes it

### DIFF
--- a/addons/sys_core/fnc_switchRadioEar.sqf
+++ b/addons/sys_core/fnc_switchRadioEar.sqf
@@ -4,6 +4,7 @@
  *
  * Arguments:
  * 0: Spatial mode (-1 = left, 0 = center, 1 = right) <NUMBER>
+ * 1: Radio ID (optional) <STRING>
  *
  * Return Value:
  * Handled <BOOL>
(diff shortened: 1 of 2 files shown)


addons/sys_core/fnc_switchRadioEar.sqf
```diff
@@ -15,9 +16,7 @@
  */
 #include "script_component.hpp"
 
-params ["_ear"];
-
-private _radioId = ACRE_ACTIVE_RADIO;
+params ["_ear",["_radioId", ACRE_ACTIVE_RADIO, [""]]];
```
jonpas: Space between arguments.

addons/sys_core/fnc_switchRadioEar.sqf
```diff
@@ -4,6 +4,7 @@
  *
  * Arguments:
  * 0: Spatial mode (-1 = left, 0 = center, 1 = right) <NUMBER>
+ * 1: Radio ID (optional) <STRING>
```
jonpas: ` * 1: Radio ID <STRING> (default: ACRE_ACTIVE_RADIO)`

jonpas (changes requested):

jonpas:
Interesting, I could bet that worked for me always, did it get bugged recently?

jonpas (approved):

PR #326, opened 2017-04-29
Open radio on base action point in addition to sub-action

### LINKED ISSUE #281: Open radio on main radio interaction in addition to sub-action
~~Simply lowers the amount of interaction nodes while keeping it straightforward.~~ Faster to access for advanced users.

jonpas:
- Title
- Close #281 
- Fix rack radio actions (all the actions that you can do on personal radios didn't show up, such as spatial... etc.)

Opening radio on base action should be done for racks as well, there is currently no way apart from keybind (which is broken from where this was branched off). I will leave that to @TheMagnetar as he knows the code.

~~I think there is no need for an "Open" text either, should be pretty self-explanatory that the radio icon opens the radio.~~ Changed to action on both nodes.

### DIFF
--- a/addons/ace_interact/fnc_radioListChildrenActions.sqf
+++ b/addons/ace_interact/fnc_radioListChildrenActions.sqf
@@ -53,10 +53,7 @@ private _radioList = [] call EFUNC(api,getCurrentRadioList);
         {
             [(_this select 2) select 0] call EFUNC(sys_radio,openRadio)
         },
-        {
-            (_this select 2) params ["_radio"];
-            !((_radio in ACRE_ACCESSIBLE_RACK_RADIOS && {isTurnedOut acre_player}) || (toLower _radio) in ACRE_HEARABLE_RACK_RADIOS)
-        },
+        {true},
         {_this call FUNC(radioChildrenActions)},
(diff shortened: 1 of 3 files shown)


addons/ace_interact/fnc_radioListChildrenActions.sqf
```diff
@@ -46,9 +46,22 @@ private _radioList = [] call EFUNC(api,getCurrentRadioList);
...
+    private _action = [
+        _x,
+        _displayName,
+        _picture,
+        {
+            [(_this select 2) select 0] call EFUNC(sys_radio,openRadio)
+        },
+        {
+            (_this select 2) params ["_radio"];
+            !((_radio in ACRE_ACCESSIBLE_RACK_RADIOS && {isTurnedOut acre_player}) || (toLower _radio) in ACRE_HEARABLE_RACK_RADIOS)
```
TheMagnetar: This is not necessary since openRadio handles all these cases already. It even shows a hint :)

TheMagnetar (approved):
I changed some lines since the EFUNC(sys_radio,openRadio) already handles if a radio can or cannot be opened. It is not that you want to hide the action like it was in fnc_radioChildrenActions.sqf

PR #332, opened 2017-05-01
No use of attach Object

### LINKED ISSUE #325: Vehicle rack issues
Investigate the following:

- [ ] Framedrop when setting the rack as active (might be fixed in #332).
- [x] Garbage-collecting message appears when deleting the vehicle (might be fixed in #324).
- [ ] When using the rack, the radio start transmitting/end transmitting sound appeared doubled (might be fixed in #332).
- [x] Add missing racks to vanilla vehicles.

TheMagnetar:
- Removes object racks when vehicle is destroyed.
- No more attachTo.
- No more framedrop (Partially fixes #325)

### DIFF
--- a/addons/sys_rack/fnc_addRackOnReturn.sqf
+++ b/addons/sys_rack/fnc_addRackOnReturn.sqf
@@ -25,7 +25,7 @@ if (isNull _rack) exitWith {
 
 private _vehicleRacks = _vehicle getVariable [QGVAR(vehicleRacks), []];
 _vehicleRacks pushBackUnique _rackId;
-_vehicle setVariable [QGVAR(vehicleRacks), _vehicleRacks];
+_vehicle setVariable [QGVAR(vehicleRacks), _vehicleRacks, true];
 _rack setVariable [QGVAR(rackVehicle), _vehicle, true];
 
 private _queue = _vehicle getVariable [QGVAR(queue),[]];
--- a/addons/sys_rack/fnc_addRackOnReturn.sqf
+++ b/addons/sys_rack/fnc_addRackOnReturn.sqf
@@ -18,7 +18,7 @@
 
 params ["_vehicle", "_rackId"];
 
-private _rack = _rackId createVehicle [0,0,0];
+private _rack = _rackId createVehicle [-1000,-1000,-1000];
 if (isNull _rack) exitWith {
     WARNING_2("Error creating rack '%2' for vehicle %1",str _vehicle,_rackId);
 };
--- a/addons/sys_rack/fnc_addRackOnReturn.sqf
+++ b/addons/sys_rack/fnc_addRackOnReturn.sqf
@@ -18,7 +18,7 @@
 
 params ["_vehicle", "_rackId"];
 
-private _rack = _rackId createVehicle [-1000,-1000,-1000];
+private _rack = createVehicle [_rackId, [-1000,-1000,-1000], [], 0, "CAN_COLLIDE"];
 _rack enableSimulation false;
 if (isNull _rack) exitWith {
     WARNING_2("Error creating rack '%2' for vehicle %1",str _vehicle,_rackId);
--- a/addons/sys_rack/fnc_addRackOnReturn.sqf
+++ b/addons/sys_rack/fnc_addRackOnReturn.sqf
@@ -19,6 +19,7 @@
 params ["_vehicle", "_rackId"];
 
 private _rack = _rackId createVehicle [-1000,-1000,-1000];
+_rack enableSimulation false;
 if (isNull _rack) exitWith {
     WARNING_2("Error creating rack '%2' for vehicle %1",str _vehicle,_rackId);
 };
--- a/addons/sys_rack/fnc_addRackOnReturn.sqf
+++ b/addons/sys_rack/fnc_addRackOnReturn.sqf
@@ -16,14 +16,17 @@
  */
 #include "script_component.hpp"
 
-params ["_vehicle","_rackId"];
+params ["_vehicle", "_rackId"];
 
 private _rack = _rackId createVehicle [0,0,0];
 if (isNull _rack) exitWith {
-    WARNING_2("Error creating rack '%2' for vehicle %1", str _vehicle, _rackId);
+    WARNING_2("Error creating rack '%2' for vehicle %1",str _vehicle,_rackId);
 };
 
-_rack attachTo [_vehicle,[0,0,0]];
+private _vehicleRacks = _vehicle getVariable [QGVAR(vehicleRacks), []];
+_vehicleRacks pushBackUnique _rackId;
+_vehicle setVariable [QGVAR(vehicleRacks), _vehicleRacks];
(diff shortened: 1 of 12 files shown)


addons/sys_server/fnc_masterIdTracker.sqf
```diff
@@ -137,9 +137,8 @@ if (GVAR(doFullSearch)) then {
                 };
             };
         };
+    } forEach ((ATLToASL [-1000,-1000,0]) nearObjects ["ACRE_baseRack", 1]);
```
jonpas: `[-1000,-1000,-1000]` like above?

addons/sys_rack/fnc_addRackOnReturn.sqf
```diff
@@ -16,14 +16,18 @@
  */
 #include "script_component.hpp"
 
-params ["_vehicle","_rackId"];
+params ["_vehicle", "_rackId"];
 
-private _rack = _rackId createVehicle [0,0,0];
+private _rack = _rackId createVehicle [-1000,-1000,-1000];
```
jonpas: Should use Alternative Syntax `createVehicle [type, position, markers, placement, special]` with `CAN_COLLIDE` to make sure `nearObjects` will pick it up in case other objects are there.

addons/sys_rack/fnc_getVehicleRacks.sqf
```diff
@@ -18,11 +18,11 @@
 params ["_vehicle"];
 
 private _racks = [];
+
 {
-    private _rackId = typeOf _x;
-    if (_rackId isKindOf "ACRE_BaseRack") then {
-        _racks pushBack _rackId;
+    if (_x isKindOf "ACRE_BaseRack") then {
```
jonpas: This check is no longer necessary as the variable will only ever have those.

addons/sys_server/fnc_masterIdTracker.sqf
jonpas: `[-1000,-1000,-1000]` like above?

addons/sys_rack/fnc_getRackFromRadio.sqf
```diff
@@ -23,6 +23,6 @@ private _return = "";
 {
     private _rackId = typeOf _x;
     if (([_rackId] call FUNC(getMountedRadio)) == _radioId) exitWith {_return = _rackId;};
-} forEach (allMissionObjects "ACRE_baseRack");
+} forEach ((ATLToASL [-1000,-1000,0]) nearObjects ["ACRE_baseRack", 1]);
```
jonpas: `[-1000,-1000,-1000]` like above?

addons/sys_rack/fnc_getHearableVehicleRacks.sqf
```diff
@@ -19,13 +19,11 @@
...
 
 private _racks = [];
+private _racks = [];
 {
-    private _rackId = typeOf _x;
-    if (_rackId isKindOf "ACRE_BaseRack") then {
-        if ([_rackId, _unit, _vehicle] call FUNC(isRackHearable)) then {
-            _racks pushBack _rackId;
-        };
+    if (_x isKindOf "ACRE_BaseRack" && {[_x, _unit, _vehicle] call FUNC(isRackHearable)}) then {
```
jonpas: First check is no longer necessary as the variable will only ever have those.

addons/sys_rack/fnc_getHearableVehicleRacks.sqf
```diff
@@ -19,13 +19,11 @@
 params ["_vehicle", "_unit"];
 
 private _racks = [];
+private _racks = [];
```
jonpas: 2 times `_rack`.

addons/sys_rack/fnc_getAccessibleVehicleRacks.sqf
```diff
@@ -20,12 +20,9 @@ params ["_vehicle", "_unit"];
 
 private _racks = [];
 {
-    private _rackId = typeOf _x;
-    if (_rackId isKindOf "ACRE_BaseRack") then {
-        if ([_rackId, _unit, _vehicle] call FUNC(isRackAccessible)) then {
-            _racks pushBack _rackId;
-        };
+    if (_x isKindOf "ACRE_BaseRack" && {[_x, _unit, _vehicle] call FUNC(isRackAccessible)}) then {
```
jonpas: First check is no longer necessary as the variable will only ever have those.

jonpas (changes requested):
Apart from the above commented improvement, this PR has numerous advantages:
- `nearObjects` within 1m range instead of super expensive `allMissionObjects`
- Using variable instead of `attachedObjects`, reducing need for checking
- Less code and nicer handling of racks.

addons/sys_rack/fnc_addRackOnReturn.sqf
jonpas: Should use Alternative Syntax `createVehicle [type, position, markers, placement, special]` with `CAN_COLLIDE` to make sure `nearObjects` will pick it up in case other objects are there. The `nearObjects` syntax used below also uses PositionAGL so you don't have to transform `ATLToASL` then.

jonpas (changes requested):

addons/sys_rack/fnc_getRackFromRadio.sqf
TheMagnetar: Same.

addons/sys_server/fnc_masterIdTracker.sqf
TheMagnetar: No, this works fine, look at the conversion. Otherwise we would end up not finding the objects. It was baffling me why it was not working.

addons/sys_rack/fnc_addRackOnReturn.sqf
TheMagnetar: If you do the PositionAGL thing without converting none of the objects will be found. It was baffling me.

jonpas: That's because AGL and ATL are not the same, but if both use AGL it should work just fine. Either way this should use the alternative syntax due to `CAN_COLLIDE`.

TheMagnetar: OK, but it does not work fine. I tested...

jonpas: Not talking about conversion anymore. :D

jonpas (approved):

jonpas:
What about just removing the rack upon vehicle killed EH?

Dorbedo:
Does it fire when the vehicle is deleted before killed?

jonpas:
Unsure, but there is also `Deleted` EH.

Dorbedo:
sorry @jonpas :
I've just delete my comment "Does it fire, when the vehicle is deleted"

I've never tested the new deleted Event, if it's reliable under all circumstances.

jonpas:
Can be optimized with EHs at a later time.

PR #336, opened 2017-05-05
Limit open radio

TheMagnetar:
- Prevent two or more players from opening the same radio GUI. This is important for radio racks and man-pack radios.
- Allow external manpack radios to be configured by both owner and external user.

### DIFF
--- a/addons/sys_external/fnc_initRadio.sqf
+++ b/addons/sys_external/fnc_initRadio.sqf
@@ -20,5 +20,5 @@ params ["_radioId"];
 [_radioId, false] call FUNC(allowExternalUse);
 [_radioId, "setState", ["isUsedExternally", [false, objNull]]] call EFUNC(sys_data,dataEvent);
 
-// To prevent the GUI from being simultaneously opened.
+// To prevent the GUI from being simultaneously opened
 [_radioId, "setState", ["isGuiOpened", false]] call EFUNC(sys_data,dataEvent);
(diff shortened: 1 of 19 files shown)


addons/sys_rack/stringtable.xml
```diff
@@ -13,6 +13,9 @@
             <English>Radio used externally. Only owner may configure.</English>
             <Japanese>無線機は外部で使われています。オーナーのみが設定可能です。</Japanese>
         </Key>
+        <Key ID="STR_ACRE_sys_radio_alreadyOpenRadio">
+            <English>Radio interface is opened by another player.</English>
```
jonpas: Too immersion breaking! 😆 
Something like "Radio in hand of an external user." might be better.

addons/sys_external/fnc_initRadio.sqf
```diff
@@ -19,3 +19,6 @@ params ["_radioId"];
 
 [_radioId, false] call FUNC(allowExternalUse);
 [_radioId, "setState", ["isUsedExternally", [false, objNull]]] call EFUNC(sys_data,dataEvent);
+
+// To prevent the GUI from being simultaneously opened.
```
jonpas: No final dot.

jonpas (changes requested):

addons/sys_external/fnc_initRadio.sqf
```diff
@@ -19,3 +19,6 @@ params ["_radioId"];
...
+[_radioId, "setState", ["isGuiOpened", false]] call EFUNC(sys_data,dataEvent);
```
Soldia1138: Although I personally don't like states beginning with a verb, this is already the case with other states. So we can leave it as it is

Soldia1138 (approved):

addons/sys_external/fnc_initRadio.sqf
TheMagnetar: No problem, I  can change  it.

jonpas: Let's do that in another PR later then, consistency is more important.

jonpas (approved):

PR #342, opened 2017-05-08
Inf. phones should not use ANY racks

TheMagnetar:
- Fixes possibility to access racks  from other vehicles when using the infantry phone.

### DIFF
--- a/addons/sys_rack/fnc_isRackHearable.sqf
+++ b/addons/sys_rack/fnc_isRackHearable.sqf
@@ -26,7 +26,7 @@ if (isNull _vehicle) then {
 if (!alive _vehicle) exitWith {false};
 
 // Infantry phone units are not allowed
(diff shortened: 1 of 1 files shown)


addons/sys_rack/fnc_isRackHearable.sqf
```diff
@@ -26,7 +26,7 @@ if (isNull _vehicle) then {
 if (!alive _vehicle) exitWith {false};
 
 // Infantry phone units are not allowed
-if (_unit isEqualTo ((_vehicle getVariable [QEGVAR(sys_intercom,unitInfantryPhone), [objNull, 0]]) select 0)) exitWith {false};
+if (_unit isEqualTo (vehicle _unit)) exitWith {false};
```
jonpas: `_unit == vehicle _unit` to be consistent with how we check if in vehicle everywhere else.

jonpas (approved):

PR #344, opened 2017-05-10
No verb at the beginning for radio states

TheMagnetar:
- As discussed in another PR. Changes radio states from having a verb at the beginning: "isShared" to "radioShared", ...

### DIFF
--- a/addons/sys_external/fnc_allowExternalUse.sqf
+++ b/addons/sys_external/fnc_allowExternalUse.sqf
@@ -18,4 +18,4 @@
 
 params ["_radioId", "_allowExternalUse"];
 
-[_radioId, "setState", ["isShared", _allowExternalUse]] call EFUNC(sys_data,dataEvent);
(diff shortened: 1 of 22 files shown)


jonpas:
This seems pretty redundant to me, all it does is make the event names longer. If those are passed over network, shorter is better, and just having `radio` everywhere seems rather weird, no?

TheMagnetar:
Shorter messages do not mean faster messages. Often in networking a minimum message size is specified or, you combine several messages that need to be sent into a single one.

jonpas:
"Often", this is Arma, this is the reason CBA uses short names such as "sCBA" for internal events. It shouldn't matter here either way.

Soldia1138:
Well. It's not necessary. There was a reason for this but I don't remember anymore. Can delete/postpone this.

Soldia1138 (approved):

Sniperhid (approved):

PR #348, opened 2017-05-14
Racks no longer use lowercase unique ids

### LINKED ISSUE #345: Lower case radio
- Treat all radios as lower case. At the moment the system is a mix of upper/lower case. fnc_MasterIDtracker handles them lowercase as well as sys_rack. This will ensure consistency.

TheMagnetar:
- No lower case treatment of radio IDs
- Closes #345


jonpas:
Should we note that in some comment somewhere or on the wiki?

TheMagnetar:
No. This is just internal. It does not affect any public function or file.

jonpas:
No, it matters to developers.

Would lower-case radio class cause issues with this?

TheMagnetar:
No. This is just how racked radios and racks were treated (internally) before. Nothing else. Racks function as before, only that the unique IDs got  converted internally to be lower case. This is no longer the case so that we avoid having to  convert them to lower case in order to avoid duplicates.

jonpas (dismissed):

jonpas:
Delaying because this needs a way closer look to make sure people can still use either case without breaking the whole system.

PR #355, opened 2017-05-19
Teamspeak channel name setting

tbeswick96:
- Add the ability to set the name of a teamspeak channel with CBA settings which will be used instead of the arma server name.
- Utilize a different version of the function used to determine the best channel match.
- Clean up the functions to remove logging.
- Update docs to reflect changes.

The setting is in CBA settings as per the other setting. This brings the total number of settings relating to teamspeak channel switching to 2:
- TeamSpeak Channel Switch: Enables/Disables automatic switching of ts channels
- TeamSpeak Channel Name: Sets the name of the ts channel to switch to

When a name is not given in the settings (""), the arma server name is used instead, as per the current behaviour.


docs/wiki/user/ts-channel-switching.md
```diff
@@ -4,10 +4,11 @@ title: TeamSpeak Channel Switching
 
 ## Setup
 
-ACRE2 includes the option to automatically switch to a dedicated TeamSpeak 3 channel upon game start. This is enabled by default and can be turned off with CBA settings. Players are moved when the game starts, after the briefing (upon pressing "Continue").
+ACRE2 includes the option to automatically switch to a dedicated TeamSpeak 3 channel upon game start. This is enabled by default and can be turned off. Players are moved when the game starts, after the briefing (upon pressing "Continue"). This feature can be disabled in the addon settings.
 
 Dedicated TS3 channels need to be created for this to function. Channel names must contain "ACRE" to be detected as a dedicated channel. 
 Channel names can contain all or part of the Arma 3 server name. If this is the case, the best matching channel name will be selected to switch to. This allows for multiple ACRE dedicated channels to be active at once for multiple Arma servers.
+It is also possible to set the name of the desired channel in the addon settings. When this is filled in, the best matching channel will be used. When empty, the Arma 3 server name will be used as above.
```
jonpas: You'll need an empty line between each paragraph, otherwise they will all jump to the same line and just wrap. You can see that on current live doc already.

docs/wiki/user/ts-channel-switching.md
```diff
@@ -16,4 +17,22 @@ Note: The name matching is considered fuzzy, meaning it will not always provide
...
+
+##### TeamSpeak Channel Switch
+
+- Enables/Disables automatic TeamSpeak channel switching
+- Default: Enabled
+
+##### TeamSpeak Channel Name
+
+- Sets the name of the TeamSpeak channel to switch to
+- Default: Empty ("")
```
jonpas: Use `code` markup, might as well for above setting too.

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -474,96 +460,95 @@ ACRE_RESULT CTS3Client::moveToPreviousTS3Channel() {
...
     uint64 *channelList;
     uint64 channelId = INVALID_TS3_CHANNEL;
     char* channelName;
     std::map<uint64, std::string> channelMap;
     int bestDistance = 40;
     uint64 bestChannelId = INVALID_TS3_CHANNEL;
+    std::string name = names[1];
+
+    if (names[0] != "") {
+        name = names[0];
```
jonpas: Since this is a vector, using `names.at()` would be safer.

jonpas (changes requested):

jonpas (approved):

PR #359, opened 2017-05-23
Prevent sharing accessible rack radios

TheMagnetar:
- Prevent sharing actions on directly accessible rack radios as well. Do not share it. Pick it yourself!

### DIFF
--- a/addons/sys_external/fnc_checkListChildrenActions.sqf
+++ b/addons/sys_external/fnc_checkListChildrenActions.sqf
@@ -25,7 +25,7 @@ private _isAvailable = true;
 // Do not allow an external user to return the headset if the radio is in use
 if (_isUsedExternally && (_unit != _user)) exitWith {false};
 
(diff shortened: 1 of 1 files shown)


addons/sys_external/fnc_checkListChildrenActions.sqf
```diff
@@ -25,7 +25,8 @@ private _isAvailable = true;
 // Do not allow an external user to return the headset if the radio is in use
 if (_isUsedExternally && (_unit != _user)) exitWith {false};
 
-// Prevent from taking a radio that can be also heard through the intercom
-if ([_radioId, _unit] call EFUNC(sys_rack,isRadioHearable)) exitWith {false};
+// Prevent from taking a radio that can be also heard through the intercom or directly accessible racks.
```
jonpas: No final dot.

jonpas (changes requested):

jonpas (approved):

PR #360, opened 2017-05-23
Proper initialisation of racks & new api functions

### LINKED ISSUE #364: Racks cannot be initialised from outside the vehicle in dedicated
Racks cannot be initialised from outside the vehicle in dedicated servers. This has not been observer in local servers, where the racks can be initialised and new racks can be added.

However, if the player enters the vehicle and locally executes one of the rack API functions, racks are properly initialised or added. Changing the locality of the vehicle (to be local to the player) or remote executing to where the vehicle is local does not solve the situation.

TheMagnetar:
- Adds a new API functions:
  - `fnc_areVehicleRacksInitialized`: returns whether the vehicle racks have been initialised for the given vehicle.
  - `fnc_addRackToVehicle`: adds a rack to a vehicle.
  - `fnc_removeRackFromVehicle`: removes a rack from a vehicle.
  - `fnc_isRackRadioRemovable`: returns if a racked radio can be unmounted.
- Executes globally the ACE interaction when initialising a rack through the API function `fnc_initVehicleRacks`.

Closes #364


TheMagnetar:
I forgot to say that this is ready before leaving. I tested this in dedicated and it worked as intended.

addons/api/XEH_postInit.sqf
```diff
@@ -0,0 +1,154 @@
...
+        if (_type == (toLower _rackId)) exitWith {
+            deleteVehicle _x;
+        };
+    } forEach (nearestObjects [[-1000,-1000], ["ACRE_baseRack"], 1, true]);
+}] call CBA_fnc_addEventHandler;
+
+[QGVAR(mountRackRadio), {
+    params ["_rackId", "_baseRadio"];
+
+    if (!([_rackId] call EFUNC(sys_radio,radioExists))) exitWith {
```
jonpas: If we do those checks in the event they won't be printed where called (server) and that might be very confusing as you won't see the log where you'd expect it to. Is there something we can't check before we send the event?

TheMagnetar: At least not the `radioExist` function, and that should be the first one to be checked I think.

jonpas: Don't really like this, but not sure if we have a good way around it. Maybe make it clear in the header that errors might be logged on clients instead? Or another backwards event for logging on where it was executed.

jonpas (approved):

PR #361, opened 2017-05-23
Allow configuring racks without intercom

TheMagnetar:
- Title


jonpas (approved):
Might look counter-intuitive, but I actually like it as it is a much rarer thing and satisfies the default nicely.

PR #380, opened 2017-07-06
Fix wrong coordination system for antenna calculation

Soldia1138:
Fix the antenna direction calculation. Currently all antennas are always looking northwards.


Sniperhid (approved):
LGTM

jonpas (approved):

PR #384, opened 2017-07-23
Remove intercom/Infantry phone from UAV and UGV vehicles

### LINKED ISSUE #379: Drones have Infantry Phone
As it can be seen in this picture, drones (at least the darter) got the infantry phone. 😄 

https://images-ext-2.discordapp.net/external/jqlrr88kZwfrwVfR9sY6BNUFJBkCFu567YsyVMcxdV8/https/i.imgur.com/8sBIdicr.jpg

TheMagnetar:
- Title
- Closes #379

### DIFF
--- a/addons/sys_intercom/CfgVehicles.hpp
+++ b/addons/sys_intercom/CfgVehicles.hpp
@@ -277,4 +277,46 @@ class CfgVehicles {
         acre_passengerIntercomExceptions[] = {};
         acre_passengerIntercomConnections = -1;
     };
+
+    class UAV_01_base_F: Helicopter_Base_F {
+        acre_hasCrewIntercom = 0;
+        acre_crewIntercomPositions[] = {};
+        acre_crewIntercomExceptions[] = {};
+        acre_hasInfantryPhone = 0;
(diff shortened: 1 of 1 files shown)


jonpas:
Do we really need to redefine all of that instead of just setting the main entries?

Sniperhid:
An easier method would be test if the vehicle is a UAV with either `unitIsUav` script command which apparently just looks up if the `isUAV` config property on the vehicle config which is true for all the vanilla vehicles.

I'd suggest a `if (unitIsUav _target) exitWith {};` in -> https://github.com/IDI-Systems/acre2/blob/master/addons/sys_intercom/fnc_intercomConfig.sqf

Soldia1138:
Are UGVs counted as `unitIsUav` as well? If so, I would stick with magnetars proposal. Heading for autonomous racks

jonpas:
> Are UGVs counted as `unitIsUav` as well?

It should be, it has `isUav = 1` in config and that's basically what the command does.

TheMagnetar:
Done in overhauled system.

jonpas:
Done in #411.

PR #393, opened 2017-08-26
updated masses for radios

kOepi123:

- it will adjust more correct weights to the radios stated:
1.: 117 : with batteries. reference:  http://olive-drab.com/od_electronics_anprc117.php
2.: 77 : with 72 cm antenna and headset. reference Field manual, pic attached.

<a href="https://ibb.co/b50JMk">77_Capture</a>


PabstMirror:
This will make it heavier than some backpacks support.
e.g. the vanilla "assault pack" `maximumLoad = 160;`
So this may break loadouts/missions

`mass` in arma is really representing both mass and volume
e.g. 20 liters of Styrofoam might weigh 2 pounds, but it won't fit in your backpack

The 117 isn't that large and probably would fit into most backpacks

kOepi123:
I understand the problem. 
Both radios would fit into a 1st generation Molle patrol pack ( 30 cm wide ),
I honestly doubt they would fit into an assault pack ( when I compare the molle patrol pack appearance in relation to the back of a person with the assault pack to a players back relation.

Arguing like this is questioning the weight-mass-result / weight system itself and continuing can cause more disproportions in the future with other subjects.

I think it does not align with the concept of ACRE/ACE to think like this, but its your mod, I just change some numbers for more accurate balancing.

jonpas:
Those are some drastic changes for the 343, 148 and 152, they will definitely break loadouts. I think our current values work well for those, since IRL they are usually strapped to the vest or similar in another pouch.

kOepi123:
if you think you will never change those settings, its definetly a good option not to change it now.
If it can be considered to be changed in the future to realistic values, it will break a lot more then, than now.

The disrelation between custom mods since the ace-era due to not having a standard value for space/weight, magazines not fitting into vests or similar will never improve if not tried to set the obvious standard to accurate masses and adjust everything depending on it.
I dont want to push anything here, but I currently try to synchronize a few mods and acre is one of them, so take it, if you want to.

jonpas:
We will discuss this internally. Can you provide more information on how you determined the values to give us a better idea? As said the problem is this value combines both weight and volume of an item.

This PR is using tabs (must be spaces) and files are missing an empty new line at the end, that is required.

kOepi123:
Converting KG to mass values is by factor 22.02 - 22.04.
I tested it with the ACE system, so measuring the actual weight / mass through the inventory.
Maybe ACE could have set a nicer factor there, I dont know.
Since the introduction of a realistic stamina system the upper load limit is finally determined by how much one person is strong enough to carry.
Modmaker who need to balance for vanilla aswell, cannot limit the load by mass and need to set capacity/volume limits to balance the load configuration.

The weight of a working prc 77 is determined consisting of one handset, one 72 cm antenna, without frame, without bag/cover, with two batteries and battery carrier and the naked radio itself.

kOepi123:
is there anything one can do with the "weight = " value?
I saw this set in some places.

jonpas:
I don't think so.

What I was asking more is how you determine a good balance between weight and volume, as in how do you put together weight and volume into a `mass` config value? Going by any formula?

kOepi123:
if I would use a formula to lessen the actual weight IRL, it would lessen the effect of the weight impact in ace stamina.
E.G.:
If I have a 15 kg weapon, which is currently in the game 10 kg and I would use a formula to combine ("balance") mass and volume ( lets say: 2/3 of real life weight ), it would destroy the balance with the stamina system and realistic loadouts. This would also destroy my point making this pull requests.

The good "balance" is depending on a realistic stamina system ( which I think was taken over from my own calculations some time ago ), that way it is focusing on stamina mainly, secondarily on real life volume.

kOepi123:
I have a different Idea:
You could make an ACE compatibility patch, which would change those values and people who want this or cannot have it ( due to destroying loadouts ) can disable or enable it themselves.
I could do this myself, probably not to all acre coding standards though.
Should I a make another issue thread for this proposal?

jonpas:
Heya @kOepi123, we had an internal discussion about this pull request, and we have decided not to merge it. We are still thankful for your wanted contribution, but at this time with Arma's "mass" system, it is simply not feasible to change it. If ACE3 or other mod would ever add volume/weight separation at any time, we would definitely be up for supporting that.

addons/sys_sem70/CfgWeapons.hpp
```diff
@@ -26,7 +26,7 @@ class CfgWeapons {
         type = 4096;
         simulation = "ItemMineDetector";
         class ItemInfo {
-            mass = 120;
+            mass = 176.2;
```
commy2, line 29: Just pointing out that these are supposed to be integers.

kOepi123: I dont know, They work, that is what I know.

PR #421, opened 2017-10-20
External radio fixes

TheMagnetar:

- [x] Fixes deadlock if the player has the radio GUI opened and gets killed.
- [x] Fixes deadlock if player gets disconnected while having a radio GUI opened.
- [x] Fixes race condition when 2 players try to open a rack or shared  radio at the same time.
- [x] SEM70 radio GUI could not be locked.
- [x] handle the race condition when disconnecting.

### DIFF
--- a/addons/sys_external/fnc_onPlayerKilled.sqf
+++ b/addons/sys_external/fnc_onPlayerKilled.sqf
@@ -18,7 +18,7 @@
 params ["_unit"];
 
 if (_unit == acre_player) then {
-    // All external radios in use are now returned to the owner.
+    // All external radios in use are now returned to the owner
     {
         private _owner = [_x] call FUNC(getExternalRadioOwner);
         [_x, _owner] call FUNC(stopUsingExternalRadio);
--- a/addons/sys_external/fnc_onPlayerKilled.sqf
+++ b/addons/sys_external/fnc_onPlayerKilled.sqf
@@ -1,0 +1,28 @@
+/*
+ * Author: ACRE2Team
+ * Handles death of unit, primarily for handling local player death. External radios are returned to the original owner.
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ *
+ * Return Value:
+ * Handled <BOOL>
+ *
+ * Example:
+ * [player] call acre_sys_external_fnc_onPlayerKilled
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit"];
+
+if (_unit == acre_player) then {
+    // All external radios in use are now returned to the owner.
+    {
+        private _owner = [_x] call FUNC(getExternalRadioOwner);
+        [_x, _owner] call FUNC(stopUsingExternalRadio);
+    } forEach ACRE_ACTIVE_EXTERNAL_RADIOS;
+};
+
+true
--- a/addons/sys_radio/XEH_postInit.sqf
+++ b/addons/sys_radio/XEH_postInit.sqf
@@ -1,14 +1,14 @@
 #include "script_component.hpp"
 
 if (isServer) then {
-    // Close the radio if it was opened in case of disconnection in order to prevend dead lock state
+    // Close the radio if it was opened in case of disconnection in order to prevent dead lock state
     GVAR(playerDisconnected) = addMissionEventHandler ["HandleDisconnected", {
         /*
          * Arguments:
          * 0: Disconnected unit <OBJECT>
          * 1: Unique DirectPlay ID <NUMBER>
          * 2: Steam ID of the leaving player <NUMBER>
-         * 3: profile name of the leaving player <STRING>
+         * 3: Profile name of the leaving player <STRING>
          */
         params ["_unit", "", "", ""];
         {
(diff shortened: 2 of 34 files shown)


addons/sys_external/fnc_onPlayerKilled.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * [player] call acre_sys_external_fnc_onPlayerKilled
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit"];
+
+if (_unit == acre_player) then {
+    // All external radios in use are now returned to the owner.
```
jonpas: No final dot.

addons/sys_radio/XEH_postInit.sqf
```diff
@@ -1,5 +1,15 @@
 #include "script_component.hpp"
 
+if (isServer) then {
+    // Close the radio if it was opened in case of disconnection in order to prevend dead lock state
+    GVAR(playerDisconnected) = addMissionEventHandler ["HandleDisconnected", {
+        params ["_unit", "", "", ""];
```
jonpas: Remove empty params?

jonpas (changes requested):
Neat fix.

addons/sys_radio/XEH_postInit.sqf
TheMagnetar: Didn't we decide to keep empty parameters? At least this is what was applied to radio functions that use  ACRE2 events...

jonpas: Oh, right, I guess so! But it doesn't say which is which in here since no header.

TheMagnetar: I will add those. However, I need more testing regarding this fix and how much impact has sending to  each player the radio list when it changes...

Sniperhid (approved):

addons/sys_radio/XEH_postInit.sqf
```diff
@@ -1,5 +1,26 @@
 #include "script_component.hpp"
 
+if (isServer) then {
+    // Close the radio if it was opened in case of disconnection in order to prevend dead lock state
```
jonpas: Prevent

addons/sys_radio/XEH_postInit.sqf
```diff
@@ -1,5 +1,26 @@
...
+    GVAR(playerDisconnected) = addMissionEventHandler ["HandleDisconnected", {
+        /*
+         * Arguments:
+         * 0: Disconnected unit <OBJECT>
+         * 1: Unique DirectPlay ID <NUMBER>
+         * 2: Steam ID of the leaving player <NUMBER>
+         * 3: profile name of the leaving player <STRING>
```
jonpas: Profile

addons/sys_radio/fnc_setRadioOpenState.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_radioId", "_isRadioOpen"];
+
+[_radioId, "setState", ["radioGuiOpened", _isRadioOpen]] call EFUNC(sys_data,dataEvent);
+
+// In case of external radios or accessible racked radios, update the status on the server as well.
```
jonpas: No final dot.

addons/sys_radio/stringtable.xml
```diff
@@ -65,5 +65,8 @@
             <Chinesesimp>无线电正在被另一名玩家进行设置。</Chinesesimp>
             <Korean>다른 플레이어가 무전기를 설정중입니다.</Korean>
         </Key>
+        <Key ID="STR_ACRE_sys_radio_radioJustOpened">
+            <English>Somebody beat you when trying to configure the radio and you respectfully step away.</English>
```
jonpas: Something shorter please. Eg. "Radio opened by someone else.".
I'd just use `alreadyOpenedRadio` tbh.

addons/sys_server/fnc_openRadioUpdateState.sqf
```diff
@@ -0,0 +1,32 @@
+/*
+ * Author: ACRE2Team
+ * Updates the open radio state.
+ *
+ * Arguments:
+ * 0: radio to update <STRING>
```
jonpas: Radio

addons/sys_server/fnc_openRadioUpdateState.sqf
```diff
@@ -0,0 +1,32 @@
...
+ * 1: radio opened state <BOOL>
```
jonpas: Radio

jonpas (changes requested):

addons/sys_radio/stringtable.xml
TheMagnetar: There is a string with that name already. This is a special case where two players open simultaneously the radio. In this case, the radio will b e opened and closed briefly for the one that registers later on the server.

TheMagnetar: So i will delete this one and use the generic one as well.

jonpas: Yeah, doesn't matter which case happens, it's already in use and should be descriptive enough.

jonpas (approved):

PR #423, opened 2017-10-21
TeamSpeak Channel Switch enhancements

tbeswick96:
- Added full string comparison. Will search for exact name first. Break if match found.
- Added word comparison. Will search for word matches second. Skip char comparison if match found.
- Char comparison (Levenshtein distance) occurs third.
- Bit of cleanup

### DIFF
--- a/extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
+++ b/extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
@@ -497,20 +497,22 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
         }
 
         uint64 bestChannelId = INVALID_TS3_CHANNEL;
+        auto bestMatches = 0;
         auto bestDistance = 10;
         for (auto& element : channelMap) {
-            auto elementChannelName = element.second;
+            auto fullChannelName = element.second;
             // Full comparison
-            if (elementChannelName.compare(name) == 0) {
+            if (fullChannelName.compare(name) == 0) {
                 bestChannelId = element.first;
                 break;
             }
-            const auto cleanChannelName = removeSubstrings(elementChannelName, DEFAULT_TS3_CHANNEL);
+            const auto cleanChannelName = removeSubstrings(fullChannelName, DEFAULT_TS3_CHANNEL);
             // Word comparison
             const int matches = getWordMatches(cleanChannelName, name);
-            if (matches > 0) {
+            if (matches > bestMatches) {
+                bestMatches = matches;
                 bestChannelId = element.first;
-                break;
+                continue;
             }
             // Char comparison
             const int distance = levenshteinDistance(cleanChannelName, name);
--- a/extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
+++ b/extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
@@ -432,7 +432,7 @@ ACRE_RESULT CTS3Client::unMuteAll( void ) {
 ACRE_RESULT CTS3Client::moveToServerTS3Channel() {
     if (!CAcreSettings::getInstance()->getDisableTS3ChannelSwitch()) {
         anyID clientId;
-        auto details = getTs3ChannelDetails();
+        std::vector<std::string> details = getTs3ChannelDetails();
 
         if (ts3Functions.getClientID(ts3Functions.getCurrentServerConnectionHandlerID(), &clientId) == ERROR_ok) {
             uint64 currentChannelId = INVALID_TS3_CHANNEL;
@@ -440,7 +440,7 @@ ACRE_RESULT CTS3Client::moveToServerTS3Channel() {
                 setPreviousTSChannel(currentChannelId);
             }
 
-            const auto channelId = findChannelByNames(details);
+            const uint64 channelId = findChannelByNames(details);
             if (channelId != INVALID_TS3_CHANNEL && channelId != currentChannelId) {
                 std::string password = "";
                 if (details.at(1) != "" && details.at(0) != "") {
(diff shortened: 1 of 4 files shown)


TheMagnetar:
I would prefer avoiding the use of the `auto` type. It makes the code not so clear in my opinion. Since this is a matter of personal preference, I will not ask for a change unless we all decide to.

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -475,38 +473,49 @@ ACRE_RESULT CTS3Client::moveToPreviousTS3Channel() {
...
+        if (details.at(0) != "") {
+            name = details.at(0);
+        }
         while (*channelList) {
             channelId = *channelList;
             channelList++;
+            char* channelName;
             if (ts3Functions.getChannelVariableAsString(ts3Functions.getCurrentServerConnectionHandlerID(), channelId, CHANNEL_NAME, &channelName) == ERROR_ok) {
-                std::string channelNameString = std::string(channelName);
+                auto channelNameString = std::string(channelName);
```
TheMagnetar: This is one of the examples where I think that `auto` should not be used.

jonpas:
I agree with `auto` usage not being preferable, where it should be one specific type, write out the type. Iterators and other things where it's not as specific or readable, `auto` is good.

extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
+    return matches;
+}
+
 unsigned int CTS3Client::levenshteinDistance(const std::string& string1, const std::string& string2) {
     int length1 = string1.size();
-    int length2 = string2.size();
+    const int length2 = string2.size();
 
-    auto columnStart = (decltype(length1))1;
+    const auto columnStart = decltype(length1)(1);
```
TheMagnetar: Just declare it as the right type, without using auto.

extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
+    const auto column = new decltype(length1)[length1 + 1];
     std::iota(column + columnStart, column + length1 + 1, columnStart);
 
     for (auto x = columnStart; x <= length2; x++) {
         column[0] = x;
         auto lastDiagonal = x - columnStart;
         for (auto y = columnStart; y <= length1; y++) {
-            auto oldDiagonal = column[y];
-            auto possibilities = {
+            const auto oldDiagonal = column[y];
```
TheMagnetar: this is also an integer.

extensions/src/ACRE2DistortionTestPlugin/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
                 column[y - 1] + 1,
                 lastDiagonal + (string1[y - 1] == string2[x - 1] ? 0 : 1)
             };
-            column[y] = std::min(possibilities);
+            column[y] = min(possibilities);
             lastDiagonal = oldDiagonal;
         }
     }
-    auto result = column[length1];
+    const auto result = column[length1];
```
TheMagnetar: same

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
+    return matches;
+}
+
 unsigned int CTS3Client::levenshteinDistance(const std::string& string1, const std::string& string2) {
     int length1 = string1.size();
-    int length2 = string2.size();
+    const int length2 = string2.size();
 
-    auto columnStart = (decltype(length1))1;
+    const auto columnStart = decltype(length1)(1);
```
TheMagnetar: same as above

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
 
-    auto column = new decltype(length1)[length1 + 1];
+    const auto column = new decltype(length1)[length1 + 1];
```
TheMagnetar: same as above

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
                 column[y - 1] + 1,
                 lastDiagonal + (string1[y - 1] == string2[x - 1] ? 0 : 1)
             };
-            column[y] = std::min(possibilities);
+            column[y] = min(possibilities);
             lastDiagonal = oldDiagonal;
         }
     }
-    auto result = column[length1];
+    const auto result = column[length1];
```
TheMagnetar: use of auto keyword

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -525,40 +534,64 @@ uint64 CTS3Client::findChannelByNames(std::vector<std::string> details) {
...
+    const auto column = new decltype(length1)[length1 + 1];
     std::iota(column + columnStart, column + length1 + 1, columnStart);
 
     for (auto x = columnStart; x <= length2; x++) {
         column[0] = x;
         auto lastDiagonal = x - columnStart;
         for (auto y = columnStart; y <= length1; y++) {
-            auto oldDiagonal = column[y];
-            auto possibilities = {
+            const auto oldDiagonal = column[y];
```
TheMagnetar: use of auto keyword

TheMagnetar (changes requested):
It looks really good. It is just those uses of auto keyword. I will update the C++ coding guidelines to reflect this.

tbeswick96:
Changes done

TheMagnetar:
Much appreciated!

TheMagnetar (approved):

PR #431, opened 2017-10-27
Rack PFH is not always executed

TheMagnetar:
- The PFH dealing with racks is only executed when:
  1. Unit enters a vehicle.
  2. Unit is outside the vehicle and uses an external radio.
-  The PFH stops executing once the unit is outside a vehicle and it is not connected to any rack radio.
- Optimises the initialisation loop. Previously it would configure the intercom/rack interaction at the next execution of the PFH.


TheMagnetar:
Ready

addons/sys_rack/fnc_configureRackIntercom.sqf
```diff
@@ -0,0 +1,27 @@
+/*
+ * Author: ACRE2Team
+ * Configures the interaction between racks and intercom systems
```
jonpas: Missing final dot.

addons/sys_rack/fnc_configureRackIntercom.sqf
```diff
@@ -0,0 +1,27 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_vehicle"];
+
+private _racks = [_vehicle] call FUNC(getVehicleRacks);
+{
+    private _intercoms = [_x] call FUNC(getWiredIntercoms);
+    private _rackRxTxConfig = _vehicle getVariable [QGVAR(rackRxTxConfig), []];
+    if (count _intercoms > 0 && {_rackRxTxConfig isEqualTo []}) exitWith {
```
jonpas:
```sqf
if (!(_intercoms isEqualTo []) && {_rackRxTxConfig isEqualTo []}) exitWith {
```

addons/sys_rack/fnc_enterVehicle.sqf
```diff
@@ -0,0 +1,62 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_vehicle", "_unit"];
+
+if (_unit != _vehicle) then {
+    private _initialized = _vehicle getVariable [QGVAR(initialized), false];
+
+    if (!_initialized) then {
+        //Only initialize if we are first in the crew array - This helps prevent multiple requests if multiple players enter a vehicle around the same time.
```
jonpas: Space after `//`.

addons/sys_rack/fnc_rackPFH.sqf
```diff
@@ -0,0 +1,73 @@
+/*
+ * Author: ACRE2Team
+ * Per frame execution. Sets if player is inside a vehicle and manages the access to rack radios.
+ *
+ * Arguments:
+ * 0: Array of arguments <ARRAY>
+ *  0: player unit <OBJECT>
```
jonpas: `Player`

addons/sys_rack/fnc_rackPFH.sqf
```diff
@@ -0,0 +1,73 @@
...
+ *  1: vehicle with intercom <OBJECT>
```
jonpas: `Vehicle`

addons/sys_rack/fnc_enterVehicle.sqf
```diff
@@ -0,0 +1,62 @@
...
+        } forEach _crew;
+
+        if (!isNull _firstPlayer) then {
+            if (_unit == _firstPlayer) then {
+                [_vehicle] spawn {
+                    params ["_vehicle"];
+
+                    [_vehicle] call FUNC(initVehicle);
+                    // Give some time for the racks to initialise properly
+                    sleep 0.5;
```
jonpas: There is no need to use scheduled here.
```sqf
[{
    [_this] call FUNC(configureRackIntercom);
}, _vehicle, 0.5] call CBA_fnc_waitAndExecute;
```

Or even better, `CBA_fnc_waitUntilAndExecute` that checks if racks are initialized.

addons/sys_rack/fnc_enterVehicle.sqf
```diff
@@ -0,0 +1,62 @@
...
+                    [_vehicle] call FUNC(configureRackIntercom);
+                };
+            };
+        } else { // No other players.
+            [_vehicle] spawn {
+                params ["_vehicle"];
+
+                [_vehicle] call FUNC(initVehicle);
+                // Give some time for the racks to initialise properly
+                sleep 0.5;
```
jonpas: Same as above.

jonpas (changes requested):

addons/sys_rack/fnc_enterVehicle.sqf
TheMagnetar: That second part is not really possible. Since i would have to check when each of the vehicle racks are properly initialised and radioId returned from the server. This is why I put that solution. So I will take the waitAndExecute option

jonpas: How sure is this then? If very low FPS and high desync, will it happen at all during those 0.5s?

addons/sys_rack/fnc_rackPFH.sqf
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private _radioId = [_x] call FUNC(getMountedRadio);
+        if (_radioId != "" && {!(_radioId in ACRE_ACCESSIBLE_RACK_RADIOS || _radioId in ACRE_HEARABLE_RACK_RADIOS)}) then {
+            private _functionality = [_radioId, _vehicle, _player, _x] call EFUNC(sys_intercom,getRxTxCapabilities);
+            if (_functionality != RACK_NO_MONITOR) then {
+                // Add the radio to the active list since it is already active in the intercom system
+                [_vehicle, _player, _radioId] call FUNC(startUsingMountedRadio);
+            };
+        };
+    } forEach (([_vehicle, _player] call FUNC(getHearableVehicleRacks)) apply {toLower _x});
```
jonpas, line 36: Just do `toLower` in the actual loop, this is effectively looping through hearable vehicle racks array twice as it is now.
```sqf
{
    private _hearableVehicleRack = toLower _x;
    // rest of code
} forEach ([_vehicle, _player] call FUNC(getHearableVehicleRacks));
```

addons/sys_rack/fnc_rackPFH.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+        if (_functionality == RACK_NO_MONITOR) then {_remove pushBackUnique _x;};
+    };
+
+    if (!(_isRackAccessible || _isRackHearable)) then {_remove pushBackUnique _x; };
+} forEach (ACRE_ACCESSIBLE_RACK_RADIOS + ACRE_HEARABLE_RACK_RADIOS);
+
+{
+    if (_x in ACRE_ACCESSIBLE_RACK_RADIOS) then {
+        ACRE_ACCESSIBLE_RACK_RADIOS = ACRE_ACCESSIBLE_RACK_RADIOS - [_x];
```
jonpas: Would probably get nice improvement by using `deleteAt + find`, especially since this can be ran quiet often.

addons/sys_rack/fnc_rackPFH.sqf
```diff
@@ -0,0 +1,73 @@
...
+    } else {
+        ACRE_HEARABLE_RACK_RADIOS = ACRE_HEARABLE_RACK_RADIOS - [_x];
```
jonpas: Same as above.

jonpas (changes requested):

addons/sys_rack/fnc_rackPFH.sqf
TheMagnetar: This is removed in #434. So this is not going to be in here.

addons/sys_rack/fnc_rackPFH.sqf
jonpas:
```
x1 = [1, 2, 3];
x1 = x1 - [2];
-> 0.0014 ms
```
```
x1 = [1, 2, 3];
x1 deleteAt (x1 find 2);
-> 0.0012 to 0.0014 ms
```
```
x1 = [2];
x1 = x1 - [2];
-> 0.0012 to 0.0014 ms
```
```
x1 = [2];
x1 deleteAt (x1 find 2);
-> 0.0011 to 0.0012 ms
```

TheMagnetar: this after 10k iterations. For the scope in here we are then looking at: no speed improvement or between 0.1 to 0.2 microseconds. Which transforms into none to 0.02 nanoseconds when executing this code once a second.

jonpas: Yes, but considering that it is not worse, we should still use it, it's still an enhancement. Even if we don't use many racks at once here, someone might, whatever mod.

jonpas (approved):

Sniperhid (approved):

PR #436, opened 2017-11-05
Hide volume slider on other dialog open, kill and unit change

### LINKED ISSUE #290: Volume slider stays visible after kill and respawn
**Arma 3 Version:** `1.68` (stable)
**CBA Version:** `3.2.1` (stable)
**ACRE2 Version:** `2.4.1` (stable)

**Mods:**
```
- CBA_A3
- acre2
```

**Description:**
- Holding open volume slider while getting killed will not hide it, but will persist even after respawn, until keybind to open it is pressed and released again.

**Steps to reproduce:**
- Open volume slider (hold the key)
- Get killed
- Observe

- Dedicated

**Placed Modules:**
- None

**RPT log file:**
- N/A

jonpas:
- Hide volume slider when some other dialog is open
- Hide volume slider when killed
- Hide volume slider when changing unit (eg. respawn, remote control)
- Fix #290 

Slider will still be visible during respawn timer or pressing Escape (pause menu) while holding down Volume Slider keybind, not really that important. It solves bigger problem of staying open unexpectedly.

### DIFF
--- a/addons/sys_gui/fnc_onVolumeControlKeyPressUp.sqf
+++ b/addons/sys_gui/fnc_onVolumeControlKeyPressUp.sqf
@@ -16,15 +16,15 @@
  */
 #include "script_component.hpp"
 
-if (!(alive acre_player)) exitWith { false };
+if (!GVAR(keyBlock)) exitWith {false};
 
 inGameUISetEventHandler ["PrevAction", "false"];
 inGameUISetEventHandler ["NextAction", "false"];
 
+GVAR(keyBlock) = false;
 disableSerialization;
-GVAR(KeyBlock) = false;
+
 57701 cutRsc [QGVAR(VolumeControlDialog_Close), "PLAIN"];
 call FUNC(closeVolumeControl);
-
 
 false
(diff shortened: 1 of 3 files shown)


Sniperhid (approved):
Awesome to see this Jonpas!

- Is it worth adding a is spectator check? and preventing the key being used in spectator (it won't change voice range in spectator) - Although I appreciate `dialog` will cover cases when a dialog is open.
- Potential concurrency issue where 'keyBlock' is set to true when but the slider is already open and then there is no way to close it? Perhaps an isOpened GVAR could be used in the keyUp event to check whether to close?

jonpas:
- Most spectators are their own display, which already blocks the key, but I don't see harm in adding that check.
- Wouldn't solve much, keyUp won't be called anyways then, and if you press the button again it will call keyUp when you release it, so you can still close it that way. Really only happens when you do silly things like holding down the button and pressing Escape. Would need a PFH to catch that, not worth it.

addons/sys_gui/fnc_onVolumeControlKeyPressUp.sqf
```diff
@@ -16,15 +16,15 @@
  */
 #include "script_component.hpp"
 
-if (!(alive acre_player)) exitWith { false };
+if (!GVAR(keyBlock)) exitWith {false};
```
Sniperhid: if `GVAR(keyBlock)` is `false` the interface will never close?

Sniperhid:
My bad I just noticed keyBlock is basically an isOpen flag, I was reading the changes and assumed it was a variable other components could use to disable the system based on its name alone.

jonpas:
Was already there, so I just used it properly.

PR #438, opened 2017-11-05
Don't monitor radios while in Virtual Arsenal

### LINKED ISSUE #426: Virtual Arsenal assigns radio IDs
We should not run ID assignment when in Virtual Arsenal.

Not sure if we can catch when you "Try" the loadout through it though.

jonpas:
- Fix #426 
- Adopts a 'Transaction model'. In essence ACRE won't change its behaviour from when Arsenal is opened to when it closes. The main advantage here is that all ACRE radios will now be visible in arsenal including the initialised ones.
- On arsenal opens initialised radios are switched for baseclases items, so they can be seen and removed. Stashes IDed radios virtually and the player can still receive on them whilst in arsenal.
- When arsenal is closed the radio changes are processed.


Sniperhid (approved):
Code looks good - I suggest we let this sit on the dev branch for a while as we need to guarantee the displayunload event always runs!

Another thought is that while we are successfully handling the case of blocking creating new IDs while arsenal is open. This won't gracefully handle radios already assigned to the unit that have been initialised. I have an idea for how we can handle that. 

If a client stores the radios into a local GVAR of radios they have we can prevent them from being garbage collected as long as `acre_sys_data_fnc_getPlayerRadioList` locally returns the radios it won't be GCed. So we could create a GVAR say `GVAR(aresnalRadioHold)` dump them in there replace and replace them with non-IDed radios and swap them out on close and empty >> `GVAR(aresnalRadioHold)`. This also has the added benefit of potentially reducing the number of radio ID/initialise requests until the user is finished with arsenal.

jonpas:
Good point, it could break everything if it isn't a reliable event. Will move to 2.6.0 to assure enough testing.

That idea is good, and I was thinking the same. However, we don't want to swap all radios when a player is using Arsenal in-game, he might already have the radios setup and just want to get an extra one for example. Doing that would mean he would get all radios reset.

I don't think we can really do much about that, if anything we should do it for exporting loadouts, but I don't know if we can hook into that.
Additionally, [ACE3's Arsenal rewrite](https://github.com/acemod/ACE3/pull/5734) will handle unique radios nicely (visible in Arsenal as remove-only).

Sniperhid:
` Doing that would mean he would get all radios reset.` - Not with my suggestion. Essentially your changes already hold back the client process of trying to checking radios (and initialising them). 

My suggestion is just updating the getPlayerRadioList function works well as it will allow continuing to receive radio comms and will prevent the radios getting GCed and swapping the radios while aresnal is open. The idea is simple:

Example walkthrough
- User has 152_id_1
- User opens arsenal. The 152 is placed into a GVAR (which getPlayerRadioList returns). The 152_id_1 is replaced with base 152 item locally.
- User adds/removes radio at will with no radios being cleaned up/initialized
- User closes arsenal. Inventory is checked for all radios in the GVAR (just 152_id_1) and the  matching base items are replaced.
- Radios that are no longer in inventory will be GCed, new radios will then be initialized

It isn't perfect as if you had say 3 initialised 152s there's no way to distinguish between them, but I think it's an acceptable compromise. There is also the danger if the player crashes the radio data would  be wiped and be reinitialized on joining but that isn't the end of the world either.

jonpas:
I see, that would indeed work.

jonpas:
Confirmed `DisplayUnload` always runs, it has to, even if display is forcefully destroyed due to whatever happens.

Sniperhid:
Updated PR description with the new commit changes.

addons/sys_core/fnc_arsenalOpen.sqf
```diff
@@ -0,0 +1,32 @@
+/*
+ * Author: ACRE2Team
+ * Handles open Arsenal
```
jonpas: Missing final dot and `Arsenal`.

addons/sys_core/fnc_arsenalClose.sqf
```diff
@@ -0,0 +1,33 @@
+/*
+ * Author: ACRE2Team
+ * Handles closing Aresnal
```
jonpas: Missing final dot and `Arsenal`.

addons/sys_core/fnc_arsenalClose.sqf
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call acre_sys_core_fnc_closeArsenal
+ *
+ * Public: No
+ */
+ #include "script_component.hpp"
```
jonpas: No leading space.

addons/sys_core/fnc_arsenalClose.sqf
```diff
@@ -0,0 +1,33 @@
...
+
+if (is3DEN) exitWith {}; // Exit if Eden Arsenal.
```
jonpas: No final dot.

addons/sys_core/fnc_arsenalClose.sqf
```diff
@@ -0,0 +1,33 @@
...
+    private _idx = _weapons find (toLower _baseClass);
+    if (_idx != -1) then {
+        [acre_player, _baseClass, _radio] call EFUNC(sys_core,replaceGear);
+        _weapons deleteAt _idx;
+    }
+} forEach GVAR(arsenalRadios);
+
+ACRE_ARSENAL_RADIOS = [];
+GVAR(arsenalRadios) = [];
+GVAR(arsenalOpen) = false;
```
jonpas: Missing newline at EOF.

addons/sys_core/fnc_arsenalOpen.sqf
```diff
@@ -0,0 +1,32 @@
...
+ * None
+ *
+ * Example:
+ * [] call acre_sys_core_fnc_arsenalOpen
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+if (is3DEN) exitWith {}; // Exit if Eden Arsenal.
```
jonpas: No final dot.

addons/sys_core/fnc_arsenalOpen.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+GVAR(arsenalRadios) = [];
+ACRE_ARSENAL_RADIOS = [];
+{
+    private _radio = _x;
+    private _baseClass = [_radio] call EFUNC(sys_radio,getRadioBaseClassname);
+    [acre_player, _radio, _baseClass] call EFUNC(sys_core,replaceGear);
+    GVAR(arsenalRadios) pushBack [_baseClass,_radio];
+    ACRE_ARSENAL_RADIOS pushBack _radio;
+} forEach (_weapons select {_x call EFUNC(sys_radio,isUniqueRadio)});
```
jonpas: Missing newline at EOF.

addons/sys_data/fnc_getPlayerRadioList.sqf
```diff
@@ -47,6 +47,11 @@ if (!ACRE_IS_SPECTATOR) then {
         _radioList pushBackUnique _x;
     } forEach ACRE_HEARABLE_RACK_RADIOS;
 
+    // If Arsenal is open radios are stashed in var until left.
```
jonpas: No final dot.

Sniperhid (approved):

jonpas:
Moving to 2.5.1!

PR #478, opened 2018-01-16
Fix RPT Errors on Server when changing Frequency

Soldia1138:
This will fix an RPT-error on the Server, when changing the Freuqency on the SEM70 in Manual Mode.


TheMagnetar (approved):
Nice and short!

jonpas (approved):

PR #480, opened 2018-01-21
AN/VRC-64

TheMagnetar:
- Adds rack for AN/PRC-77, the AN/VRC-64


jonpas:
Requires antenna @Soldia1138.

Should also cleanup those radio configs one day.

jonpas (approved):

PR #482, opened 2018-02-12
Don't recompile XEH Display EH's before every call

### LINKED ISSUE #439: Don't recompile XEH EH's before every call
`preStart`, `preInit` and `postInit` don't matter as much, as they are only called once on mission start. Others have shown to have a bigger impact: https://github.com/acemod/ACE3/pull/5728

jonpas:
- Close #439 
- Improve Display EH's performance
- Fix script error (usage of `QEGVAR` instead of `GVAR`) introduced in #487


TheMagnetar (approved):
From the code itself I see no issues

jonpas:
Tested and also fixed a script error introduced in #487, which would prevent rack's "Receive Only" action from being generated.

Issue #483, opened 2018-02-12
TODO list for Intercom rework and Racks


jonpas:
- UI
  - [x] Movable element in layout editor
  - [x] Prevent hiding when opening displays
  - [x] Support for variable amount of intercoms and racks (auto-resize in height)
- ACE3 Actions
  - [x] Streamline action names (length, capitalization)
  - [x] Group "Transmit Only", "Receive Only" and "Transmit and Receive" under an action point (like spatial "Ear" group)
  - [x] Fix "Unmountable (Locked)" never showing up.
- ACE3 Hints
  - ~~Streamline hint texts (length, capitalization)~~
- Keys
  - [x] Streamline key names (capitalization)
  - [x] Streamline default binds with older keys
    - Unbind both of them by default (voice activation is default and there are ACE3 interactions)
  - ~~Make "Intercom PTT" actually work like PTT (and not activate toggle)~~
- Fixes
  - [x] Fix infantry phone warning (config).
  - [x] Add master station to copilots (config)
  - [x] `[ACRE] (sys_data) WARNING: System event handler function attachComponent is not defined! idi\acre\addons\sys_data\fnc_noApiSystemFunction.sqf:21`

TheMagnetar:
Updated description with fixes so that we can track them more easily.

jonpas:
All done.

PR #484, opened 2018-02-13
Improve Vehicle Info UI

jonpas:
- Work on #483 
- Make Vehicle Info UI movable element in layout editor
- Prevent Vehicle Info UI from being hidden when opening displays (eg. ACE3 Interaction Menu)
- Add support for variable amount of intercoms and racks to Vehicle Info UI (auto-resize in height)
- Streamline ACE3 action names (length, capitalization)
- Streamline key names (capitalization)

Vehicle Info UI is now being created with `cutRsc` (like most ACE3 UIs), so it does not get hidden when opening displays. Drawback of that is that it also doesn't hide itself when pause menu is open (not really an issue).


addons/sys_intercom/fnc_intercomChildrenActions.sqf
```diff
@@ -22,10 +22,10 @@ private _actions = [];
 private _intercomNames = _target getVariable [QGVAR(intercomNames), []];
 
 {
-    (_intercomNames select _forEachIndex) params ["_intercomName", "_intercomDisplayName"];
+    (_intercomNames select _forEachIndex) params ["_intercomName", "", "_intercomShortName"];
```
TheMagnetar, line 25: Personally I would leave as it was before.

jonpas: Previously it was:
```
Main Action
    Intercom
        Crew Intercom
        Pax Intercom
```
This changes it to:
```
Main Action
    Intercom
        Crew
        Pax
```
Which I think is better, it's obviously Intercom, you already select that with first action, and it takes less space.

We can of course handle that differently for better compatibility with user-defined intercom names and so on.

TheMagnetar: Let's keep your changes.

TheMagnetar (approved):
Works like a charm

PR #497, opened 2018-03-03
Enhance  rack initialisation

TheMagnetar:
- Allow rack initialisation in all vehicle positions (previously only supported in positions returned by crew command).
- Ensure that player is initialised  in ` fnc_addRackToVehicle`. Previously extra  code was needed in init.sqf in order to work properly.

### DIFF
--- a/addons/api/fnc_addRackToVehicle.sqf
+++ b/addons/api/fnc_addRackToVehicle.sqf
@@ -85,7 +85,7 @@ private _allowed = [_vehicle, _allowed] call EFUNC(sys_core,processVehicleSystem
 private _disabled = [_vehicle, _disabled] call EFUNC(sys_core,processVehicleSystemAccessArray);
 _intercoms = _intercoms apply {toLower _x};
 
-[{
+private _condition = {
     // A player must do the action of adding a rack
     private _player = objNull;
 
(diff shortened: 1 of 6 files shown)


addons/api/fnc_addRackToVehicle.sqf
```diff
@@ -85,16 +85,31 @@ private _allowed = [_vehicle, _allowed] call EFUNC(sys_core,processVehicleSystem
...
-
-if (isDedicated) then {
-    // Pick the first player
-    _player = (allPlayers - entities "HeadlessClient_F") select 0;
-} else {
-    _player = acre_player;
-};
+[{
+    // A player must do the action of adding a rack
+    private _player = objNull;
```
jonpas: Doubled code in condition and statement, put into function (can be local function above).

jonpas (changes requested):

addons/api/fnc_addRackToVehicle.sqf
jokoho48: why is the condition !isNil "_player" that is the same as writing true? did you maybe mean !isNull _player?

TheMagnetar: if the player is not fully initialised it is nil. This is what testing showed

jonpas (approved):

PR #502, opened 2018-03-11
Polish translation

sdsznsk:
I've translated almost everything. The only thing I can't translate is STR_ACRE_sys_radio_noTransmitIntercom , since I don't know context or what it does. Please keep in mind that I'm not a professional translator ;)

- Add polish translation for most strings

### DIFF
--- a/addons/sys_core/stringtable.xml
+++ b/addons/sys_core/stringtable.xml
@@ -69,6 +69,7 @@
             <Chinesesimp>Teamspeak频道切换</Chinesesimp>
             <Korean>팀스피크 채널 변경</Korean>
             <Portuguese>Trocar de Canal do TeamSpeak</Portuguese>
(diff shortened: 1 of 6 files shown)


TheMagnetar (approved):
Looks good.

jonpas:
> The only thing I can't translate is STR_ACRE_sys_radio_noTransmitIntercom , since I don't know context or what it does.

That the rack currently can't transmit over intercom as well. @TheMagnetar that string could probably be written in a better way.

TheMagnetar:
Well it essentially means that the radio cannot be used to transmit when being connected to that intercom. There are some intercom stations where such functionality is not available.

jonpas:
@Teqashi I will merge this, you can fill out the remaining string in a separate PR.

PR #519, opened 2018-04-22
Update installation.md

ColinM9991-zz:
Add a description for the `i_channel_needed_subscribe_power` error on teamspeak when the teamspeak server has a channel with a higher subscribe power than what the user is granted.

This error on teamspeak can cause confusion as the error doesn't indicate what channel the permission restriction is for, and so users may think it is a client permission issue for the channel they're currently in.

To understand this error, I looked through the source code for the teamspeak plugin and found the following lines of code
```cpp
// subscribe to all channels to receive event
ts3Functions.requestChannelSubscribeAll(ts3Functions.getCurrentServerConnectionHandlerID(), NULL);
if (CEngine::getInstance()->getClient()->getState() != ACRE_STATE_RUNNING) {
    CEngine::getInstance()->getClient()->start((ACRE_ID)id);
}
```

I do think people need to be aware of this for easier troubleshooting.

- Describe what this pull request will do
- Each change in a separate line
- Respect the [SQF](https://github.com/IDI-Systems/acre2/wiki/Coding-Guidelines-(SQF)) and [C++](https://github.com/IDI-Systems/acre2/wiki/Coding-Guidelines-(CPP)) Coding Guidelines

### DIFF
--- a/docs/wiki/user/installation.md
+++ b/docs/wiki/user/installation.md
@@ -28,5 +28,5 @@ Make sure to launch both Arma 3 (Steam) and TeamSpeak 3 as admin, or neither.
 - **Radio beeps not audible on dedicated server**  
 Make sure your `server.cfg` has `"b64"` whitelisted in `allowedLoadFileExtensions[]` and `allowedPreprocessFileExtensions[]` if you are using that. Let your server administrator know if you have no idea what this means.
 
-- **insufficient client permissions (failed on i_channel_needed_subscribe_power)**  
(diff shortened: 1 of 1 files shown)


jonpas (changes requested):

jonpas (approved):

ColinM9991-zz:
I have briefly looked at the TeamSpeak API to resolve this, but both my understanding of ACRE2 (primarily the need to subscribe to channels) and C++ need a bit of work.

I'll review the Slack discussion and see if I can contribute a fix to reduce your workload.

PR #521, opened 2018-04-28
Remove Intercom, Infantry Phone, and Vehicle Racks from parachutes

Drofseh:
- See title

### DIFF
--- a/addons/sys_rack/CfgVehicles.hpp
+++ b/addons/sys_rack/CfgVehicles.hpp
@@ -148,7 +148,7 @@ class CfgVehicles {
     };
 
     class ParachuteBase : Helicopter {
-        class AcreIntercoms {};
+        class AcreRacks {};
     };
 
     class VTOL_01_base_F;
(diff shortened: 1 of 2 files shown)


jonpas:
What do you have against parachutes?!

Drofseh:
> What do you have against parachutes?!

They did a bad thing and lost their telephone privileges.

addons/sys_rack/CfgVehicles.hpp
```diff
@@ -147,6 +147,10 @@ class CfgVehicles {
         };
     };
 
+    class ParachuteBase : Helicopter {
+        class AcreIntercoms {};
```
TheMagnetar: Shouldn't it be  `class AcreRacks{};`

TheMagnetar (changes requested):

addons/sys_rack/CfgVehicles.hpp
Drofseh: Derp, I copied and pasted for the format and forgot to change the class.

TheMagnetar (approved):

PR #532, opened 2018-05-11
Update Japanese translation

jonpas:
@classicarma some of those strings already existed, some @EtoileRei added already. Can you confirm this is all correct?


classicarma:
They are all fine. thanks to fix the conflict!

TheMagnetar (approved):

PR #543, opened 2018-06-12
Enable channel fast switch for racked radios

tbeswick96:
- Add check to fast channel switch to allow fast switching of racked radios
- Rationale is a radio in a rack is usually as easily accessible as a normal radio

### DIFF
--- a/addons/sys_core/fnc_switchChannelFast.sqf
+++ b/addons/sys_core/fnc_switchChannelFast.sqf
@@ -23,8 +23,9 @@ private _radioId = ACRE_ACTIVE_RADIO;
 private _radioType = [_radioId] call EFUNC(sys_radio,getRadioBaseClassname);
 private _typeName = getText (configFile >> "CfgAcreComponents" >> _radioType >> "name");
 private _isManpack = getNumber (configFile >> "CfgAcreComponents" >> _radioType >> "isPackRadio");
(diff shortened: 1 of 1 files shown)


TheMagnetar (approved):

TheMagnetar (changes requested):
It needs to be considered as well the case where the player has no physical access to the rack. This can be for example a gunman in a helicopter... it would be weird to be able to switch channela when by config definition he is not able to open the radio GUI

tbeswick96:
@TheMagnetar THink that is fixed. I've included a check to ensure the radio is accessible

TheMagnetar (changes requested):

addons/sys_core/fnc_switchChannelFast.sqf
```diff
@@ -23,8 +23,9 @@ private _radioId = ACRE_ACTIVE_RADIO;
 private _radioType = [_radioId] call EFUNC(sys_radio,getRadioBaseClassname);
 private _typeName = getText (configFile >> "CfgAcreComponents" >> _radioType >> "name");
 private _isManpack = getNumber (configFile >> "CfgAcreComponents" >> _radioType >> "isPackRadio");
+private _isRackRadio = (vehicle acre_player != acre_player) && {_radioId in (([vehicle acre_player] call EFUNC(api,getVehicleRacks)) apply {[_x] call EFUNC(api,getMountedRackRadio)})} && {[_radioId, acre_player] call EFUNC(sys_rack,isRadioAccessible)};
```
jonpas, line 26: `_isRackRadioAccessible` would be more fitting.

TheMagnetar (approved):

PR #565, opened 2018-08-09
Update Chinese

GodofMonkeys:
Update Chinese


jonpas (changes requested):
Needs to be converted back to spaces, you used tabs.

jonpas (approved):

PR #566, opened 2018-08-11
C++ Code cleanup

TheMagnetar:
- Clean c++ code
- Const correctness
- Use of base types (int16_t, int32_t, ...)
- Get rid of microsoft specific types (DWORD, BYTE,...)
- Get rid of microsoft "safe" printf variants
- Get rid of macro codes (DECLARE_MEMBER, ...)
- Consistent naming of arguments and member types


extensions/src/ACRE2Core/AmplitudeAttenuation.cpp
```diff
@@ -9,43 +9,41 @@
...
     return ACRE_OK;
 }
 
-ACRE_RESULT CAmplitudeAttenuation::process(short* samples, int sampleCount, int channels, CPlayer * player) {
-    float averageSum, currentSample, averageAmplitude, ampCoef;
-    averageSum = 0;
-    for (int i = 0; i < sampleCount*channels && i < 4095; i++, iter++) {
-        if (samples[i] > 0) {
-            currentSample = 20.0f * log10((float)samples[i]);
+ACRE_RESULT CAmplitudeAttenuation::process(int16_t *const a_samples, const int32_t ac_sampleCount, const int32_t ac_channels, CPlayer *const a_player) {
```
NouberNou: Can we please not do pseudo hungarian notation? If you want to signify a variable is an argument I've always prefered a trailing _, so `player_`. Noting the constiness of an argument is also not super helpful in most modern editors/IDEs. It should show a redline.

NouberNou (commented):
In general not a big fan of the naming of function arguments or member variables. Hungarian type notation should really be avoided as it is totally redundant information.

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
```diff
@@ -16,13 +16,13 @@
 
 // trim from start
 static inline std::string &ltrim(std::string &s) {
-    s.erase(s.begin(), std::find_if(s.begin(), s.end(), std::not1(std::ptr_fun<int, int>(std::isspace))));
+    s.erase(s.begin(), std::find_if(s.begin(), s.end(), std::not1(std::ptr_fun<int32_t, int32_t>(std::isspace))));
```
dedmen, line 19: This should really be converted to using a lambda. std::ptr_fun has been deprecated in C++11. and std::not1 has been deprecated in c++17.

dedmen: Here is example code for ltrim/rtrim https://github.com/dedmen/armake/blob/cpp/src/utils.cpp#L251
Also supporting tabs.

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
```diff
@@ -210,7 +212,7 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function)
                 char *value = new char[4096];
                 //DEBUG("Read from pipe [%d]\n", hPipe);
                 ret = ReadFile(readHandle, (LPVOID)value, 4096, &cbRead, NULL);
-                if ( ! ret) {
+                if ( !ret) {
```
dedmen, line 215:
```suggestion
                if (!ret) {
```
?

extensions/src/ACRE2Core/BabbelEffect.h
```diff
@@ -16,52 +17,50 @@ class CBabbelEffect : public CSoundMonoEffect {
...
-        offset = 128;
-        sfreq = 48000;
-        mfreq = 5200;
-        lpFilter1.setup(4, sfreq, mfreq*0.3);
-        lpFilter2.setup(4, sfreq, mfreq*0.5);
+        m_offset = 128.0f;
+        m_sfreq = 48000.0f;
+        m_mfreq = 5200.0f;
+        lpFilter1.setup(4, (float64_t) m_sfreq, (float64_t) m_mfreq*0.3);
+        lpFilter2.setup(4, (float64_t) m_sfreq, (float64_t) m_mfreq*0.5);
```
dedmen, line 29: Should use proper c++ static_cast

extensions/src/ACRE2Core/BabbelEffect.h
```diff
@@ -16,52 +17,50 @@ class CBabbelEffect : public CSoundMonoEffect {
...
+    void process(int16_t *const samples, const int32_t sampleCount) {
+        float32_t *floatPointer[1], *buffer;
+        buffer = new float32_t[sampleCount];
         floatPointer[0] = buffer;
-        short *tempSamples = new short[sampleCount];
-        memcpy(tempSamples, samples, sizeof(short)*sampleCount);
-        short maxAmp = 1;
-        for (int i = 0; i < sampleCount; ++i) {
-            samples[i] = (short)(((float)tempSamples[i]*1.0f)*cos(((float)i/(sfreq/mfreq))*1.4*M_PI));
+        int16_t *tempSamples = new int16_t[sampleCount];
```
dedmen, line 35: use std::vector with resize for a temp array. That way you don't have to worry about cleanup.

extensions/src/ACRE2Core/Engine.cpp
```diff
@@ -74,15 +74,15 @@ ACRE_RESULT CEngine::initialize(IClient *client, IServer *externalServer, std::s
     return ACRE_OK;
 }
 
-ACRE_RESULT CEngine::initialize(IClient *client, IServer *externalServer, std::string fromPipeName, std::string toPipeName, std::string loggingPath) {
+ACRE_RESULT CEngine::initialize(IClient *const client, IServer *const externalServer, const std::string fromPipeName, const std::string toPipeName, const std::string loggingPath) {
 
-    g_Log = (Log *)new Log(const_cast<char *>(loggingPath.c_str()));
+    g_Log = (Log *) new Log(const_cast<char *>(loggingPath.c_str()));
```
dedmen, line 79: Why is the reason of casting `Log*` to `Log*` here?

dedmen: And why is it casting a constant string to a non constant char*, That might be undefined behavior.

extensions/src/ACRE2Core/FFT.cpp
```diff
@@ -5,56 +5,52 @@
...
+                m_result[k + j + (m >> 1)] = u - t;
+                current_omega *= m_omega[s];
             }
         }
     }
-    if (inverse == false)
-        for (int i = 0; i < n; ++i)
-            result[i] /= n;
-    return result;
+    if (m_inverse == false)
```
dedmen, line 50: `== false`?
```suggestion
    if (!m_inverse)
```

extensions/src/ACRE2Core/FilterOcclusion.cpp
```diff
@@ -6,12 +6,13 @@
 #include <cmath>
 #include <math.h>
 
-#define TS_SAMPLE_RATE 48000
-#define CUTTOFF_FREQ    15000
+#define TS_SAMPLE_RATE_Hz 48000
+#define CUTTOFF_FREQ_Hz    15000
```
dedmen, line 10:
```suggestion
#define CUTTOFF_FREQ_Hz   15000
```

dedmen: Macros are usually identified by being all upper case. The z is not.

extensions/src/ACRE2Core/FilterOcclusion.h
```diff
@@ -11,7 +11,11 @@ class CFilterOcclusion
 public:
     CFilterOcclusion(void);
     ~CFilterOcclusion(void);
-    ACRE_RESULT process(short* samples, int sampleCount, int channels, ACRE_VOLUME volume, Dsp::Filter *&filter);
+    ACRE_RESULT process(int16_t *const samples, const int32_t sampleCount, const int32_t channels, const ACRE_VOLUME volume, Dsp::Filter *&filter);
 
-    DECLARE_MEMBER(int, ChannelCount);
+    virtual __inline void setChannelCount(const int32_t value) { this->m_channelCount = value; }
```
dedmen, line 16: __inline is MSVC specific, `inline` is the real keyword.
Also inline doesn't make much sense on virtual functions. If the compiler doesn't know exactly that the class will be that exact type, then it won't be able to inline. And if it knew that it was that type, then it would also inline if you don't put the inline keyword there.

extensions/src/ACRE2Core/NamedPipeServer.h
```diff
@@ -22,33 +22,58 @@ class CNamedPipeServer : public IServer, public CLockable {
...
-    DECLARE_MEMBER(BOOL, ConnectedWrite);
-    DECLARE_MEMBER(BOOL, ConnectedRead)
-    DECLARE_MEMBER(ACRE_ID, Id);
-    DECLARE_MEMBER(BOOL, ShuttingDown);
-    DECLARE_MEMBER(std::string, FromPipeName);
-    DECLARE_MEMBER(std::string, ToPipeName);
+    bool getConnected() const { return (this->getConnectedRead() && this->getConnectedWrite()); };
+    void setConnected (const bool value) { this->setConnectedRead(value); this->setConnectedWrite(value); };
+
+    virtual __inline void setPipeHandleRead(const HANDLE value) { this->m_pipeHandleRead = value; }
```
dedmen, line 38: These functions should be `override`, not `virtual`
```suggestion
    void setPipeHandleRead(const HANDLE value) override { this->m_pipeHandleRead = value; }
```

extensions/src/ACRE2Core/SoundMonoChannel.cpp
```diff
@@ -6,111 +6,111 @@
...
+CSoundChannelMono::CSoundChannelMono(const int32_t length, bool oneShot ) {
     this->init(length, oneShot);
 }
 
 CSoundChannelMono::~CSoundChannelMono() {
-    if (this->buffer) {
-        delete this->buffer;
-        this->buffer = NULL;
+    if (this->m_buffer) {
+        delete this->m_buffer;
```
dedmen, line 36: just make it a vector and it will cleanup itself.

extensions/src/ACRE2Core/ext_remoteStopSpeaking.h
```diff
@@ -21,11 +21,11 @@ RPC_FUNCTION(ext_remoteStopSpeaking) {
             this->getSelf()->getCurveScale()
         ) */
 
-    ACRE_ID id = (ACRE_ID)vMessage->getParameterAsInt(0);
+    ACRE_ID id = (ACRE_ID) vMessage->getParameterAsInt(0);
```
dedmen, line 24: static_cast and auto?
Is the cast even necessary? Can't it implictly convert on =?

extensions/src/ACRE2Shared/Log.cpp
```diff
@@ -5,7 +5,7 @@
 
 Log *g_Log = NULL;
 
-Log::Log(char *logFile) { 
+Log::Log(char *logFile) {
```
dedmen, line 8: const char* ?

extensions/src/ACRE2TS/TS3Client.h
```diff
@@ -8,70 +8,113 @@
...
-    DECLARE_MEMBER(UINT32, SpeakerMask);
-    DECLARE_MEMBER(uint64, PreviousTSChannel);
-    DECLARE_MEMBER(std::vector<std::string>, Ts3ChannelDetails);
-    DECLARE_MEMBER(BOOL, ShouldSwitchTS3Channel)
+    std::string removeSubstrings(std::string &string, const std::string &substring);
+    ACRE_RESULT updateTs3ChannelDetails(const std::vector<std::string> details);
+    ACRE_RESULT updateShouldSwitchTS3Channel(const bool state);
+    bool shouldSwitchTS3Channel();
+
+    virtual __inline void sethadVAD(const bool value) { this->m_hadVAD = value; }
```
dedmen, line 60: setHadVad* ?
Also override.

extensions/src/Wav2B64/Wave.h
```diff
@@ -6,53 +6,48 @@
 #include <string>
 
 
-typedef struct __WAVEDESCR
+typedef struct __waveDescr
```
dedmen, line 9: This is a library struct
https://github.com/xinshangbo/WendySpeech/blob/master/src/CutVoiceCPP/Wave.h#L6

extensions/src/ACRE2Core/FilterOcclusion.cpp
TheMagnetar: This is accepted when trying to define units. mHz is not the same as MHz and so on.

jonpas:
No further activity and tons of conflicts, would need to get re-done.

PR #588, opened 2018-10-02
Expand argument checks in API functions

TheMagnetar:
- Title


jonpas:
I would like to keep `params` with 3 or less arguments to be one-lined (depending on length of checks of course), seems like a waste of space to use 3 lines for 1 argument.

I know it's not perfectly consistent, but I think it makes it a bit more readable as well.

addons/api/fnc_isRadio.sqf
```diff
@@ -13,16 +14,16 @@
...
 // check that it at least inherits from ACRE_BaseRadio *AND* has a unique ID
 // If its scoped 1, that means its either a base or not a radio.
 private _flag = getNumber (configFile >> "CfgWeapons" >> _weapon >> "acre_isRadio");
 private _id = getNumber (configFile >> "CfgWeapons" >> _weapon >> "acre_uniqueId");
 private _scope = getNumber (configFile >> "CfgWeapons" >> _weapon >> "scope");
 if (isNil "_flag" || isNil "_scope" || isNil "_id") exitWith { false };
-if (_flag == 1 && _scope == 1) exitWith { true };
 
-false
+(_flag == 1) && (_scope == 1)
```
jonpas, line 29: Could do `_flag && _scope`, no? Unsure of readability though. :D

jonpas (approved):
Looks good to me. Still not a fan of 1 param using 3 lines, but it might prove useful when we start generating documentation for API functions.

addons/api/fnc_isRadio.sqf
TheMagnetar: I prefer like it is now just for readability. Personally i find it more readable than `_flag == 1 && _scope == 1`

PR #591, opened 2018-10-04
Cleanup of sys_external

TheMagnetar:
- Title

### DIFF
--- a/addons/sys_external/fnc_childrenActions.sqf
+++ b/addons/sys_external/fnc_childrenActions.sqf
@@ -33,8 +33,8 @@ if (_playerOwnsRadio) then {
         QGVAR(giveRadio),
         _string,
         "",
-        {[(_this select 2) select 0, _target] call FUNC(stopUsingExternalRadio)},
-        {!([(_this select 2) select 0] call FUNC(isExternalRadioUsed))},
+        {[_radio, _target] call FUNC(stopUsingExternalRadio)},
+        {!([_radio] call FUNC(isExternalRadioUsed))},
         {},
         _params
     ] call ace_interact_menu_fnc_createAction;
@@ -48,8 +48,8 @@ if (_playerOwnsRadio) then {
         QGVAR(useRadio),
         _string,
         "",
-        {[(_this select 2) select 0, acre_player] call FUNC(startUsingExternalRadio)},
-        {!([(_this select 2) select 0] call FUNC(isExternalRadioUsed))},
+        {[_radio, acre_player] call FUNC(startUsingExternalRadio)},
+        {!([_radio] call FUNC(isExternalRadioUsed))},
         {},
         _params
     ] call ace_interact_menu_fnc_createAction;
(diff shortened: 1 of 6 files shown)


addons/sys_external/fnc_childrenActions.sqf
```diff
@@ -33,8 +33,8 @@ if (_playerOwnsRadio) then {
         QGVAR(giveRadio),
         _string,
         "",
-        {[(_this select 2) select 0, _target] call FUNC(stopUsingExternalRadio)},
-        {!([(_this select 2) select 0] call FUNC(isExternalRadioUsed))},
+        {[_radio, _target] call FUNC(stopUsingExternalRadio)},
```
jonpas: Where does this _radio come from? Action and condition code don't know about outer variables.

jonpas (changes requested):

addons/sys_external/fnc_childrenActions.sqf
TheMagnetar: Oh true! I was just testing and wondering :D

jonpas: There is no reason we don't just pass the radio through instead of all params though, could just use `_this#2`/`_this select 2` and done.

addons/sys_external/fnc_childrenActions.sqf
```diff
@@ -48,16 +48,16 @@ if (_playerOwnsRadio) then {
...
+        {!([(_this select 2) select 0] call FUNC(isExternalRadioUsed))},
         {},
         _params
     ] call ace_interact_menu_fnc_createAction;
     _actions pushBack [_action, [], _target];
 };
 
 // Check if we are giving or returning the headset
-if ([_radio, _target] call FUNC(checkReturnGive)) then {
+if ([(_this select 2) select 0, _target] call FUNC(checkReturnGive)) then {
```
jonpas: Could've left it here?

addons/sys_external/fnc_childrenActions.sqf
```diff
@@ -33,8 +33,8 @@ if (_playerOwnsRadio) then {
         QGVAR(giveRadio),
         _string,
         "",
-        {[(_this select 2) select 0, _target] call FUNC(stopUsingExternalRadio)},
-        {!([(_this select 2) select 0] call FUNC(isExternalRadioUsed))},
+        {[_this select 2, _target] call FUNC(stopUsingExternalRadio)},
+        {!([_this select 2] call FUNC(isExternalRadioUsed))},
         {},
         _params
```
jonpas: Replace with `_radio`.

jonpas (changes requested):

addons/sys_external/fnc_childrenActions.sqf
TheMagnetar: You sure? Isn't this what we mentioned before?

jonpas: You are passing through `[_radio] == _params`, inside ACE3's function that is `_this#2`, so you have to do `_this#2#0`. So you have to pass just `_radio` and you can use `_this#2`.

jonpas (approved):

PR #593, opened 2018-10-04
Cleanup of sys_rack

TheMagnetar:
- Title

### DIFF
--- a/addons/sys_rack/fnc_generateMountableRadioActions.sqf
+++ b/addons/sys_rack/fnc_generateMountableRadioActions.sqf
@@ -35,7 +35,7 @@ _radioList = [_rackClassName, _radioList] call FUNC(getMountableRadios);
     //private _isActive = _x isEqualTo _currentRadio;
 
     private _action = [_x, _displayName, _picture, {
-        params ["_target","_unit","_params"];
+        params ["","_unit","_params"];
         _params params ["_rackClassName","_radioId"];
         [_rackClassName, _radioId, _unit] call FUNC(mountRadio);
     }, {true}, {}, [_rackClassName, _x]] call ace_interact_menu_fnc_createAction;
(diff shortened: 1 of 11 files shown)


addons/sys_rack/fnc_generateMountableRadioActions.sqf
```diff
@@ -35,7 +35,7 @@ _radioList = [_rackClassName, _radioList] call FUNC(getMountableRadios);
     //private _isActive = _x isEqualTo _currentRadio;
 
     private _action = [_x, _displayName, _picture, {
-        params ["_target","_unit","_params"];
+        params ["","_unit","_params"];
```
jonpas, line 38: Spaces after commas.

addons/sys_rack/fnc_setState.sqf
```diff
@@ -16,6 +16,6 @@
  */
```
jonpas, line 18: Header?

addons/sys_rack/sem90/fnc_sem90MountableRadio.sqf
```diff
@@ -20,7 +20,7 @@
 
 params ["_rackId", "", "_eventData", ""];
 
-if (([_rackId] call FUNC(getMountedRadio)) != "") exitWith {false}; // If a radio is already mounted we can't mount another.
+if !(([_rackId] call FUNC(getMountedRadio)) isEqualTo "") exitWith {false}; // If a radio is already mounted we can't mount another.
```
jonpas: This is technically slightly slower, 2 commands instead of 1 (`isEqualTo` + `!`). Only advantage is it won't throw an error if types don't match, but I think we should prefer if it does in this case.

It's also longer. I wouldn't change this.

jonpas (changes requested):

jonpas (approved):

PR #594, opened 2018-10-04
Cleanup of sys_server

TheMagnetar:
- Title


jonpas (approved):
Looks good, but we should think about using unique items function from CBA for iterating through items.

Soldia1138 (approved):
LGTM

PR #597, opened 2018-10-08
Optimise item functions

TheMagnetar:
- Title
- It mainly affects the performance of masterIdTracker function.
- It also optimises the onInventoryAction function.
- Originally it was intended to use CBA_fnc_uniqueUnitItems but the result was actually worse in terms of performance

Test: 44 or 43ms after 10000 iterations
```
private _gear = weapons player;
_gear append (items player);
_gear append (assignedItems player);
```

Test: 220 or 198 after 10000 iterations
``` private _gear = [player] call CBA_fnc_uniqueItems;```

### CODE REVIEW

**Comment by jonpas** [comment]
Needs test with hundreds of items as discussed on Slack.

File: addons/sys_core/fnc_getGear.sqf [hunk 1]
```diff
@@ -21,7 +21,9 @@ if (isNull _unit) exitWith {[]};
 
 // diag_log text format["Assigned Items: %1", (assignedItems _unit)];
 
-private _gear = (weapons _unit) + (items _unit) + (assignedItems _unit);
+private _gear = weapons _unit;
+_gear append (items _unit);
+_gear append (assignedItems _unit);
```
jonpas: We only ever check for `"ItemRadio"` and `"ItemRadioAcreFlagged"` from assigned items (always 7 long), which both fit only into radio slot.

This could be improved so:
```sqf
_gear pushBack ((assignedItems _unit) select 3);
```
And we get rid of looping through 6 items pointlessly.

TheMagnetar:
More tests (with  500  ace morphine items)

New method: 470ms
CBA UniqueItems 646
Old method: 11857

TheMagnetar:
Some more tests show that getItemCargo yields faster results than itemCargo. For the same implementation, itemCargo was 3 times slower:
ItemCargo 11556ms
getItemCargo: 470ms

jonpas (approved):

PR #602, opened 2018-10-16
Fix shared radios in vehicles

TheMagnetar:
When a player was sharing a radio inside a vehicle and the speed was 0, everything was ok. This was not the case when the vehicle was moving.


Soldia1138 (approved):
Works

PR #607, opened 2018-11-06
Introduce factor for AI revealing

### LINKED ISSUE #606: Add a scalable factor for AI revealing
**Arma 3 Version:** `1.84` (stable)
**CBA Version:** `3.9.0.181012` (stable)
**ACRE2 Version:** `2.6.2.996` (stable)

**Mods:**
```
- CBA_A3
- acre2
```

**Description:**
With the current setting, an AI unit will instantly identify you on open ground
 * ~91m away when speaking at default volume (3/5)
 * ~52m away when speaking one level lower (2/5)
 * ~12m away when using the lowest volume (1/5)

From experience, this is way too much compared to what players can hear (ie. 1-2m on the quietest setting).

As such, it would be great to have some kind of factor that could be set via CBA settings to scale this down, ie. for stealth-based scenarios. Currently, there is only the on/off toggle.

Soldia1138:
Convert the AI revealing setting from a boolean to a number. This enables communities and mission makers to set how good (how far) the AI hears players.

Closes #606

### DIFF
--- a/addons/sys_core/fnc_enableRevealAI.sqf
+++ b/addons/sys_core/fnc_enableRevealAI.sqf
@@ -1,4 +1,3 @@
-#define DEBUG_MODE_FULL
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
--- a/addons/sys_core/fnc_enableRevealAI.sqf
+++ b/addons/sys_core/fnc_enableRevealAI.sqf
@@ -1,3 +1,4 @@
+#define DEBUG_MODE_FULL
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
@@ -27,7 +28,7 @@ DFUNC(monitorAI_PFH) = {
 
     //soundFactor is how loud the local player is speaking.
     private _soundFactor = EGVAR(api,selectableCurveScale); // typically 0.1 -> 1.3
-    private _multiplier = 250*(_soundFactor^2);
+    private _multiplier = GVAR(revealToAI)*250*(_soundFactor^2);
 
     private _nearUnits = (getPosATL acre_player) nearEntities ["CAManBase", (130 * _soundFactor)];
     private _startTime = diag_tickTime;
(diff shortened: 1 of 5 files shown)


addons/sys_core/fnc_enableRevealAI.sqf
```diff
@@ -1,3 +1,4 @@
+#define DEBUG_MODE_FULL
```
TheMagnetar: We should remove this before merging.

addons/sys_core/initSettings.sqf
```diff
@@ -120,10 +120,10 @@
 // Reveal to AI
 [
     QGVAR(revealToAI),
-    "CHECKBOX",
+    "SLIDER",
     localize LSTRING(revealToAI_displayName),
     "ACRE2",
-    true,
+    [0, 1, 0.50, 2],
```
TheMagnetar:
```suggestion
    [0, 1, 1, 2],
```

We should have a default value of 1. This is the default value at the moment.

TheMagnetar (changes requested):

addons/sys_core/initSettings.sqf
jonpas: Makes sense to set it realistically though, as pointed out by the issue.

Soldia1138: I'll go with 1. So the player's experience is the same as before

TheMagnetar (approved):

addons/sys_core/fnc_enableRevealAI.sqf
```diff
@@ -27,7 +27,7 @@ DFUNC(monitorAI_PFH) = {
 
     //soundFactor is how loud the local player is speaking.
     private _soundFactor = EGVAR(api,selectableCurveScale); // typically 0.1 -> 1.3
-    private _multiplier = 250*(_soundFactor^2);
+    private _multiplier = GVAR(revealToAI)*250*(_soundFactor^2);
```
jonpas: Could do with some spaces around operators.

jonpas (approved):
Code looks good, but I think we should still adjust default value to be realistic, as per the issue this PR addresses.

PR #608, opened 2018-11-07
Fix  function getGear

TheMagnetar:
- Fix function getGear.
- As it is now, it assumes that radios are in position 3, which may not be always the case (if no radios). It may cause the following: 
```
_gear = _gear select {(_x se>
22:44:56   Error Zero divisor
22:44:56 File idi\acre\addons\sys_core\fnc_getGear.sqf, line 27
22:44:56 Error in expression <;
_gear pushBack ((assignedItems _unit) select 3);
```

### DIFF
--- a/addons/sys_core/fnc_getGear.sqf
+++ b/addons/sys_core/fnc_getGear.sqf
@@ -24,7 +24,7 @@ if (isNull _unit) exitWith {[]};
 private _gear = (getItemCargo (uniformContainer _unit)) select 0;
 _gear append ((getItemCargo (vestContainer _unit)) select 0);
 _gear append ((getItemCargo (backpackContainer _unit)) select 0);
-_gear pushBack ((assignedItems _unit) select 3);
+_gear append (assignedItems _unit);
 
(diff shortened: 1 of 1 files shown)


jonpas:
Awww. :(

addons/sys_core/fnc_getGear.sqf
```diff
@@ -24,7 +24,7 @@ if (isNull _unit) exitWith {[]};
 private _gear = (getItemCargo (uniformContainer _unit)) select 0;
 _gear append ((getItemCargo (vestContainer _unit)) select 0);
 _gear append ((getItemCargo (backpackContainer _unit)) select 0);
-_gear pushBack ((assignedItems _unit) select 3);
+_gear append (assignedItems _unit);
```
jonpas, line 27:
```suggestion
_gear pushBack ((assignedItems _unit) param [3, ""]);
```

TheMagnetar: Hmmm no, I would  not risk  it. From the description: https://community.bistudio.com/wiki/assignedItems
>But if something is missing it get's omitted so you can not be sure that some element is at a constant index. 

So it may be our "item radios" are on position 1...

jonpas: I have never seen that happen before. :thinking:

jonpas (approved):

PR #628, opened 2019-02-09
Disable performance counters

TheMagnetar:
- Disables performance counters. 
- In Speaking function:
  - Average time lowered from 282.63 to 194.7 uS
  - Median time lowered from 161.00 to 147 uS

### DIFF
--- a/addons/sys_core/XEH_preInit.sqf
+++ b/addons/sys_core/XEH_preInit.sqf
@@ -95,9 +95,13 @@ acre_player = player;
 GVAR(coreCache) = HASH_CREATE;
 
 acre_sys_io_ioEventFnc = {
-    BEGIN_COUNTER(ioEventFunction);
+    #ifdef ENABLE_PERFORMANCE_COUNTERS
+        BEGIN_COUNTER(ioEventFunction);
+    #endif
     _this call EFUNC(sys_rpc,handleData);
-    END_COUNTER(ioEventFunction);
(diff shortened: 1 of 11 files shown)


jonpas (approved):
Lovely!

dedmen:
Why not just alter the `BEGIN_COUNTER`, `END_COUNTER` defines in one place, instead of putting ifdefs around them everywhere?
Just like ACE does.
https://github.com/acemod/ACE3/blob/e5a15d200f44df5fccc0bc5575d18d80b35538dd/addons/main/script_debug.hpp#L45

jonpas:
@dedmen we do have that. Just talked with @TheMagnetar profiling was done with ASP, and that's the difference it has shown. Any clue why?

dedmen:
>Any clue why?

Well I guess the defines are set even though they shouldn't be.
Have you checked the post-preprocess output to check whether the macros are defined although they shouldn't be?
Maybe someone is def'ing ENABLE_PERFORMANCE_COUNTERS before the script_debug.hpp include and undef'ing it again right after.

Maybe that difference is just in measurement tolerance, between Arma restarts things can change.
Have to check if the post-preprocess script really changed by making these changes, if it didn't change then obviously the measurements are just tolerance.

jonpas:
I didn't check the output yet, I will when I get home. Nothing is defining it in-code that I could find.

It was tested a few times, all same results, tolerance like that wouldn't be consistent at all, definitely interesting.

PR #641, opened 2019-03-04
Move antenna parsing to preStart

### LINKED ISSUE #630: Cache postInit in preStart
Ref.: https://github.com/acemod/ACE3/pull/6803

Example case: [antenna postInit](https://github.com/IDI-Systems/acre2/blob/8a48bc44e7b192ba0347438a9e419dea8f398807/addons/sys_antenna/XEH_postInit.sqf)

TheMagnetar:
- partially closes #630


TheMagnetar:
I need to test if this works at all

addons/sys_antenna/config.cpp
```diff
@@ -6,7 +6,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"acre_main", "acre_sys_components"};
+        requiredAddons[] = {"acre_main", "acre_sys_core", "acre_sys_components"};
```
jonpas: `acre_sys_core` already requires `acre_main`.

jonpas:
Load antenna returns `-1` (how even, should only return `1` or `0`). Can't test further due to crashing at the moment.

jonpas:
There, this fixes it. Should still be tested in multiplayer to make sure nothing else fails loading because of it though.

jonpas (approved):

PR #643, opened 2019-03-04
Debug - Add dump performance counters function and cleanup

jonpas:
- Bring back performance counters dump
- Move call stack dump to function


jonpas:
- Dump performance counters is using old CBA PFH internal setup, delays are always `any` (need to read from new data https://github.com/CBATeam/CBA_A3/blob/7bdde2e52b9df520368b30c167ff2afde2799c03/addons/common/fnc_addPerFrameHandler.sqf#L46)
- Call stack breaks `FUNC` macro (need to adjust)

PabstMirror:
Using updated PFEH dumper from #938
Did not touch callstack stuff, but this could still be merged and fixed later

jonpas:
Looks good to me. Can't approve, because it's my own PR. :D

TheMagnetar (approved):

PR #644, opened 2019-03-05
Support for other signal models

TheMagnetar:
- Adds support for other signal models.
- Converts the model in http://acre2.idi-systems.com/wiki/frameworks/custom-signal-processing and labels it as "arcade mode" (boost ranges).


jonpas:
This still keeps custom signal function possibility, right?

- Some configuration files were committed
- Code style inconsistencies

TheMagnetar:
Yes, that does not change anything. It is backwards compatible.

addons/sys_signal/XEH_preInit.sqf
```diff
@@ -6,10 +6,14 @@ PREP_RECOMPILE_START;
...
 PREP_RECOMPILE_END;
 
+// CBA Settings
+#include "initSettings.sqf"
+
 if (hasInterface) then {
     DGVAR(showSignalHint) = false;
     DGVAR(terrainScaling) = 1;
     DGVAR(omnidirectionalRadios) = 0;
+    DGVAR(signalModel) = SIGNAL_MODEL_LOS_MULTIPATH;
```
jonpas, line 13: Why is this necessary? It will be set with the setting registration.

addons/sys_signal/fnc_getSignal.sqf
```diff
@@ -31,11 +31,13 @@ if (_count == 0) then {
         private _txAntenna = _x;
         {
             private _rxAntenna = _x;
+            private _model = GVAR(signalModel); // TODO: Change models on the fly if compatible (underwater, better frequency matching)
```
jonpas, line 34: Changing models on the fly happens already, no? Except if you disable that in the setting below.

extensions/src/ACRE2Arma/signal/models/Arcade.cpp
```diff
@@ -0,0 +1,40 @@
+#include "Arcade.hpp"
+#include <glm/geometric.hpp>
+
+acre::signal::model::Arcade::Arcade() : SignalModel()
+{
```
jonpas: > Opening braces should always be on the same line as their block identifier.

jonpas (changes requested):

addons/sys_signal/XEH_preInit.sqf
TheMagnetar: Never have a variable uninitialised. It is just safer this way.

addons/sys_signal/fnc_getSignal.sqf
TheMagnetar: This is a note for later on. There may  be models that depend on how the map is processed but right now it is not the case.

extensions/src/ACRE2Arma/signal/models/Arcade.cpp
TheMagnetar: Done it in new files. Changes in older files will be done in separate PRs

addons/sys_signal/XEH_preInit.sqf
jonpas: But it's not uninitialized. If anything, you are changing the default from settings here.

I am not sure at which stage CBA does it, but you might be overwriting user set option.

TheMagnetar: All other variables are initialised here: Omnidirectional radios, terrain loss scaling (which is set to 1 and later changed to 0.5 by CBA_settings). So it is the right place in my opinion to init all these. Testing shows that as well.

jonpas: Seems weird and unneeded. :thinking:

TheMagnetar: Moved CBA settings after the initialisation of the variables.

jonpas: This makes a lot more sense now! :laughing: :+1:

jonpas (approved):

PR #686, opened 2019-06-03
Use tags for TS metadata

### LINKED ISSUE #665: Use tags for TS metadata
I'm not sure if this is not possible due to a limitation of TS3, haven't looked into that.

Right now you have to disable TFAR when using ACRE, and vice versa. TFAR is already using a tag system for the metadata. `<TFAR>datas</TFAR>`. Would it be possible for ACRE to use `<ACRE>datas</ACRE>`

This would allow both plugins to be active at the same time.

BrettMayson:
- Closes #665 

Tested in MP with both TFAR and ACRE plugins enabled at the same time.


extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -86,10 +86,29 @@ AcreResult CTS3Client::start(const acre_id_t id_) {
...
+    char* clientInfo;
+    anyID myID;
+    ts3Functions.getClientID(ts3Functions.getCurrentServerConnectionHandlerID(), &myID);
+    ts3Functions.getClientVariableAsString(ts3Functions.getCurrentServerConnectionHandlerID(), myID, CLIENT_META_DATA, &clientInfo);
+	std::string to_set;
+	std::string sharedMsg = clientInfo;
+	if (sharedMsg.find(START_DATA) == std::string::npos || sharedMsg.find(END_DATA) == std::string::npos) {
+		to_set = to_set + START_DATA + data + END_DATA;
+	} else {
+		std::string before = sharedMsg.substr(0, sharedMsg.find(START_DATA));
```
TheMagnetar:
```suggestion
		const std::string before = sharedMsg.substr(0, sharedMsg.find(START_DATA));
```

extensions/src/ACRE2TS/TsCallbacks_static.cpp
```diff
@@ -34,27 +34,33 @@ void ts3plugin_infoData(uint64 serverConnectionHandlerID, uint64 id, enum Plugin
...
         case PLUGIN_CHANNEL:
             break;
         case PLUGIN_CLIENT:
-            if (ts3Functions.getClientVariableAsString(serverConnectionHandlerID, (anyID)id, CLIENT_META_DATA, &metaData) == ERROR_ok) {
-                if (!metaData) { 
-                    noAcre = TRUE;
-                }
-                int metaDataLength = strlen(metaData);
-                if (metaDataLength < 3 || metaDataLength > (INFODATA_BUFSIZE - 2)) {
+            // Some code used from https://github.com/michail-nikolaev/task-force-arma-3-radio under APL-SA
```
TheMagnetar: APL SA is not compatible with GPL. Either this is refactored or this PR cannot be merged.

extensions/src/ACRE2TS/TsCallbacks_static.cpp
```diff
@@ -34,27 +34,33 @@ void ts3plugin_infoData(uint64 serverConnectionHandlerID, uint64 id, enum Plugin
...
+            if (ts3Functions.getClientVariableAsString(ts3Functions.getCurrentServerConnectionHandlerID(), (anyID)id, CLIENT_META_DATA, &metaData) == ERROR_ok) {
+                if (!metaData) {
                     noAcre = TRUE;
+                } else {
+                    sharedMsg = metaData;
+                    if (sharedMsg.find(START_DATA) == std::string::npos || sharedMsg.find(END_DATA) == std::string::npos) {
```
TheMagnetar:
```suggestion
                    if ((sharedMsg.find(START_DATA) == std::string::npos) || (sharedMsg.find(END_DATA) == std::string::npos)) {
```

extensions/src/ACRE2TS/inc/public_definitions.h
```diff
@@ -359,6 +359,10 @@ struct FileTransferCallbackExport{
 /*define for file transfer bandwith limits*/
 #define BANDWIDTH_LIMIT_UNLIMITED 0xFFFFFFFFFFFFFFFFll
 
+/*define for metadata tags*/
+#define START_DATA "<ACRE>"
```
TheMagnetar: These defines are only used in `TsCallbacks_static.cpp`. They should be declared as private defines in the cpp file or declared as `static const` in the cpp file.

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -86,10 +86,29 @@ AcreResult CTS3Client::start(const acre_id_t id_) {
...
+		std::string after = sharedMsg.substr(sharedMsg.find(END_DATA) + strlen(END_DATA), std::string::npos);
```
TheMagnetar:
```suggestion
		const std::string after = sharedMsg.substr(sharedMsg.find(END_DATA) + strlen(END_DATA), std::string::npos);
```

TheMagnetar (changes requested):

TheMagnetar:
Last but not least. Can TFAR TS plugin interfere?

extensions/src/ACRE2TS/TsCallbacks_static.cpp
BrettMayson: It's refactored quite a bit, but it's still recognizable as from TFAR. Didn't realize ACRE was GPL and not APL-SA. I can rewrite this at some point. Optionally would we be able to get permission from the author of this code to use it under GPL for ACRE?

extensions/src/ACRE2TS/inc/public_definitions.h
BrettMayson: They are not only used in `TsCallback_static.cpp`

BrettMayson:
The TFAR plugin will not interfere, right now it is only ACRE interfering with TFAR.

extensions/src/ACRE2TS/TsCallbacks_static.cpp
TheMagnetar: I am sure there are other examples on how to add and use tags

extensions/src/ACRE2TS/inc/public_definitions.h
TheMagnetar: You are right. Somehow I missed it when reviewing it using the Github interface.

extensions/src/ACRE2TS/TsCallbacks_static.cpp
BrettMayson: It's not a matter of difficulty, this isn't a hard thing to fix in the slightest. I just thought if I've already used the TFAR code for inspiration and got it fixed it may just be quicker to ask if we can use it rather than rewriting it. If that is not possible I can refactor it.

TheMagnetar: It is better that no code from TFAR enters acre2. We do not even have to take the same approach and if there is a better solution to this then we take it. Compatibility between plugins is not a top priority since loading tfar and and acre is not really an option :D

BrettMayson: I'm not doing this to have both TFAR and ACRE used at the same time. ACRE will erase data from any TeamSpeak plugin, not just TFAR. A benefit of this is that both TeamSpeak plugins can be left enabled. As someone that does joint ops with groups using TFAR but uses ACRE in my group, this will let me leave both plugins enabled rather than disabling and enabling all the time.

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -94,7 +94,7 @@ AcreResult CTS3Client::setSelfVariable(char* data) {
 	std::string newData;
 	const std::string curString = currentData;
 	if ((curString.find(START_DATA) == std::string::npos) || (curString.find(END_DATA) == std::string::npos)) {
-		newData = newData + START_DATA + data + END_DATA;
+		newData = before + START_DATA + data + END_DATA;
```
TheMagnetar: `before` is undefined in that if.

TheMagnetar (changes requested):

extensions/src/ACRE2TS/TsCallbacks_static.cpp
```diff
@@ -34,27 +34,32 @@ void ts3plugin_infoData(uint64 serverConnectionHandlerID, uint64 id, enum Plugin
...
-                if (metaDataLength < 3 || metaDataLength > (INFODATA_BUFSIZE - 2)) {
+            if (ts3Functions.getClientVariableAsString(ts3Functions.getCurrentServerConnectionHandlerID(), (anyID)id, CLIENT_META_DATA, &metaData) == ERROR_ok) {
+                if (!metaData) {
                     noAcre = TRUE;
+                } else {
+                    sharedMsg = metaData;
+                    if ((sharedMsg.find(START_DATA) == std::string::npos) || (sharedMsg.find(END_DATA) == std::string::npos)) {
+                        noAcre = TRUE;
+                    } else {
+                        result = sharedMsg.substr(sharedMsg.find(START_DATA) + strlen(START_DATA), sharedMsg.find(END_DATA) - sharedMsg.find(START_DATA) - strlen(START_DATA));
```
dedmen: you are potentially calling `strlen` at runtime. If you have c++17 you can use a string_view's length() method to get the length at compiletime.

dedmen: Also, technically the `.find(END_DATA)` should only start searching AFTER the START_DATA, if someone is funny and puts a END_DATA before a START_DATA in his metadata, you'll do a substr with negative length.

Also you can store `sharedMsg.find(START_DATA)` in a variable instead of calculating it 3 times.
`sharedMsg.find(END_DATA)` too

dedmen: Also if you have c++17 you should do the find/substr's on a string_view to prevent unnecessary copies.

dedmen: Did I make the same mistakes? *shakes head*

jonpas: We can use C++17 no problem.

TheMagnetar: Then C++17 forward on... I was under the impression we had support for only c++14.
@synixebrett if you can ammend the code as suggested by dedmen, we can greenlight the PR.

jonpas: Since recently `make.py` takes most up-to-date compiler version on the given platform. Jenkins is running VS2019.

BrettMayson:
So no clue what happened here, it's fucked though eh

jonpas:
Replaced by #719

PR #688, opened 2019-06-03
New program to generate fakewrp

TheMagnetar:
- Fixes routines mapgen and acre_map_test.
- Use of Enum class  in landscape.cpp instead of default values (-2, -1 and 0).
- Remove compile warnings.
- New routine to generate ACRE wrp formats.

TODO:
- [x] Add fakewrp for the new Arma map.
- [ ] Add ACRE wrp for GM.

### DIFF
--- a/extensions/src/ACRE2Arma/common/wrp/landscape.cpp
+++ b/extensions/src/ACRE2Arma/common/wrp/landscape.cpp
@@ -1,54 +1,67 @@
 #include "landscape.hpp"
 #include <exception>
 
-acre::wrp::landscape::landscape(std::string path_)
-{
+acre::wrp::landscape::landscape(const std::string &path_) {
     std::ifstream stream(path_, std::ifstream::binary);
     _process(stream);
     stream.close();
 }
 
-acre::wrp::landscape::landscape(std::istream &stream_)
-{
+acre::wrp::landscape::landscape(std::istream &stream_) {
     _process(stream_);
 }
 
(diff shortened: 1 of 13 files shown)


TheMagnetar:
This is ready for review. I need to test GM regen in game, but it is just a binary file.

extensions/src/ACRE2Arma/common/wrp/landscape.cpp
```diff
@@ -153,20 +166,78 @@ bool acre::wrp::landscape::_process(std::istream &stream_)
...
+AcreResult acre::wrp::landscape::generateAcreWrp(std::ofstream &out) {
+
+    /*
+     * Following the file format https://community.bistudio.com/wiki/Wrp_File_Format_-_OPRWv17_to_24
+     * in order to be compatible with our parser.
+     */
+
+    // ===================== Header =====================
+    out.write("ACRE", 4);                                // Filetype
+    out.write((char *)&version, sizeof(uint32_t));       // Version
```
jonpas, line 197: C++-style casts?

jonpas (changes requested):

extensions/src/ACRE2Arma/common/wrp/landscape.cpp
TheMagnetar: better not in here.

jonpas (approved):

PR #689, opened 2019-06-04
Babel Notifications

### LINKED ISSUE #657: Language notification
With Babel there is no way to know what language someone is speaking. In reality you can easily tell what language someone is speaking. This is also an issue when someone else switches between languages that the player knows. The player will have no idea that they have switched languages.

I think there should be a little notification saying "Brett is speaking German" when someone either switches languages or you encounter someone for the first time not speaking your default language.

BrettMayson:
- Close #657 

There are going to be some limitation due to where the processing is done and not having enough info in SQF. This seems to be a pretty good setup. https://youtu.be/sIlVnOP0fpI Short demo but I've since switched the player name and language name.

### DIFF
--- a/addons/sys_core/fnc_remoteStartSpeaking.sqf
+++ b/addons/sys_core/fnc_remoteStartSpeaking.sqf
@@ -83,8 +83,8 @@ private _result = false;
 
     _unit setVariable [QGVAR(ts3id), _speakingId];
 
-    private _previousLanguage = _unit getVariable [QGVAR(languageId), -1];
     if (_languageId in (call FUNC(getSpokenLanguages))) then {
+        private _previousLanguage = _unit getVariable [QGVAR(languageId), -1];
         if (_previousLanguage == -1) then {
             if (ACRE_CURRENT_LANGUAGE_ID != _previousLanguage) then {
                 // notify if a person is speaking a different language than the player on first encounter
--- a/addons/sys_core/fnc_remoteStartSpeaking.sqf
+++ b/addons/sys_core/fnc_remoteStartSpeaking.sqf
@@ -86,7 +86,7 @@ private _result = false;
     if (_languageId in (call FUNC(getSpokenLanguages))) then {
         private _previousLanguage = _unit getVariable [QGVAR(languageId), -1];
         if (_previousLanguage == -1) then {
-            if (ACRE_CURRENT_LANGUAGE_ID != _previousLanguage) then {
+            if (ACRE_CURRENT_LANGUAGE_ID != _languageId) then {
                 // notify if a person is speaking a different language than the player on first encounter
                 private _language = GVAR(languages) select _languageId;
                 _language params ["","_languageName"];
--- a/addons/sys_core/fnc_remoteStartSpeaking.sqf
+++ b/addons/sys_core/fnc_remoteStartSpeaking.sqf
@@ -82,7 +82,27 @@ private _result = false;
     };
 
     _unit setVariable [QGVAR(ts3id), _speakingId];
+
+    private _previousLanguage = _unit getVariable [QGVAR(languageId), -1];
+    if (_languageId in (call FUNC(getSpokenLanguages))) then {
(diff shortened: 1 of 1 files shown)


addons/sys_core/fnc_remoteStartSpeaking.sqf
```diff
@@ -82,7 +82,27 @@ private _result = false;
     };
 
     _unit setVariable [QGVAR(ts3id), _speakingId];
+
+    if (_languageId in (call FUNC(getSpokenLanguages))) then {
+        private _previousLanguage = _unit getVariable [QGVAR(languageId), -1];
+        if (_previousLanguage == -1) then {
+            if (ACRE_CURRENT_LANGUAGE_ID != _previousLanguage) then {
```
TheMagnetar: Isn't this redundant? Can ACRE_CURRENT_LANGUAGE_ID  be - 1? Initial value is 0

BrettMayson: On line 87 it uses -1 as the default value, this will indicate that this is the first time this person has been encountered with voice

BrettMayson: Oh nevermind I see what you mean, yes that is out of place

BrettMayson:
I still want to do a little bit more testing, but this is basically final.

Drofseh:
NGL I'm not in love with how this looks.

How will it work when there are multiple people talking, or if you are also talking on the radio at the same time as listening to someone else?

I'd prefer if it showed the language above the players head, similar to how the name/rank/speech indicator are.

Also, correct me if I'm blind, but I don't see a setting to enable/disable it.

BrettMayson:
This will only appear when they speak a different language than when you last heard them, or if you are speaking a different language than them on the first encounter. Not every time they talk.

TheMagnetar:
This works well for close players. However, you get  notifications as well when the player is also far away. Additionally, I  am not sure if this is going to be good for TvT. You will get constant notifications each time a different player speaks.

What  Drofseh proposes would address both issues.

BrettMayson:
I would rather fix the distance issue than switch to the language above their head.

You may not always be able to see the person talking to you. It will only pop up when someone switches languages so you will not get notifications each time someone talks

TheMagnetar:
This will still show a hint when a person is whispering but is still within MAX_DIRECT_RANGE. I am mostly concerned of breaking TvT in here.

BrettMayson:
I'd still like to look at this, it's something my community would still like.

As for TvT concerns, I have some ideas.

Options for when to notify:
- Always
- If the player speaks the language being heard (Default)
- Never

Who to notify for:
- Everyone
- Same Side (Default)
- Same Group

Range slider from 1m to MAX_DIRECT_RANGE that is default to 10m, could even scale with volume, whisper would be `range / 2`, yelling would be `range * 2`

I think that would make it configurable enough that groups could get a setting they are happy with, or disable it outright.

TheMagnetar:
Your ideas make sense on TvT. @jonpas any other suggestions?

jonpas:
Yeah that sounds perfect to me really.

jonpas:
Progress? :D

BrettMayson:
I do want to finish this, it's just been low on the todo list, after HEMTT...

jonpas:
:+1: HEMTT first, absolutely. :P

jonpas:
No further activity and interest, feel free to re-open though.

mrschick:
Another way to make spoken languages easily recognizable could be a voice pitcher.

Since I guess that in the vast majority of missions a translator wouldn't speak more than 3 languages, his "native" language (the first langID in his spokenLanguages Array) would nave normal pitch, the second higher and third lower.

When cycling spoken languages via AltGr, the info text in the bottom right could indicate "normal/higher/lower pitch" below the language name, to allow the player to always look up how each of his spoken languages should sound like.

I'm interested in implementing this but don't possess the TS3 scripting notions that are most likely required for it, *yet*.

PR #697, opened 2019-06-20
ACRE2Shared cleaning

TheMagnetar:
- Cleaning of  ACRE2Shared
  - Remove dependency of  MSVC concurrent_queue. All its use was thread locked so no performance loss
  - Remove of Windows specific time getters and use chrono  instead.
  - Remove  WriteFile and CriticalSection in Log and use C++ native functions.
- ACRE2TS and distortion plugin: Remove dependency of  MSVC concurrent_queue since it was not used at  all


kymckay:
Admittedly I have next to no C++ experience, but rather than checking conditions to return booleans, couldn't you just return one big boolean conditional?

TheMagnetar:
You would have to still go through all the conditions and code is easier  to  understand in this way.

TheMagnetar:
This is ready for review

jonpas (approved):

NouberNou (approved):
Looks good, pending testing.

NouberNou:
> Admittedly I have next to no C++ experience, but rather than checking conditions to return booleans, couldn't you just return one big boolean conditional?

It'd look the same to the compiler, this is just a more explicit and cleaner way to write a very long `&&`ed together boolean statement.

TheMagnetar:
Tested locally @NouberNou. I need a partner for that.

PR #700, opened 2019-06-23
Improve TeamSpeak connection notifications scale

jonpas:
- Prevent notification from going off-screen and being too big.


TheMagnetar (approved):

TheMagnetar:
The only question I have is if you have tested it with different resolutions

jonpas:
Yes, specifically on an ultra-wide. Scaling used in this PR is the same as CBA example for slightly bigger text, which I think fits perfectly for this.

Could use improvements on CBA side, I already noted it to commy2.

PR #707, opened 2019-07-07
Pass on SEM70

TheMagnetar:
- Lazy evaluation
- Optimisations

### DIFF
--- a/addons/sys_sem70/functions/fnc_clearDisplay.sqf
+++ b/addons/sys_sem70/functions/fnc_clearDisplay.sqf
@@ -21,4 +21,4 @@
 params ["_display"];
  {
     RADIO_CTRL(_x) ctrlSetText "";
-} forEach [301,302,303,304,305/*,109*/]; // purge.
+} forEach [301, 302, 303, 304, 305]; // purge.
--- a/addons/sys_sem70/radio/fnc_openGui.sqf
+++ b/addons/sys_sem70/radio/fnc_openGui.sqf
@@ -41,7 +41,7 @@
 params ["_radioId", "", "", "", ""];
 
 // Prevent radio from being opened if it is externally used or it is not accessible
-if (!([_radioId] call EFUNC(sys_radio,canOpenRadio))) exitWith { false };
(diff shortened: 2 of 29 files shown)


addons/sys_sem70/functions/fnc_clearDisplay.sqf
```diff
@@ -21,4 +21,4 @@
 params ["_display"];
  {
     RADIO_CTRL(_x) ctrlSetText "";
-} forEach [301,302,303,304,305/*,109*/]; // purge.
+} forEach [301, 302, 303, 304, 305]; // purge.
```
jonpas, line 24: That should really be a define in `script_component.hpp`. But all radio components could benefit from those.

addons/sys_sem70/radio/fnc_openGui.sqf
```diff
@@ -56,15 +56,15 @@ TRACE_2("OpenGui",GVAR(currentRadioId),GVAR(lastAction));
...
-    _input params ["_radioId","_open"];
-    if (_open) then { _input set [2,GVAR(lastAction)]; };
+    if (GVAR(currentRadioId) isEqualTo -1) then {_input set [1, false]}; // Remove PFH on exit.
+    _input params ["_radioId", "_open"];
+    if (_open) then { _input set [2, GVAR(lastAction)]; };
     private _lastAction = _input select 2;
 
-    if (_lastAction+5 < time) then {
-        if (GVAR(backlightOn) && !GVAR(displayButtonPressed)) then {
+    if ((_lastAction + 5) < time) then {
```
jonpas, line 66:
```suggestion
    if (_lastAction + 5 < time) then {
```
Isn't this more readable and how you would write it in C/C++?

TheMagnetar: In C++ this would be written in brackets (MISRA standard)

addons/sys_sem70/functions/fnc_clearDisplay.sqf
TheMagnetar: Let's have a complete clean of those in the future and make it common to all radios.

jonpas (approved):

PR #715, opened 2019-07-12
Update Load Order for 1.94 Contact DLC

TheMagnetar:
- Load order for contact DLC.

**Merge after 1.94 releases!**


jonpas (changes requested):
Needs required Arma 3 version increase to 1.94 as well.

jonpas (approved):

PR #717, opened 2019-07-12
Convert acre::Vector3 to a template

TheMagnetar:
- Title
- This is needed for later on.

### DIFF
--- a/extensions/src/ACRE2Core/Player.h
+++ b/extensions/src/ACRE2Core/Player.h
@@ -25,8 +25,8 @@ public:
     };
 
     DECLARE_MEMBER(acre::id_t, Id);
-    DECLARE_MEMBER(acre::Vector3_t, WorldPosition);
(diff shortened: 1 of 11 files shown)


jonpas (approved):

NouberNou (changes requested):
We should pull up the specific type of the template into a typedef to avoid having to type the type out. That way if we ever wanted to use double precision (for some reason) we would just need to change the typedef.

TheMagnetar:
What do you think of the following types?

```
using V3Float32_t = Vector3<float32_t>;
using V3Float64_t = Vector3<float64_t>;
```

NouberNou:
Maybe vec3_float32 or vec3_fp32? v3 seems to ambiguous. I think it is more important to know the type of variable vs. it's precision of the constituent values.

TheMagnetar:
I will go for vec3_fp32_t and vec3_fp64_t

NouberNou (approved):
Looks good.

PR #723, opened 2019-07-23
Mass replace __inline with inline

TheMagnetar:
- Mass replace __inline MSVC keyword with the C++ standard inline


jonpas (approved):

jonpas:
Merge after 2.7.0.

PR #725, opened 2019-07-23
Zeus - start speaking on PTT

### LINKED ISSUE #712: Zeus toggle for talking source
Toggle for talking from camera of Zeus would be nice instead or in addition to current PTT button, which is problematic if your TS already has PTT and you don't want to bind 2 buttons in TeamSpeak (or alternatively hold 2 buttons to talk).

Alternatively we could possibly do it like for radios, automatically enable TeamSpeak PTT when holding down "talk from camera" PTT.

BrettMayson:
- Close #712


extensions/src/ACRE2Core/stopZeusSpeaking.h
```diff
@@ -0,0 +1,28 @@
...
+#include "Macros.h"
+#include "Log.h"
+#include "IRpcFunction.h"
+
+#include "IServer.h"
+#include "Engine.h"
+
+#include "TextMessage.h"
+
+#include <string>
```
BrettMayson: I used `stopRadioSpeaking` and `stopIntercomSpeaking` as the base for this file, but as far as I can tell there is no reason for it to be in any of them.

addons/sys_zeus/fnc_handleZeusSpeakPressUp.sqf
```diff
@@ -18,6 +18,9 @@
 [{!(GVAR(keyDownWait))}, {
     player setVariable [QGVAR(inZeus), false, true];
 
+    // Stop speaking
```
TheMagnetar, line 24: Have a look at how MultiPTT are handled. It may solve your issue: https://github.com/IDI-Systems/acre2/blob/master/addons/sys_core/fnc_handleMultiPttKeyPress.sqf#L22

Otherwise this looks good.

jonpas: Yeah, should be handled the same way, but can be done in separate PR.

jonpas:
Tested - did not function.

addons/sys_zeus/fnc_handleZeusSpeakPressUp.sqf
TheMagnetar: I would rather not. It should be something handled in this PR

jonpas: Let's improve it then @synixebrett. :D

jonpas:
Works well now. Can you possibly improve on above comment?

BrettMayson:
Needs MP testing with latest change

TheMagnetar:
This looks fine. Testing in MP  is all is left.

BrettMayson:
Tested in MP, all good

TheMagnetar (approved):

jonpas (approved):

PR #731, opened 2019-07-24
Default radio volume

### LINKED ISSUE #730: Default radio volume setting
ACRE2's default radio volumes at 100% are a little loud for me, so I commonly turn my AN/PRC-343 and/or 152 to ~60%-80% on mission start. I know I'm not the only one that does this, either, as some people like having the extra overhead to turn up to in case things suddenly get loud. 

TFAR has had this option for a while. It's very convenient and not at all immersion-breaking to assume I would always lower the knob to 60%-80%, and we already have busywork-prevention features like automatic antennae adjustment.

BrettMayson:
- Close #730 

~Doesn't work with the SEM 52 SL~

### DIFF
--- a/addons/sys_core/initSettings.sqf
+++ b/addons/sys_core/initSettings.sqf
@@ -19,6 +19,14 @@
 ] call CBA_Settings_fnc_init;
 
 [
+    QGVAR(defaultRadioVolume),
+    "LIST",
+    localize LSTRING(defaultRadioVolume_displayName),
+    "ACRE2",
+    [[0.2, 0.4, 0.6, 0.8, 1], ["20%", "40%", "60%", "80%", "100%"], 3]
+] call CBA_Settings_fnc_init;
+
+[
     QGVAR(spectatorVolume),
     "SLIDER",
     localize LSTRING(spectatorVolume_displayName),
--- a/addons/sys_core/initSettings.sqf
+++ b/addons/sys_core/initSettings.sqf
@@ -24,7 +24,7 @@
     localize LSTRING(defaultRadioVolume_displayName),
     "ACRE2",
     [[0.2, 0.4, 0.6, 0.8, 1], ["20%", "40%", "60%", "80%", "100%"], 3]
-] call CBA_Settings_fnc_init;
+] call CBA_fnc_addSetting;
 
 [
     QGVAR(spectatorVolume),
--- a/addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
+++ b/addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
@@ -77,7 +77,7 @@ for "_i" from 0 to (count _channels)-1 do {
 };
 
 private _volume = EGVAR(sys_core,defaultRadioVolume);
-_volume = 1 max (_volume + .125 - (_volume % .125));
+_volume = _volume + .125 - (_volume % .125);
 
 HASH_SET(_radioData,"volume",_volume);
 HASH_SET(_radioData,"radioOn",1);
--- a/addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
+++ b/addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
@@ -76,6 +76,7 @@ for "_i" from 0 to (count _channels)-1 do {
     PUSH(_currentChannels, _channelData);
 };
 
+// Rounds up 20/40/60/80/100 to the nearest valid value
 private _volume = EGVAR(sys_core,defaultRadioVolume);
 _volume = _volume + .125 - (_volume % .125);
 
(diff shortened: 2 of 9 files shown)


TheMagnetar:
SEM52SL initialises like any other radio.  https://github.com/IDI-Systems/acre2/blob/master/addons/sys_sem52sl/radio/fnc_initializeRadio.sqf#L79

BrettMayson:
I may be wrong, but I don't think the values line up with other radios.

All the others either step by 10 or 20

TheMagnetar:
you are right (knob has 8 positions), but I see bug reports like... hey my volume is not properly set in the sem52sl...

BrettMayson:
Yeah I'm not sure what to do about it, setting it to a value off the knobs could result in weird behaviour when using the knobs. I guess I could just map the value to the closest option for the sem52sl

TheMagnetar:
Defaulting it to the closest value should be the way to go.

BrettMayson:
It doesn't use the closest value, but it rounds up to the nearest value. 80 will go up to 87.5 rather than going down to 75.

jonpas (commented):
Can we macro this for readability and single point of change?

Or comment at least, it's not obvious what the code does and why.

addons/sys_core/initSettings.sqf
```diff
@@ -18,6 +18,14 @@
     {["premixGlobalVolume", _this] call FUNC(setPluginSetting)}
 ] call CBA_Settings_fnc_init;
 
+[
+    QGVAR(defaultRadioVolume),
+    "LIST",
+    localize LSTRING(defaultRadioVolume_displayName),
+    "ACRE2",
+    [[0.2, 0.4, 0.6, 0.8, 1], ["20%", "40%", "60%", "80%", "100%"], 3]
+] call CBA_Settings_fnc_init;
```
jonpas: Should use `CBA_fnc_addSetting`.

addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
```diff
@@ -76,11 +76,14 @@ for "_i" from 0 to (count _channels)-1 do {
     PUSH(_currentChannels, _channelData);
 };
 
-HASH_SET(_radioData,"volume",1);
+private _volume = EGVAR(sys_core,defaultRadioVolume);
+_volume = _volume + .125 - (_volume % .125);
```
jonpas: Comment what this does, or macro with self-explanatory name.

jonpas (changes requested):

jonpas (approved):

addons/sys_sem70/radio/fnc_initializeRadio.sqf
```diff
@@ -100,7 +100,7 @@ HASH_SET(_radioData,"squelch",0);
 // Knobs
 HASH_SET(_radioData,"mainKnobPosition",2); // High Power Setting
 HASH_SET(_radioData,"functionKnobPosition",2); // Manual Frequency Selection (no Relais)
-HASH_SET(_radioData,"volumeKnobPosition", 5);
+HASH_SET(_radioData,"volumeKnobPosition", EGVAR(sys_core,defaultRadioVolume)*5);
```
jonpas:
```suggestion
HASH_SET(_radioData,"volumeKnobPosition",EGVAR(sys_core,defaultRadioVolume) * 5);
```

jonpas (changes requested):

addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
```diff
@@ -76,11 +76,15 @@ for "_i" from 0 to (count _channels)-1 do {
...
+// Rounds up 20/40/60/80/100 to the nearest valid value
+private _volume = EGVAR(sys_core,defaultRadioVolume);
+_volume = _volume + .125 - (_volume % .125);
+
+HASH_SET(_radioData,"volume",_volume);
 HASH_SET(_radioData,"radioOn",1);
 HASH_SET(_radioData,"currentChannel",0);
 HASH_SET(_radioData,"channelKnobPosition", 2); // Channel 1 (after on/off options)
-HASH_SET(_radioData,"volumeKnobPosition", 0);// mid-way
+HASH_SET(_radioData,"volumeKnobPosition", _volume / .125);
```
jonpas:
```suggestion
HASH_SET(_radioData,"volumeKnobPosition", _volume / 0.125);
```

addons/sys_sem52sl/radio/fnc_initializeRadio.sqf
```diff
@@ -76,11 +76,15 @@ for "_i" from 0 to (count _channels)-1 do {
     PUSH(_currentChannels, _channelData);
 };
 
-HASH_SET(_radioData,"volume",1);
+// Rounds up 20/40/60/80/100 to the nearest valid value
+private _volume = EGVAR(sys_core,defaultRadioVolume);
+_volume = _volume + .125 - (_volume % .125);
```
jonpas:
```suggestion
_volume = _volume + 0.125 - (_volume % 0.125);
```

jonpas (changes requested):

jonpas:
Works well.

jonpas (approved):

PR #736, opened 2019-07-24
Use the bent antenna icon when prone or crouch in automatic mode

TheMagnetar:
Right now, the antenna icon does not get updated  in automatic mode, showing always the "straight antenna" when stance is "standing" despite being  crouch  or prone. With this PR

- Use the bent antenna icon when prone or crouch in automatic mode
- When standing, show the straight icon.

image


Soldia1138:
IIRC the intention was that the icon is not shown in automatic mode at all. But I'm ok with this solution to fix the bug I introduced 😁

Soldia1138 (approved):

PR #738, opened 2019-07-26
Rewrite volume control

mharis001:
- Rewrite the volume control GUI
    - Uses a progress bar instead of a slider
    - New color scale: yellow -> orange -> red
    - Can be repositioned now
- Some general cleanup of the `sys_gui` component
    - Remove unused include files `MyDialogDefines.hpp` and `script_strings.hpp`

<details>
<summary>Images</summary>

screen_2

volume_control_comparison

</details>


mharis001:
Not tested in MP and I think I changed all of the volume control things in other components.

~~Can also increase the number of steps further by changing the `VOLUME_LEVEL_CHANGE` macro.~~

Soldia1138:
Nice addition :)
Regarding the 10 steps: We already had that discussion once or twice (and even had 9 steps at one point). The outcome was that more than 5 steps are too confusing for the majority of players and it's too hard to hear the differences between two adjacent steps. So I would prefer to stay at 5 steps unless we can provide a better and more natural feedback on how far the direct speech is being propagated (it's not that easy IRL btw.)

mharis001:
Changed back to 5 steps. Removed white from the color scale.

TheMagnetar:
Outstanding!

jonpas:
Great!

addons/sys_gui/CfgUIGrids.hpp
```diff
@@ -16,13 +26,19 @@ class CfgUIGrids {
...
             };
         };
-
         class Variables {
+            class GVAR(volumeControl) {
+                displayName = CSTRING(VolumeControlGrid_DisplayName);
+                description = CSTRING(VolumeControlGrid_Description);
+                preview = QPATHTOF(data\ui\volume_control_preview.paa);
+                saveToProfile[] = {0, 1};
+                canResize = 0;
```
jonpas:
```suggestion
                canResize = 1;
```
Not sure if anything else has to be changed.

It is by default much smaller now, is that by design?

jonpas (changes requested):

addons/sys_gui/CfgUIGrids.hpp
mharis001: Done. Can increase the default size if you think it should be bigger.

jonpas: I am ok with it being smaller, it's very visible with the rest of the changes. @IDI-Systems/acre-developers what do the rest of you think?

addons/sys_gui/RscTitles.hpp
```diff
@@ -0,0 +1,67 @@
+class RscTitles {
+    #define VOLUME_CONTROL_POSITION \
+        x = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(volumeControl),X)', VOLUME_CONTROL_DEFAULT_X]; \
```
jonpas: Those should be quoted.

addons/sys_gui/RscTitles.hpp
```diff
@@ -0,0 +1,67 @@
...
+        y = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(volumeControl),Y)', VOLUME_CONTROL_DEFAULT_Y]; \
+        w = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(volumeControl),W)', VOLUME_CONTROL_DEFAULT_W]; \
+        h = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(volumeControl),H)', VOLUME_CONTROL_DEFAULT_H] \
```
jonpas: Semi-colon here and not below.

jonpas (changes requested):
A few minor things left.

jonpas (approved):
Looks good to me and works nicely. Requesting review from others as well since it's a big change.

Sniperhid (approved):
LGTM

TheMagnetar (approved):
Outstanding!

NouberNou (approved):

PR #745, opened 2019-07-31
Zeus - Communicate

### LINKED ISSUE #743: ZeusListenViaCamera does not affect listening through Zeus camera
**Arma 3 Version:** `1.94 stable` (stable / rc / dev)
**CBA Version:** `3.12.0` (stable / dev + commit hash)
**ACRE2 Version:** `2.7.0` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- acre2
- a bunch of the other usual stuff
```

**Description:**
- Setting does not affect listener position, always forced to listen through camera
- Setting title and effect is confusing, should consider changing one or the other

While playing as a player with Zeus access, it seems the ZeusListenViaCamera setting does not fully work. The intended functionality, according to Brett, is to toggle whether or not both the emitter and listener position is either the player, or the Zeus camera. In the current implementation, **the setting only toggles the ability to speak through the camera - not listen.**

There may be some missing checks here.
https://github.com/IDI-Systems/acre2/blob/master/addons/sys_core/fnc_processDirectSpeaker.sqf#L51

In addition, I find the setting, or perhaps the entire UX, a bit confusing. If the setting is on, I can only listen through the camera, but I can speak either through my player or through the camera. If the setting is off, I can neither listen nor speak through the camera. It almost seems to make more sense to either split this into two settings, `ZeusListenViaCamera` and `ZeusSpeakViaCamera`, or have communications go either through the camera or through the player. At the very least, the setting description could be changed to make it clear that this affects speaking through the camera as well - perhaps it could be entitled `Communicate through camera` to imply both listening and speaking are affected by this.

Lastly, it would prove very useful, at least to my community, if you could bind a key to toggle between speaking/listening through the camera, and through your player. Often you want to communicate with another Zeus when building something, but may want to jump in and listen to friendlies for a moment.

@synixebrett

BrettMayson:
- Improve the setting for communicate via the Zeus camera
- Close #743

### DIFF
--- a/addons/sys_zeus/initSettings.sqf
+++ b/addons/sys_zeus/initSettings.sqf
@@ -13,11 +13,11 @@ private _zeusCategory = format ["ACRE2 %1", localize "str_a3_cfghints_curator_cu
     ]
 ] call CBA_fnc_addSetting;
 
-// Ability to hear through the Zeus camera
+// Ability to communicate through the Zeus camera
 [
-    QGVAR(zeusListenViaCamera),
+    QGVAR(zeusCommunicateViaCamera),
     "CHECKBOX",
-    LLSTRING(ZeusListenViaCamera_DisplayName),
+    LLSTRING(ZeusCommunicateViaCamera_DisplayName),
     _zeusCategory,
     true
 ] call CBA_fnc_addSetting;
--- a/addons/sys_zeus/stringtable.xml
+++ b/addons/sys_zeus/stringtable.xml
@@ -32,7 +32,7 @@
             <Japanese>Zeus の音源</Japanese>
         </Key>
         <Key ID="STR_ACRE_sys_zeus_zeusCommunicateViaCamera_DisplayName">
-            <English>Communicate through camera</English>
+            <English>Communicate Through Camera</English>
         </Key>
         <Key ID="STR_ACRE_sys_zeus_zeusCommunicateViaCamera_Description">
             <English>Hear and talk with players through the Zeus Interface's camera.</English>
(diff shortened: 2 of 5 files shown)


addons/sys_zeus/initSettings.sqf
```diff
@@ -13,11 +13,11 @@ private _zeusCategory = format ["ACRE2 %1", localize "str_a3_cfghints_curator_cu
     ]
 ] call CBA_fnc_addSetting;
 
-// Ability to hear through the Zeus camera
+// Ability to communicate through the Zeus camera
 [
-    QGVAR(zeusListenViaCamera),
+    QGVAR(zeusCommunicateViaCamera),
```
jonpas, line 18: Note: this will reset the setting for everyone.

BrettMayson: I think it is a significant enough change.

The last setting didn't work as some people assumed it would. In this case since this new setting is more clear I don't think having it reset is a huge issue.

jonpas: We still have to note it in release.

addons/sys_zeus/stringtable.xml
```diff
@@ -31,13 +31,11 @@
             <English>Default Zeus Voice Source</English>
             <Japanese>Zeus の音源</Japanese>
         </Key>
-        <Key ID="STR_ACRE_sys_zeus_ZeusListenViaCamera_DisplayName">
-            <English>Listen through camera</English>
-            <Japanese>カメラを通じて聞く</Japanese>
+        <Key ID="STR_ACRE_sys_zeus_zeusCommunicateViaCamera_DisplayName">
+            <English>Communicate through camera</English>
```
jonpas:
```suggestion
            <English>Communicate Through Camera</English>
```

jonpas (changes requested):

jonpas (approved):

PR #759, opened 2019-08-05
Support reserved keys in radio and intercom dialogs

### LINKED ISSUE #726: Intercom UI keybind doesn't close UI
**Arma 3 Version:** `1.92` (stable)
**CBA Version:** `3.12.0` (stable)
**ACRE2 Version:** `2.7.0` (stable)

**Mods:**
```
- CBA_A3
- acre2
- ace
```

**Description:**
- Intercom UI keybind should also close the UI, as that happens with radios.

**Steps to reproduce:**
- N/A

- Everywhere

**Placed Modules:**
- N/A

**RPT log file:**
- N/A

jonpas:
- Fix #726 - Tab was blocking it as it's a reserved key to switch elements
- Add support for reserved keys in radio and intercom dialogs (eg. Tab)
- Prettify Intercom UI keybind documentation (`<kbd>`)
- Block other keybinds on radio UI open/close (use `openRadio` return value)
- Remove redundant `disableSerialization` - only needed when saving UI elements into variables


addons/sys_list/XEH_postInit.sqf
```diff
@@ -8,6 +8,5 @@ if (!hasInterface) exitWith {};
 }, "", [DIK_CAPSLOCK, [true, false, true]]] call CBA_fnc_addKeybind;
 
 ["ACRE2", "OpenRadio", localize LSTRING(OpenRadio), {
-    [ACRE_ACTIVE_RADIO] call EFUNC(sys_radio,openRadio);
-    false
+    [ACRE_ACTIVE_RADIO] call EFUNC(sys_radio,openRadio)
```
TheMagnetar, line 11: Don't we need a return value for addKeybind?

jonpas: `openRadio` returns `true`, I just made it consistent with other keybinds. Will test either way.

jonpas:
Keybind is being blocked.

TheMagnetar (approved):

PR #760, opened 2019-08-05
Display frequency on single-channel radios in ACE Interact

### LINKED ISSUE #529: Show freq instead of channel for -77 in vehicle rack interaction UI
**Description:**
When interacting with a AN/PRC-77 mounted in AN/VRC-64 show frequency instead of channel number.

Ref: #526 and my discussion with @jonpas on [slack](https://gyazo.com/10acae6ecccac2a6c7019458e29f818a)

jonpas:
- Close #529 
- Remove redundant variable given to `format` in ACE Interact display name
- Fix return value of `getRadioChannel` API
- Shorten `getRadioChannel` API code (less variable assignment)


addons/ace_interact/fnc_radioListChildrenActions.sqf
```diff
@@ -42,12 +42,18 @@ private _radioList = [] call EFUNC(api,getCurrentRadioList);
     };
 
     private _currentChannel = [_x] call EFUNC(api,getRadioChannel);
+    if (/* TODO: max channels == 1*/ false) then {
```
Soldia1138: Example code for that:
`_channelsCount = count ([_radioId, "getState", "channels"] call EFUNC(sys_data,dataEvent)) - 1;`

Source: https://github.com/IDI-Systems/acre2/blob/master/addons/sys_prc343/radio/fnc_setCurrentChannel.sqf#L25

jonpas: Thanks!

jonpas: @Soldia1138 That returns a large number still, looks like AN/PRC-77 counts band channels?

TheMagnetar (approved):

PR #774, opened 2019-08-09
Optimisations in sys_mode

TheMagnetar:
- Some optimisations on sys_mode.

### DIFF
--- a/addons/sys_modes/fnc_sc_muting.sqf
+++ b/addons/sys_modes/fnc_sc_muting.sqf
@@ -24,7 +24,7 @@ private _radioRxData = [_radioId2, "getCurrentChannelData"] call EFUNC(sys_data,
 private _mode1 = HASH_GET(_radioTxData, "mode");
 private _mode2 = HASH_GET(_radioRxData, "mode");
 private _match = false;
-if ((_mode1 == "singleChannel" && {_mode2 == "singleChannel"}) || {_mode1 == "singleChannelPRR" && {_mode2 == "singleChannelPRR"}})) then {
+if ((_mode1 == "singleChannel" && {_mode2 == "singleChannel"}) || {_mode1 == "singleChannelPRR" && {_mode2 == "singleChannelPRR"}}) then {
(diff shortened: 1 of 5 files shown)


addons/sys_modes/fnc_sc_muting.sqf
```diff
@@ -16,18 +16,15 @@
...
 private _radioRxData = [_radioId2, "getCurrentChannelData"] call EFUNC(sys_data,dataEvent);
 
 private _mode1 = HASH_GET(_radioTxData, "mode");
 private _mode2 = HASH_GET(_radioRxData, "mode");
 private _match = false;
-if (
-    (_mode1 == "singleChannel" && _mode2 == "singleChannel") ||
-    (_mode1 == "singleChannelPRR" && _mode2 == "singleChannelPRR")
-) then {
+if ((_mode1 == "singleChannel" && {_mode2 == "singleChannel"}) || {_mode1 == "singleChannelPRR" && {_mode2 == "singleChannelPRR"}}) then {
```
jonpas: Save to 2 variables and compare those.

jonpas (changes requested):

addons/sys_modes/fnc_sc_muting.sqf
TheMagnetar: it will result at the end in the same amount of comparisons.

is mode == signleChannelPRR and mode1 == mode2
and mode == singleChannel and mode1 == mode2

jonpas: Will be more readable, that line is very long.

jonpas:
```suggestion
private _isSingleChannel = _mode1 == "singleChannel" && {_mode2 == "singleChannel"};
private _isSingleChannelPRR = _mode1 == "singleChannelPRR" && {_mode2 == "singleChannelPRR"};
if (_isSingleChannel || {_isSingleChannelPRR}) then {
```

jonpas (approved):

Issue #747, opened 2019-07-31
Intercom UI alignment modified


BizBomb42:
**Arma 3 Version:** `1.94.145903` (stable)
**CBA Version:** `3.12.0` (stable)
**ACRE2 Version:** `2.7.0.1010` (stable)

**Mods:**
```
- CBA_A3
- acre2
- ace (Version 3.12.6 - stable)
- ACE Compat - RHS United States Armed Forces (Version 3.12.6)
- RHSUSAF (Version 0.4.8)
```

**Description:**
- The Intercom UI seen when occupying a vehicle with intercoms equipped has been re-positioned near the middle of the screen (previously positioned in the top-left corner underneath the Vehicle Info UI).
- ACE and RHSUSAF mods used for intercom UI visibility.

![intercom-ui](https://user-images.githubusercontent.com/36508144/62236457-21cf9380-b39d-11e9-929f-3503686fcd32.png)

**Steps to reproduce:**
- Load Arma into the 3den Editor on the Altis terrain.
- Place an M2A3 (class RHS_M2A3_wd) on the terrain with the crew.
- Start a multiplayer game through the 3den Editor, and select a playable character in the vehicle.
- After loading into the player unit, the intercom UI is positioned around the middle of the screen (see attached screenshot) where it was previously positioned in the top-left corner underneath the Vehicle Info UI.

**Where did the issue occur?**
- Editor (Multiplayer)

**Placed Modules:**
- None

**RPT log file:**
- [Client RPT and ACRE Log](https://gist.github.com/BizBomb42/292ef42245ca32d19ca21fd7d58cfe49)

jonpas:
Known, thank you.

BizBomb42:
You're welcome. I didn't see another issue describing this so I wanted to make sure you were aware. I saw the same issue with my mod UI grid because I modeled it after the `grid_ACRE_vehicleInfo` definition utilizing the `profileNamespace` variables.

jonpas:
I wonder if ACE MicroDAGR and other tools have a similar issue (same system)?

BizBomb42:
Appears the MicroDAGR is still in the proper location after a quick test.

![microdagr-centered](https://user-images.githubusercontent.com/36508144/62238300-b25ba300-b3a0-11e9-8d64-cc9b8d5cc03e.png)
[...]

PR #780, opened 2019-08-09
Fix Vehicle Info UI text cut-off and apply retroactive fix

jonpas:
- Fix text width on Vehicle Info UI - was cutting text off right after first intercom name (regression from #764)
- Fix position on broken configurations from <=2.7.0 resulting from changes in Arma 3 v1.94 by resetting broken profile values - #747 
- _Remove intruding space in intercom stringtable_


TheMagnetar (dismissed):

jonpas:
Ok, finally done, I like this solution more even though we save 2 extra things unnecessarily.

TheMagnetar (approved):
Works like a charm

PR #833, opened 2019-09-16
Zeus - Fix Direct Voice Angles 2

BrettMayson:
Electric Boogaloo

Related to #831 cause I messed it up twice as bad as originally thought


BrettMayson:
Actually don't merge this yet, now I'm doubting this whole operation

PR #835, opened 2019-09-17
API - setRadioData & getRadioData

BrettMayson:
- Add `acre_api_fnc_getRadioData`
- Add `acre_api_fnc_setRadioData`

### DIFF
--- a/addons/api/fnc_getRadioData.sqf
+++ b/addons/api/fnc_getRadioData.sqf
@@ -1,0 +1,22 @@
+#include "script_component.hpp"
+/*
+ * Author: ACRE2Team
+ * Get the internal state of a radio
+ *
+ * Arguments:
+ * 0: Radio ID <STRING>
+ *
+ * Return Value:
+ * Internal Radio Data <ARRAY>
+ *
+ * Example:
+ * ["ACRE_PRC148_ID_1"] call acre_api_fnc_getRadioData;
+ *
+ * Public: Yes
+ */
+
+params ["_radioId"];
+
+private _baseRadio = [_radioId] call EFUNC(api,getBaseRadio);
+private _radioData = HASH_GET(EGVAR(sys_data,radioData), _radioId) call EFUNC(sys_data,serialize);
+[_baseRadio, _radioData]
--- a/addons/api/fnc_getRadioData.sqf
+++ b/addons/api/fnc_getRadioData.sqf
@@ -1,13 +1,13 @@
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
- * Get the internal state of a radio
+ * Get the internal state / serialized data of a radio.
  *
  * Arguments:
  * 0: Radio ID <STRING>
  *
  * Return Value:
- * Internal Radio Data <ARRAY>
+ * Serialized Radio Data <ARRAY>
  *
  * Example:
  * ["ACRE_PRC148_ID_1"] call acre_api_fnc_getRadioData;
@@ -18,5 +18,5 @@
 params ["_radioId"];
 
 private _baseRadio = [_radioId] call EFUNC(api,getBaseRadio);
-private _radioData = HASH_GET(EGVAR(sys_data,radioData), _radioId) call EFUNC(sys_data,serialize);
-[_baseRadio, _radioData]
+private _radioDataSerialized = HASH_GET(EGVAR(sys_data,radioData),_radioId) call EFUNC(sys_data,serialize);
+[_baseRadio, _radioDataSerialized]
--- a/addons/api/fnc_setRadioData.sqf
+++ b/addons/api/fnc_setRadioData.sqf
@@ -1,11 +1,11 @@
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
- * Set the internal state of a radio
+ * Set the internal state / de-serialize data of a radio.
  *
  * Arguments:
  * 0: Radio ID <STRING>
- * 1: Radio Data <ARRAY>
+ * 1: Radio Data (Serialized) <ARRAY>
  *
  * Return Value:
  * None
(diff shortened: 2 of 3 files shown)


addons/api/fnc_getRadioData.sqf
```diff
@@ -0,0 +1,22 @@
+#include "script_component.hpp"
+/*
+ * Author: ACRE2Team
+ * Get the internal state of a radio
```
jonpas:
```suggestion
 * Get the internal state / serialized data of a radio.
```

addons/api/fnc_setRadioData.sqf
```diff
@@ -0,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: ACRE2Team
+ * Set the internal state of a radio
```
jonpas:
```suggestion
 * Set the internal state / de-serialize data of a radio.
```

addons/api/fnc_setRadioData.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Arguments:
+ * 0: Radio ID <STRING>
+ * 1: Radio Data <ARRAY>
```
jonpas:
```suggestion
 * 1: Radio Data (Serialized) <ARRAY>
```

addons/api/fnc_setRadioData.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Public: Yes
+ */
+
+params ["_radioId", "_data"];
+_data params ["_radioType", "_radioData"];
+
+private _baseRadio = [_radioId] call EFUNC(api,getBaseRadio);
+if (_baseRadio isEqualTo _radioType) then {
+    HASH_SET(EGVAR(sys_data,radioData), _radioId, _radioData call EFUNC(sys_data,deserialize));
```
jonpas:
```suggestion
    private _radioData = _radioDataSerialized call EFUNC(sys_data,deserialize);
    HASH_SET(EGVAR(sys_data,radioData),_radioId,_radioData);
```

addons/api/fnc_getRadioData.sqf
```diff
@@ -0,0 +1,22 @@
...
+ * Example:
+ * ["ACRE_PRC148_ID_1"] call acre_api_fnc_getRadioData;
+ *
+ * Public: Yes
+ */
+
+params ["_radioId"];
+
+private _baseRadio = [_radioId] call EFUNC(api,getBaseRadio);
+private _radioData = HASH_GET(EGVAR(sys_data,radioData), _radioId) call EFUNC(sys_data,serialize);
```
jonpas:
```suggestion
private _radioDataSerialized = HASH_GET(EGVAR(sys_data,radioData),_radioId) call EFUNC(sys_data,serialize);
```
Better play safe with macros and spaces.

addons/api/fnc_setRadioData.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Example:
+ * private _data = ["ACRE_PRC148_ID_1"] call acre_api_fnc_getRadioData;
+ * ["ACRE_PRC148_ID_2", _data] call acre_api_fnc_setRadioData;
+ *
+ * Public: Yes
+ */
+
+params ["_radioId", "_data"];
+_data params ["_radioType", "_radioData"];
```
jonpas:
```suggestion
_data params ["_radioType", "_radioDataSerialized"];
```

addons/api/fnc_getRadioData.sqf
```diff
@@ -0,0 +1,22 @@
...
+[_baseRadio, _radioData]
```
jonpas:
```suggestion
[_baseRadio, _radioDataSerialized]
```

addons/api/fnc_getRadioData.sqf
```diff
@@ -0,0 +1,22 @@
+#include "script_component.hpp"
+/*
+ * Author: ACRE2Team
+ * Get the internal state of a radio
+ *
+ * Arguments:
+ * 0: Radio ID <STRING>
+ *
+ * Return Value:
+ * Internal Radio Data <ARRAY>
```
jonpas:
```suggestion
 * Serialized Radio Data <ARRAY>
```

jonpas (changes requested):

Sniperhid:
setRadioData is potentially a source of synchronisation issues between clients. The key is to mention that it has to be called globally - otherwise radio details will not be in sync. e.g. radio frequencies for one client may be out causing all sorts of issues. Please update the header to make this clear! (Might even be worth logging it's use should issues begin to appear!)

BrettMayson:
What if we just made the effects global to prevent that issue?

Sniperhid:
> What if we just made the effects global to prevent that issue?

Most API methods in ACRE2 tend to have local effects so for consistency sake the local effect is useful.

BrettMayson:
In this case though would it not be more useful to be global? Since it would be required anyway?

I could add a `_global` param that is true by default so you can set it to default if you do want local only effect for some reason

TheMagnetar (changes requested):

TheMagnetar (approved):

Sniperhid (commented):
I was thinking a bit about this and locality. The global parameter is a good idea, but something more important dawned on me.

Simple overriding the data doesn't quite feel right - This technically allows it to override radio data at any moment even. ACRE has an extensive system for synchronising data and processing events for setting channel data through `sys_data` which houses a fair bit of the code.

I think opting to use events (it would probably require a new event) that sets the data is a better one. It would even allow the interface of the radio to update should the user have the device open/is broadcasting etc.

The major downside to the event approach is there is more code. I think there's two approaches either use the existing events and send a lot of them  - I believe one would be required for each property inside the data as the current internal setData event sets one key/value at a time. The alternative would be to make a new event to set the entire radio data. The event system would also then handle the global transmission of data.

ThymoNL:
Having a function to save and load radio states will allow radio's to be persisted between play sessions on long campaigns.

jonpas:
This should be changed to use data events as @Sniperhid got "dawned on".

jonpas:
No further activity, would have to be re-done with data framework.

PR #846, opened 2019-10-02
Non-Blocking Zeus Speak Keybind

### LINKED ISSUE #838: Volume control on tilde key blocks vanilla keybind (used to work with previous versions)
**Mods (CBA, ACRE2):**
- **Arma 3:** `1.94` 
- **CBA:** `3.12.2` 
- **ACRE2:** `2.7.1.1016` 

**Description:**
I had a strange problem when I updated to latest version of acre, tilde key that I use for both ACRE volume control and AI commanding (select all units - by default I think), stopped working properly.
I can change volume in ACRE, but it doesn't select units anymore.

I checked my keybinds and saw that it (select all units) has been erased, but when I added it back, it didn't work anyway.
When I removed ACRE and tried vanilla only, that keybind still didn't work for selecting all units.
I thought maybe this isn't even ACRE causing it, might be some bug after the 1.94 update.

But, when I restored a backup of my profile, the keybind started working again on vanilla. 
When I added ACRE, it broke again. So I downgraded to v2.6.4.1006 (version I was using before the update to latest) - this one, and versions before all worked fine for me.

**Steps to reproduce:**
Try using tilde for both volume control in ACRE, and select all units (commanding section - vanilla arma).
See if it breaks select all units keybind.

First noticed it on dedicated, but same thing happened later when tested it in editor with AI.

**Log Files:**
I don't have log files here, but if it;s not simple to reproduce, I could get them in a day or two.

BrettMayson:
- Fix #838

### DIFF
--- a/addons/sys_zeus/fnc_handleZeusSpeakPress.sqf
+++ b/addons/sys_zeus/fnc_handleZeusSpeakPress.sqf
@@ -34,4 +34,4 @@ if (call EFUNC(sys_core,inZeus) && {GVAR(zeusCommunicateViaCamera)}) then {
     }, ZEUS_POSITION_FREQUENCY] call CBA_fnc_addPerFrameHandler;
 };
 
-true
+false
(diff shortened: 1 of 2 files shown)


jonpas:
Same in `fnc_handleZeusSpeakPress.sqf`.

BrettMayson:
According to the CBA wiki only the up code matters, don't see why not though

jonpas (approved):

TheMagnetar (approved):

PR #848, opened 2019-10-02
API - Add unit loadout filtering

### LINKED ISSUE #651: Add API function for replacing ID radios with non-ID radios
For use with `get/setUnitLoadout`.

Basically copy this: https://github.com/Theseus-Aegis/Mods/blob/ec647ff24eee0d647bc0062d3f1210823b1af92f/addons/apollo/functions/fnc_playerSingletonSave.sqf#L30-L51

BrettMayson:
- Close #651


jonpas:
Nice! Someone finally copied my code. :D

Let's wait for CBA decision on this before merging until 2.7.2.

jonpas:
I have a recommendation here. Instead of calling `getUnitLoadout` in the function, we should accept a loadout in `unitLoadout` format, modify it and return it. That would allow multiple loadout sources.

BrettMayson:
Perhaps two functions? One that takes a loadout like you describe and one that passes `getUnitLoafout` to that function

jonpas:
I don't really see the point in that when you can just do `[unit, getUnitLoadout unit] call acre_api_fnc_filterUnitLoadout`.

BrettMayson:
Yeah I guess, I'm just lazy and don't like typing

jonpas:
Thoughts on `filterUnitLoadout` (last 3 commits)?

jonpas:
Just need to test call it in-game.

jonpas:
Tested working.

mharis001:
Not necessary but an extension would be to make this function accept all `getUnitLoadout` command parameter types (OBJECT, STRING, CONFIG) as well as a loadout array.

jonpas:
@mharis001 but it already does? Anything in valid `unitLoadout` format can be used, that includes all 4 variants of `getUnitLoadout` command. Only if nothing is passed `getUnitLoadout acre_player` gets used as fall-back.

mharis001:
I meant the function itself would accept the different `getUnitLoadout` parameter types. Something like:

```sqf
params [["_loadout", getUnitLoadout acre_player, [objNull, "", configNull, []]]];

if !(_loadout isEqualType []) then {
    _loadout = getUnitLoadout _loadout;
};
```

Wouldn't add any functionality that doesn't already exist from the user calling `getUnitLoadout` themselves, just a small improvement.

mharis001 (approved):

PR #850, opened 2019-10-04
Revert use of == and use isEqualTo

TheMagnetar:
- Title


jonpas (approved):

jonpas:
Maybe a comment so it doesn't happen again?

PR #857, opened 2019-10-23
Signal Map - Fix marker creating with `_`

TheShinriel:
The problem is when you exec : `[] call acre_sys_signalmap_fnc_open;`
You can set the **TX Position**, but not the **RX Area**.
We have the right error when create a too small area, but just nothing when create any area.

The reason is simple, the functions https://community.bistudio.com/wiki/createMarkerLocal doesn't support underscore anymore in the marker name.

You can try by executing this code :

```sqf
marker1 = createMarkerLocal ["rxarea_0",  position player];
marker1 setMarkerSizeLocal [400,400]; 
marker1 setMarkerShapeLocal "RECTANGLE"; 
marker1 setMarkerColorLocal "ColorRed";
```
Will return a marker : **marker1** == ""
But no error, because thx Bohemia.

But you can create a marker "**rxarea0**"
You can pass empty string... it works... like it is used for rxAreaStartMarker, txPosMarker in sys_signalmap. But it is not possible to use both of them, so no error.

### DIFF
--- a/addons/sys_signalmap/fnc_setRxAreaEnd.sqf
+++ b/addons/sys_signalmap/fnc_setRxAreaEnd.sqf
@@ -75,7 +75,7 @@ if (_this select 1 == 0) then {
 
                 private _markerPos = [(_start select 0) + ((TILE_SIZE * _xScale) / 2), (_start select 1) + ((TILE_SIZE * _yScale) / 2)];
 
-                private _marker = createMarkerLocal [format["rxarea_%1", (count GVAR(rxAreas))], _markerPos];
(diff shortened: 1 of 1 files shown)


TheMagnetar (commented):
At the moment the module is broken and there is a branch dealing with it. If you have interest in fixing signalmap, you can have a look at it.

TheShinriel:
Thx for your response, I may have a look.
Just need to rebuild all tools to compile the projet :)

jonpas:
Signal map fixed by #925.

jonpas (approved):

PR #885, opened 2020-02-15
Extensions - Fix variable shadowing

TheMagnetar:
- Try to fix Multipath issues.

### DIFF
--- a/extensions/src/ACRE2Arma/signal/models/los_simple.cpp
+++ b/extensions/src/ACRE2Arma/signal/models/los_simple.cpp
@@ -5,8 +5,8 @@
 
 #include <cmath>
 
-static const float32_t PI = 3.14159265f;
-static const float32_t magic = 10.0f;
+static constexpr float32_t PI = 3.14159265f;
+static constexpr float32_t magic = 10.0f;
 
 acre::signal::model::los_simple::los_simple(map_p map_) : SignalModel() {
     _map = map_;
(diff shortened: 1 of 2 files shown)


jonpas:
Does **not** fix signal calculation issues on LOS Multi-Path. Repeated same test as we did without the fix, distance up to 2km and over the hill, no matter the distance, 343 stays at -35 dB, 152 at -18 dB.

LOS Simple still works better, but still at -81 dB at 2km.

jonpas:
Confirmed again, 343 works even at 10km behind smaller hills (cut-off over large hills works fine).

jonpas:
863e92e8273e4853aa9d567acdf6be76ae5e98d3 broken (1007+ - includes ACRE2Shared cleanup)
33105a1b1328d33873971b9d6dfdcc112705e126 broken (1007+)
4a88705af056624c4a4db95954e04f3d23b09ea4 broken (1007)
0cc8fbf98edd6db2f28861e208872820867e9bbf works (2.6.4.1006 release)

There are only 2 commits between 1006 and 1007 builds:
- eb8ee2035aa39a87848da818908f08a36c904314 unlikely this breaks anything
- 569e64c5448b28a1c5bf89b393492e8f8ea4e6d9 a lot more likely this does!

jonpas:
There are more commits from 2.6.0 to 2.7.0 in 1007 that were not part of hotfixes due to branch diversion.

- Longley-Rice is NOT the cause
- FakeWRP new program is NOT the cause

Investigation continues...

jonpas:
7ab7ba89c7815a6e1e3e4c169246dc34867edb0c still broken

Before that requires TS roll-back.

Soldia1138:
@TheMagnetar Do you want it merged even if #925 fixes the issue?

jonpas:
Title needs to be changed before merge.

jonpas (approved):

Issue #870, opened 2019-11-28
Zeus cannot hear both remote and local sources at the same time


BojanOnTheInternet:
**Mods (complete and add to the following information):**
- **Arma 3:** `latest stable`
- **CBA:** `latest stable`
- **ACRE2:** `v2.7.2.1022`
- Several others, but none have changed recently except ACRE

**Description:**
Previously a Zeus player would be able to hear remote sources (through camera) as well as local (through unit) at the same time. (I'm unsure if I had checked off `Communicate Through Camera`).

Following the last update, Zeus can only hear through remote if `Communicate Through Camera` is selected and only through unit if it is not selected.

**Steps to reproduce:**
- Enter Zeus, fly far away from yourself
- Have somebody talk next to your unit and next to your camera
- You will only hear one of them, depending on your setting

**Expected behavior:**
A Zeus player should hear both remote and local sources at the same time if `Communicate Through Camera` is enabled. 

**Where did the issue occur?**
- Dedicated

JamesTheClarke:
I have a suspicion that this may be an intended feature rather than a bug. Meaning you have the option to choose either local hearing or camera hearing but not both. Some may prefer the split but it would be neat if there was a third option which would enable both simultaneously for those who want that (I'd count myself to that group).

bux:
Hearing both would be very welcome.

BrettMayson:
This is a bug and will be fixed, previously both were heard

jonpas:
@synixebrett presumably #834 broke it? There have not been other large changes to Zeus from 2.7.1 to 2.7.2

PR #886, opened 2020-03-10
Update fnc_remoteStartSpeaking.sqf

PabstMirror:
Not tested, but I believe this was the intent


Soldia1138 (approved):

jonpas:
Should fix #870?

PabstMirror:
I don't think so, this should only be a trivial performance improvement when searching for nearby radios on External Audio

jonpas (approved):

PR #897, opened 2020-04-05
GUI - Hide antenna display in cargo seat or no radio

PabstMirror:
- Fix  - `stance player` will be `"undefined"` when in vehicles
- Hide antenna display when player has no radios

### DIFF
--- a/addons/sys_components/fnc_getAntennaDirMan.sqf
+++ b/addons/sys_components/fnc_getAntennaDirMan.sqf
@@ -1,3 +1,4 @@
+#include "script_component.hpp"
 /*
  * Author: ACRE2Team
  * Calculates the antenna direction
@@ -13,7 +14,6 @@
  *
  * Public: No
  */
-#include "script_component.hpp"
 
 private ["_forwardV", "_upV"];
 params ["_obj"];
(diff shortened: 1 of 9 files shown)


jonpas (approved):

Drofseh:
Could you also hide it when the player has no radio plz

PabstMirror:
I think that would be fixed by also exiting if `{EGVAR(sys_radio,oldUniqueItemList) isEqualTo []}` but I'm not positive if that's safe

Drofseh:
`{call acre_api_fnc_getCurrentRadio == ""}` should also do it

jonpas:
What @Drofseh said, but without using the API function, those have extra safeties that we don't need to use, just using `ACRE_ACTIVE_RADIO == ""` would work, as it is set to `""` on `preInit` and nothing should ever set it to `nil`.

jonpas:
Found an issue with this, does not reset when getting out of vehicle, stays hidden until user changes antenna position.

jonpas:
Works nicely now, @PabstMirror if you want to give it another test, otherwise we can merge.

PabstMirror:
Only because it's a everyFrame PFEH I switched it to a macro so we can skip 2 `params`

jonpas (approved):

PR #903, opened 2020-04-08
GUI - Fix Inventory CBA compatibility and Vehicle Info seat changing

### LINKED ISSUE #871: Vehicle Info / Comms Bar does not display when boarding a vehicle from non-crew seats
- **Arma 3:** `1.96`
- **CBA:** `3.12.2`
- **ACRE2:** `2.7.2.1022`
Reproduced with only ARMA 3, CBA 3.12.2, and ACRE2 2.7.2.1022

**Description:**
The ACRE 'Vehicle Info' Comms Bar doesn't show up consistently, even when you have access to  intercom. To be clear, I'm talking about the HUD element show in this screenshot:
image

**Steps to reproduce:**
- Spawn any vehicle that has passenger intercom or rack.
- Jump into a passenger seat of a vehicle with at least one intercom.
- Check if you can see the ACRE2 Vehicle Info Comms Bar (you should not see it)
- Get out of the vehicle, and get into any CREW position of that vehicle (driver, gunner, commander, copilot, etc)
- Observe that you can see the ACRE2 Comms Bar (you should now see it)
- Use the scrollwheel menu OR ACE Interact to change vehicle seats to a passenger seat
- Observe that you can still see the ACRE2 Comms Bar
- Do the same in reverse, go to a passenger seat and switch to a crew seat while staying inside of the vehicle via scrollwheel / ACE.
- Observe that you will NOT see the comms bar.

**Expected behavior:**
- The ACRE2 Vehicle Info Comms Bar should show up if you can access any intercom or rack in the vehicle (provided you have access to that intercom or rack from your seat), even if you are a passenger.

- Occurs in Dedicated Multiplayer
- Occurs in Editor Singleplayer / Multiplayer

**Log Files:**
- This issue does not generate any error messages in the .rpt, and it does not cause any script error pop-up messages either.

**Additional context:**
- From my testing, the Comms Bar always shows up if you're in a crew seat. It never shows up if you're in a passenger cargo seat, or a passenger FFV seat.
- I used the FV-720 Mora (AKA the FV510 Warrior if you're using ACE3's realistic names).
- Here is the ACRE2 config I used for that vehicle in this bug report:
```
class AcreIntercoms
{
    class Intercom_1
    {
        allowedPositions[] = {"crew"};
        connectedByDefault = 1;
        displayName = "Crew Intercom";
        limitedPositions[] = {["cargo","all"]};
        masterPositions[] = {"commander"};
        numLimitedPositions = 1;
        shortName = "Crew";
    };
    class Intercom_2
    {
        allowedPositions[] = {"crew",["cargo","all"]};
        connectedByDefault = 0;
        displayName = "Pax Intercom";
        limitedPositions[] = {};
        masterPositions[] = {"commander"};
        numLimitedPositions = 0;
        shortName = "Pax";
    };
};
class AcreRacks
{
    allowedPositions[] = {"driver","gunner","commander","copilot"};
    componentName = "ACRE_VRC103";
    disabledPositions[] = {};
    displayName = "Radio Rack 1";
    intercom[] = {"intercom_1"};
    isRadioRemovable = 0;
    mountedRadio = "ACRE_PRC117F";
    shortName = "R.1";
};
```

**Screenshots:**
As described, here's me jumping into a CREW seat. I can see the Comms Bar.
image

If I use ACE or the scroll wheel menu to change to the passenger seat from the crew seat, you can still see the Comms Bar.
image

If I jump into the vehicle from a passenger seat, I get no Comms Bar.
image

If I switch to a crew seat from the passenger seat, I still have no Comms Bar.
image

### LINKED ISSUE #901: ACRE2 interference with CBA Item Context Menu Framework
**Mods: CBA, ACRE2**
- **Arma 3: 1.96 stable**
- **CBA: 3.15.0 stable**
- **ACRE2: 2.7.2 stable**

**Description:**
ACREs system to be able to double click to open radios in the inventory seems to interfere with CBAs new "Item Context Menu Framework".

https://github.com/CBATeam/CBA_A3/blob/master/addons/ui/fnc_addItemContextMenuOption.sqf

**Steps to reproduce:**
- Go to editor and place player/playable unit.
- Click play
- Execute example given in CBA_fnc_addItemContextMenuOption (Debug console)
```
["#All", "ALL", ">DEBUG ACTION<", nil, nil, {true}, {
	params ["_unit", "_container", "_item", "_slot", "_params"];
	systemChat str [name _unit, typeOf _container, _item, _slot, _params];
	true
}, false, [0,1,2]] call CBA_fnc_addItemContextMenuOption;
```
- Double Click on various items

**Expected behavior:**
While the >DEBUG ACTION< should show up on things like the uniform/vest/backpack and guns themselves, it will not show up on items in uniforms/backpacks/vests. (although it should)

I tried it in:
- Self-Hosted Multiplayer
- Singleplayer
- Editor (Singleplayer)
- Editor (Multiplayer)

jonpas:
- Fix #901 - Replace vanilla inventory click event handlers with CBA Item Context Menu Framework
  - Increase required CBA version to 3.15.0
- Fix #871 - ~~Use `"turret"` CBA Player Event Handler in addition to `"vehicle"`~~ Add proper support for seat switching
  - Including switching between cargo positions
  - Including switching with ACE "Change seat"
  - Rework Vehicle Info UI logic to be event-based with `sys_intercom` as source (`sys_rack` goes through `sys_intercom` anyways)
  - Streamlined CBA Player Event Handlers use, now single `"featureCamera"`, `sys_gui` has no events use anymore
- Improve picking up vanilla `"ItemRadio"` radios in reference to `"ItemRadioAcreFlagged"` being replaced in the hidden radio slot and causing duplicates
- Cleanup GUI component:
  - Switch config XEH to scripted XEH
  - Move `acre_sys_gui_fnc_enterVehicle` definition to its own file
  - Remove unused handlers and functions
  - Organized `postInit` into sections
- Allow `acre_sys_radio_fnc_isUniqueRadio` to work with array argument
- Improve remote control radio usage
  - Add a notification when radios are switched to remote controlled upon opening a radio on a remote controlled unit


jonpas:
Tested and ready.

TheMagnetar (approved):

Soldia1138 (approved):

jonpas:
Need to prevent `acre_sys_intercom_fnc_enterVehicle` from running twice (`"vehicle"` and `"turret"` source). `sys_rack` should already handle that but need to verify.

mharis001 (approved):
Other than needing to handle `vehicle` and `turret` events calling `enterVehicle` twice, code looks good.

jonpas:
Properly fixed #871
- Switching between cargo positions
- Switching with ACE "Change seat"
- Rework Vehicle Info UI logic to be event-based with `sys_intercom` as source (`sys_rack` goes through `sys_intercom` anyways)
- Streamlined CBA Player Event Handlers use, now single `"featureCamera"`, `sys_gui` has no events use anymore

CBA having a `"seat"` Player Event Handler would allow this to be implemented in a much neater way, but we'll need PFHs for status monitoring anyways, so this just makes use of it and streamlines the previous mess of events and use of `seatSwitchedMan` which didn't work for remote control, as well as optimizes various multiple calls to UI updating.

addons/sys_gui/XEH_postInit.sqf
```diff
@@ -3,6 +3,7 @@
 
 if (!hasInterface) exitWith {};
 
+
```
TheMagnetar, line 6: Remove new line?

addons/sys_intercom/XEH_postInit.sqf
```diff
@@ -61,19 +61,16 @@ if (!hasInterface) exitWith {};
...
     params ["_player", "_featureCamera"];
     if (_featureCamera isEqualTo "") then {
-        [FUNC(enterVehicle), [vehicle _player, _player]] call CBA_fnc_execNextFrame; // Make sure vehicle info UI is created
+        [vehicle _player, _player] call FUNC(updateVehicleInfoText); // Update & Show
+    } else {
+        [false] call EFUNC(sys_gui,showVehicleInfo); // Hide
     };
 }, true] call CBA_fnc_addPlayerEventHandler;
 
-player addEventHandler ["seatSwitchedMan", {
```
TheMagnetar, line 71: What happens then when switching seats? How is it captured in cargo seats?

addons/sys_intercom/fnc_enterVehicle.sqf
```diff
@@ -25,20 +25,18 @@ if (_unit != _vehicle) then {
     // Save vehicle
     _unit setVariable [QGVAR(intercomVehicle), _vehicle];
 
-    // Save current seat variable
```
TheMagnetar, line 28: Here the seat information was stored for the unit. Why is this removed? I initially wanted to avoid calling it in a PFH and decided to check it and update it only once whenever it was necessary.

addons/sys_intercom/fnc_seatSwitched.sqf
```diff
@@ -54,11 +56,9 @@ if (GVAR(broadcastKey)) then {
         if (_connectionStatus > INTERCOM_DISCONNECTED) then {
             [_vehicle, _unit, _forEachIndex] call FUNC(setStationUnit);
         };
-        [_vehicle, _unit] call FUNC(updateVehicleInfoText);
```
TheMagnetar, line 57: Here it was called  in order to remove as soon as possible the text in green when using the intercom. Is it still working the same way?

TheMagnetar (changes requested):

addons/sys_gui/XEH_postInit.sqf
jonpas: Split into categories: Inventory, Volume Control and now removed Vehicle Info UI.

addons/sys_intercom/XEH_postInit.sqf
jonpas: Intercom PFH is the source of all information. It has to run anyways, so I capture seat changes there by already existing functions you wrote.

This method was poor, it only works on `player` (no remote control) and only  for vanilla switching (no ACE "Change seat" or scripted). New method captures all of it, including switching between specific cargo seats, to comply with the configuration ability which can do the same.

Reference last commit for that change.

addons/sys_intercom/fnc_enterVehicle.sqf
jonpas: > Here the seat information was stored for the unit. Why is this removed?

This was redundant. It got saved in intercom PFH anyways right after. All it did was run unnecessary procedure in `seatSwitched` which is for invalidation when leaving a certain seat (which can't happen when you enter a vehicle).

> I initially wanted to avoid calling it in a PFH and decided to check it and update it only once whenever it was necessary.

That is not how it worked.

addons/sys_intercom/fnc_seatSwitched.sqf
jonpas: From what I can see, yes. This loop over intercoms is really fast, because there are never a lot of intercoms, so calling it once for all of them at the end is actually less time spent in UI work.

EDIT: Besides, it's unscheduled, so it has to happen in the same frame. Thus no difference if we run it multiple times here or once below, only optimization.

jonpas: It is indeed faster, just tested.

Only not-instant part is switching certain seats due to 1.1s intercom PFH, but those eats cannot be caught by anything else as there are no EHs that work reliably or support all possibilities.

Why is that EH 1.1s btw? 1s would seem more "standard"? :thinking:

addons/sys_intercom/XEH_postInit.sqf
TheMagnetar: is the frequency of the PFH enough or should it run faster?

addons/sys_intercom/fnc_seatSwitched.sqf
TheMagnetar: see above. so 1s is more standard than 0.9s or 0.75s? :D

TheMagnetar (approved):

PR #907, opened 2020-04-09
ACE Interact - Only do a single call to getPlayerRadioList

PabstMirror:
`acre_sys_data_fnc_getPlayerRadioList` has some cost (0.3ms with a few items/radios)

before the func was called twice in a row by both `condition` and `insertChildren`
this skips condition check and just returns zero children if array is empty
ace will skip actions if their statement is empty and they have no children

### DIFF
--- a/addons/ace_interact/fnc_radioListChildrenActions.sqf
+++ b/addons/ace_interact/fnc_radioListChildrenActions.sqf
@@ -1,13 +1,13 @@
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
- * Adds child actions for self interaction
+ * Adds child actions for self-interaction. 
  *
  * Arguments:
  * 0: Interaction Target (player) <OBJECT>
  *
  * Return Value:
- * Ace Child Actions <ARRAY>
+ * ACE Child Actions <ARRAY>
  *
  * Example:
  * [player] call acre_ace_interact_fnc_radioListChildrenActions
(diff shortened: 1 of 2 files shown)


addons/ace_interact/fnc_radioListChildrenActions.sqf
```diff
@@ -1,27 +1,28 @@
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
- * SHORT DESCRIPTION
+ * Adds child actions for self interaction
```
jonpas:
```suggestion
 * Adds child actions for self-interaction. 
```

addons/ace_interact/fnc_radioListChildrenActions.sqf
```diff
@@ -1,27 +1,28 @@
...
  *
  * Arguments:
- * 0: ARGUMENT ONE <TYPE>
- * 1: ARGUMENT TWO <TYPE>
+ * 0: Interaction Target (player) <OBJECT>
  *
  * Return Value:
- * RETURN VALUE <TYPE>
+ * Ace Child Actions <ARRAY>
```
jonpas:
```suggestion
 * ACE Child Actions <ARRAY>
```

jonpas (changes requested):

jonpas (approved):

addons/ace_interact/fnc_radioListChildrenActions.sqf
```diff
@@ -1,27 +1,28 @@
 #include "script_component.hpp"
 /*
  * Author: ACRE2Team
- * SHORT DESCRIPTION
+ * Adds child actions for self-interaction.
```
jonpas:
```suggestion
 * Adds child actions for self-interaction.
```

Issue #638, opened 2019-02-26
Reserved variable error in processRadioSpeaker


jonpas:
**Arma 3 Version:** `1.88` (stable)
**CBA Version:** `3.9.1` (stable)
**ACRE2 Version:** `2.6.2.996` (stable)

**Mods:**
```
- @sweet_markers
- @Ruha
- @jo_ssg
- @CUP_Terrains_Core
- @CUP_Terrains_ACE
- @cTab
- @CBA_A3
- @bwa3_ace
- @bwa3
- @acre2
- @acex
- @ace
```

**Description:**
```
18:49:10 [ACRE] (sys_data) WARNING: Non-existent radio 'any' called getCurrentChannelData radio event! idi\acre\addons\sys_data\fnc_processRadioEvent.sqf:25
18:49:10 Error in expression <> "speaking");
(acre_sys_core_coreCache setVariable [ "okRadios"+_radioId,  _okR>
18:49:10   Error position: <setVariable [ "okRadios"+_radioId,  _okR>
18:49:10   Error Reserved variable in expression
18:49:10 File idi\acre\addons\sys_core\fnc_processRadioSpeaker.sqf, line 43
```

Means `_radioId` is `nil`, because `X setVariable [""+nil, ""]` errors with "Reserved variable in expression".

Was supposed to be fixed by #279 with `getVariable` default value.

**Steps to reproduce:**
- Ǹo clue

**Where did the issue occur?**
- Dedicated

**Placed Modules:**
- Unknown

**RPT log file:**
- https://gist.github.com/commy2/0a2ea41fcc1a2fead5c8f1ddd41e84e1

Reported by @commy2.

Cyruz143:
Had this occur recently on 2.6.2.1001

Mods:
```
@cba_a3;
@acre2;
@ace_ark;
@acex;
@ark;
@ark_inhouse;
@cup_tp;
@CUP_Weapons;
@CUP_Units;
@CUP_Vehicles;
@CUP_Vehicles_ACE_compat;
@CUP_Weapons_ACE_compat
```

Description:
```
23:16:36 [ACRE] (sys_data) WARNING: Non-existent radio 'any' called getCurrentChannelData system event! idi\acre\addons\sys_data\fnc_processSysEvent.sqf:27
23:16:36 [ACRE] (sys_data) WARNING: Non-existent radio 'any' called getCurrentChannelData radio event! idi\acre\addons\sys_data\fnc_processRadioEvent.sqf:25
[...]

PR #909, opened 2020-04-14
Core - Add debug information for reserved variable error

jonpas:
- Debug for #638

### DIFF
--- a/addons/sys_core/fnc_processRadioSpeaker.sqf
+++ b/addons/sys_core/fnc_processRadioSpeaker.sqf
@@ -30,7 +30,7 @@ if (_radioId == "") exitWith { false };
 
 // Debug #638
 if (isNil "_radioId") exitWith {
-    WARNING_1("reserved variable debug 1 - report on GitHub!",_radioId);
+    WARNING_1("reserved variable debug 1 - report on GitHub! [%1]",_radioId);
     false
 };
 
@@ -52,7 +52,7 @@ if (!GVAR(speaking_cache_valid)) then {
 
     // Debug #638
     if (isNil "_radioId") exitWith {
-        WARNING_3("reserved variable debug 2 - report on GitHub!",_radioId,_okRadios,_playerRadios);
+        WARNING_3("reserved variable debug 2 - report on GitHub! [%1-%2-%3]",_radioId,_okRadios,_playerRadios);
         false
     };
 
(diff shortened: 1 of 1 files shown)


addons/sys_core/fnc_processRadioSpeaker.sqf
```diff
@@ -27,6 +27,13 @@ TRACE_2("",_unit,_playerRadios);
 
 private _radioId = _unit getVariable [QGVAR(currentSpeakingRadio), ""];
 if (_radioId == "") exitWith { false };
+
+// Debug #638
+if (isNil "_radioId") exitWith {
+    WARNING_1("reserved variable debug 1 - report on GitHub!",_radioId);
```
PabstMirror:
```suggestion
    WARNING_1("reserved variable debug 1 - report on GitHub! [%1]",_radioId);
```

addons/sys_core/fnc_processRadioSpeaker.sqf
```diff
@@ -43,6 +50,12 @@ if (!GVAR(speaking_cache_valid)) then {
     #endif
     _okRadios = (_okRadios select 0) select 1;
 
+    // Debug #638
+    if (isNil "_radioId") exitWith {
+        WARNING_3("reserved variable debug 2 - report on GitHub!",_radioId,_okRadios,_playerRadios);
```
PabstMirror:
```suggestion
        WARNING_3("reserved variable debug 2 - report on GitHub! [%1-%2-%3]",_radioId,_okRadios,_playerRadios);
```

jonpas:
I always forget this is not the same as `TRACE_n`.

addons/sys_core/fnc_processRadioSpeaker.sqf
```diff
@@ -43,6 +50,12 @@ if (!GVAR(speaking_cache_valid)) then {
     #endif
     _okRadios = (_okRadios select 0) select 1;
 
+    // Debug #638
+    if (isNil "_radioId") exitWith {
```
PabstMirror, line 54: Should this nil check be on `_okRadios`? _radioId can't change from the last check
also exitWith return value will be lost to scope

jonpas: That's what I am trying to figure out, somehow `_radioId` is `nil` in the below `HASH_SET` operation, even though it should be `""` due to `getVariable` default value on `nil`.

Does `setVariable` also throw "reserved variable" error if you try to assign nil to it? I don't think so.

PabstMirror (approved):

PR #917, opened 2020-04-15
Use RVExtensionRegisterCallback

PabstMirror:
Use RVExtensionRegisterCallback so extension can return data when ready instead of running polling loop in-game.

Can drop a PFEH, callExtension and call compile each frame. 
But these really aren't that expensive and this is a delicate area so this may not be worth it.


jonpas:
I think it should still be worth it. It's delicate but easily testable, either data comes through, or it doesn't.

PabstMirror:
This is included in #933

Issue #820, opened 2019-08-30
LOS Multipath line of sight range is too far


BlackAlpha:
**Mods:**
- **Arma 3:** `1.94.145977 (Stable)` 
- **CBA:** `3.12.1 (Steam)`
- **ACRE2:** `v2.7.1.1016 (Steam)`

**Description:**
AN/PRC-343 line of sight range is 2000-2500 meters with LOS Simple. The line of sight range is 5000+ meters with LOS Multipath. In both cases the Terrain Loss Coefficient is set to 1. I've managed to reproduce this various times in actual missions and a test mission (see below).

**Steps to reproduce:**
- Download test mission: https://www.dropbox.com/s/9b4a4grvkromdqs/acretest1.Altis.pbo?dl=0
- Run on dedicated server.
- The mission allows you to teleport to various distances with a clear line of sight to the person up on the hill.
- Try talking to each other from as far away as possible.

**Expected behavior:**
I believe the AN/PRC-343 line of sight range in real life is closer to 500-800 meters.

**Where did the issue occur?**
Dedicated

**Log Files:**
- Server RPT: https://pastebin.com/5nD9ndm3
- Client RPT: https://pastebin.com/BC2RrUdW
- ACRE Log: https://pastebin.com/tgsxR0tW

**Additional context:**
ACRE settings used:
Model Interference: On
Full Duplex Radios: Off
Ignore Antenna Direction: On
Automatic Antenna Direction: Off

TheMagnetar:
Thanks for your detailed bug report. We will have a loot at it asap.

arismagicgr:
I do have the same issue as @BlackAlpha although to be honest, I can not present all these info he did. I can just say that I have almost the same settings with the difference that I have "Ignore Antenna Direction: off". If there is a way to acquire all the info you need, please give me some guidance and I will do my best to follow.

JamesTheClarke:
Our community utilises the same settings as @BlackAlpha with the exception of Model Interference and our results are more or less the same. As long as there is nothing breaking LOS the 353 range is very far indeed. Easy testing site is the Altis desert for example.

I agree with the OP that the 343 should be heavily distorted at around 800m even with LOS being maintained.

TheMagnetar:
At the moment (since 2.7.0) there are some subtle problems on how multipath is being calculated. LOS Simple and ITM should give more accurate results for the time being.

JamesTheClarke:
[...]

PR #925, opened 2020-04-17
Signal - Fix excessive range bug described in #820 in multipath signal model 

NouberNou:
This will fix the bug referenced in #820.

Fix was a typo when moving help functions for converting between dBm and volts. Volts to dBm uses the natural log of the square of the voltage. New code was using the base 10 log, creating an excessively high voltage.


TheMagnetar (approved):

jonpas:
Use something like "fix #820" and it'll automatically close issue ticket. ;)

Issue #276, opened 2017-03-30
Use new alternate callExtension syntax


jonpas:
New syntax since 1.67 (1.68 stable) [`extension callExtension [function, arguments]`](https://community.bistudio.com/wiki/callExtension) improves performance by not needing to separate the string in the extension (minimal) and using `format` in SQF (bigger, but still small). Although mild, this will prove rather a big gain due to the number of extension calls we do.

Additionally, return of the new `callExtension` syntax can be used with [`parseSimpleArray`](https://community.bistudio.com/wiki/parseSimpleArray) for large performance improvements when returning multiple values.

jonpas:
It has been tested that `parseSimpleArray` nests big improvement, but alternate syntax for `callExtension` not so much (not using `format`).

TheMagnetar:
Is this still a thing?

jonpas:
Not really, parseSimpleArray might be faster but it's harder to work with. Let's leave it open and I might take a look some day.

jonpas:
Closed by #933. Remaining extension calls are only occasional or only on load.

PR #933, opened 2020-04-21
Extensions - Improve callExt for getSignal and update dynload

PabstMirror:
Continuation of #917
Tag issue: #276
That would break use of dynload as it didn't support `RVExtensionRegisterCallback`

This PR:
- Update dynload for x64, rvArgs and rvCallback, and fix temp file block
- Use direct callExtensionArgs for `fnc_getSignal`'s call to `fnc_callExt` (~0.13ms to 0.03ms)
The risk is that the game is parsing data to string instead of our code, but I believe it is safe for "process_signal"


addons/sys_core/fnc_callExt.sqf
```diff
@@ -21,6 +21,18 @@
 
 params ["_command", "_params", ["_threaded", false], ["_callBack",{}], ["_callBackArgs",[]]];
 
+if (_threaded isEqualTo 2) exitWith {
```
TheMagnetar, line 24:
```suggestion
if (_threaded == 2) exitWith {
```
isn't this faster?

extensions/src/ACRE2Arma/acre/acre.cpp
```diff
@@ -48,21 +59,23 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function) {
     }
     acre::arguments _args(argument_str);
 
-    std::string result = "-1";
+    std::string result_string = "-1";
+    int result_dummy = 0;
```
TheMagnetar:
```suggestion
    std::int32_t result_dummy = 0;
```

extensions/src/ACRE2Arma/acre/acre.cpp
```diff
@@ -77,9 +90,29 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function) {
         _threaded = true;
     }
 
-    acre::controller::get().call(command, _args, result, _threaded);
-    if (result.length() > 0) {
-        sprintf_s(output, outputSize, "%s", result.c_str());
+    acre::controller::get().call(command, _args, result_string, result_dummy, _threaded);
+    if (result_string.length() > 0) {
```
TheMagnetar:
```suggestion
    if (!result_string.empty()) {
```

extensions/src/ACRE2Arma/acre/acre.cpp
```diff
@@ -77,9 +90,29 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function) {
...
+    ZERO_OUTPUT();
+    if (argc < 2) { EXTENSION_RETURN_VALUE(-2); } // callExtension ["calla", ["extension.dll", "func", ...]]
+
+    std::string command(function);
+    acre::arguments _args(argv, argc);
+
+    //LOG(INFO) << "RVExtensionArgs [" << function << " - " << _args.to_string() << "]";
+    if (command != "process_signal") { EXTENSION_RETURN_VALUE(-3); } // only supported command for now
+
+    std::string result_string = "";
```
TheMagnetar:
```suggestion
    std::string result_string;
```

no need to initialize as an empty string

extensions/src/ACRE2Arma/acre/acre.cpp
```diff
@@ -77,9 +90,29 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function) {
...
+    int result_code = -4;
```
TheMagnetar:
```suggestion
    std::int32_t result_code = -4;
```

extensions/src/ACRE2Arma/common/arguments.hpp
```diff
@@ -39,12 +39,22 @@ namespace acre {
...
     public:
-        arguments(const std::string & str) : _original(str), _internal_index(0) {
+        arguments(const std::string & str) : _internal_index(0) {
             _args = acre::split(str, ',');
             for (size_t i = 0; i < _args.size(); i++) {
                 _args[i] = trim(_args[i]);
             }
         }
+        arguments(const char** argv, int argc) : _internal_index(0) {
+            for (int i = 0; i < argc; i++) {
```
TheMagnetar:
```suggestion
            for (std::int32_t i = 0; i < argc; i++) {
```

addons/sys_core/fnc_callExt.sqf
Drofseh: isEqualTo is faster iirc

extensions/src/ACRE2Arma/common/arguments.hpp
```diff
@@ -72,8 +82,12 @@ namespace acre {
             return acre::vector3<float>(to_float(t[0]), to_float(t[1]), to_float(t[2]));
         }
 
-        const std::string & get() const {
-            return _original;
+        const std::string to_string() const { // For Debugging
+            std::stringstream ss;
+            for (int i = 0; i < _args.size(); i++) {
```
TheMagnetar:
```suggestion
            for (std::uint32_t i = 0U; i < _args.size(); i++) {
```

extensions/src/ACRE2Arma/common/dispatch.hpp
```diff
@@ -32,15 +34,15 @@ namespace acre {
     public:
         dispatcher() : _ready(true) { }
 
-        virtual bool call(const std::string & name_, arguments & args_, std::string & result_) {
+        virtual bool call(const std::string & name_, arguments & args_, std::string & result_string_, int & result_code_) {
```
TheMagnetar:
```suggestion
        virtual bool call(const std::string & name_, arguments & args_, std::string & result_string_, std::int32_t & result_code_) {
```

extensions/src/ACRE2Arma/common/dispatch.hpp
```diff
@@ -142,39 +126,43 @@ namespace acre {
...
+                        ss << "[" << _message.id << ",[" << result_message << "]]";
+
+                        bool cbRecieved = false; // call back buffer can only hold 100 messages a frame (doubt acre will ever hit this limit)
+                        while (!_stop && !cbRecieved) {
+                            int cbBufferRemaining = callbackFunc("ACRE_TR", _message.command.c_str(), ss.str().c_str());
+                            // LOG(TRACE) << "sending callback [id: " << _message.id << ", buffer: " << cbBufferRemaining << "]";
+                            if (cbBufferRemaining > -1) {
+                                cbRecieved = true;
+                            } else {
+                                sleep(5);
```
TheMagnetar: you can also use std::this_thread::sleep_for(xxx) where xxx is in milliseconds

extensions/src/ACRE2Arma/common/dispatch.hpp
```diff
@@ -142,39 +126,43 @@ namespace acre {
...
+                                // LOG(TRACE) << "\t{ " << _message.args.to_string() << " }";
                             }
-                        }
 #endif
-                        dispatcher::call(_message.command, _message.args, result.message);
-                        {
-                            std::lock_guard<std::mutex> lock(_results_lock);
-                            _results.push(result);
+                        std::string result_message;
+                        int result_code = -1;
```
TheMagnetar:
```suggestion
                        std::int32_t result_code = -1;
```

extensions/src/ACRE2Arma/dynload/dynloader.hpp
```diff
@@ -51,7 +58,7 @@ namespace acre {
                 LOG(ERROR) << "GetTempPath() failed, e=" << GetLastError();
                 return false;
             }
-            if (!GetTempFileNameA(tmpPath, "acre_dynload", TRUE, buffer)) {
+            if (!GetTempFileNameA(tmpPath, "acre_dynload", FALSE, buffer)) { // get a unique temp filename each time
```
TheMagnetar:
```suggestion
            if (!GetTempFileNameA(tmpPath, "acre_dynload", false, buffer)) { // get a unique temp filename each time
```

extensions/src/ACRE2Arma/dynload/dynloader.hpp
```diff
@@ -112,26 +144,24 @@ namespace acre {
...
-            //LOG(INFO) << "Calling [" << args_.as_string(0) << "]";
-
+        bool callRV(const arguments & args_, std::string & result_string) {
             if (_modules.find(args_.as_string(0)) == _modules.end()) {
                 return false;
             }
 
-            result = "";
-            result.resize(4096);
+            result_string = "";
```
TheMagnetar:
```suggestion
            result_string.clear();
```

extensions/src/ACRE2Arma/dynload/dynloader.hpp
```diff
@@ -140,6 +170,34 @@ namespace acre {
...
+            std::string extension_file = args_.as_string(0);
+            if (_modules.find(extension_file) == _modules.end()) {
+                LOG(WARNING) << "callRVArgs could not find module [" << extension_file << "]";
+                return false;
+            } else if (!_modules[extension_file].func_rvArgs) {
+                LOG(ERROR) << "module [" << extension_file << "] does not support CallExtensionArgs";
+                return false;
+            }
+
+            result_string = "";
```
TheMagnetar:
```suggestion
            result_string.clear();
```

extensions/src/ACRE2Arma/dynload/dynloader.hpp
```diff
@@ -140,6 +170,34 @@ namespace acre {
...
+
+            std::string function = args_.as_string(1);
+            std::vector<const char*> cstrings;
+            for (int i = 2; i < args_.size(); i++) {
+                cstrings.push_back(args_.as_string(i).c_str());
+            }
+
+            result_code = _modules[extension_file].func_rvArgs((char*)result_string.c_str(), 4096, function.c_str(), cstrings.data(), static_cast<int>(cstrings.size()));
+#ifdef _DEBUG
+            //if (args_.as_string(0) != "ready") {
```
TheMagnetar: this is already inside an ifdef. Either remove the commented code or remove it alltogether

TheMagnetar (changes requested):
I would suggest to change all the new declared variables as `int` to `std::int32_t`

addons/sys_core/fnc_callExt.sqf
TheMagnetar: For comparing integers too?

PabstMirror: the param is overloaded now, was bool, now `true/false/2`

addons/sys_core/XEH_preInit.sqf
```diff
@@ -148,5 +126,21 @@ for "_i" from 1 to (_m/2) do {
...
+    (parseSimpleArray _data) params ["_id", "_args"];
+    TRACE_3("ExtensionCallback",_function,_id,_data);
+
+    private _callBack = GVAR(threadedExtCalls) select _id;
+
+    if (IS_ARRAY(_callBack)) then {
+        if !(_args isEqualTo []) then {
+            _args = _args select 0;
+        };
+        [_callBack select 0, _args] call (_callBack select 1);
```
PabstMirror, line 141:
```suggestion
        [_callBack select 0, _args] call (_callBack select 1);
        GVAR(threadedExtCalls) set [_id, nil];
```
Would it make sense to clear the entry after use?
`count str acre_sys_signal_fnc_handleSignalReturn` is 5368 and the whole func is copied each time
so after several thousand calls this will start to take up a non-trivial amount of memory

could also store ID in each element and pop off when used
```
(GVAR(threadedExtCalls) param [0,[-1]]) params ["_firstId", "_firstArgs", "_firstFunc"];
if (_id != _firstId) exitWith {ERROR...};
GVAR(threadedExtCalls) deleteAt 0;
```

or use the _function param to invoke the handler
```
if (_function == "process_signal") then { ... call EFUNC(...,handleSignalReturn)};
```

TheMagnetar: I don't think this will break anything. But tagging @Soldia1138 and @jonpas for another opinion.

Soldia1138: It should work, yes. Everything's set to missionNamespace within `acre_sys_signal_fnc_handleSignalReturn`

At least we should test it in a smaller scale test on a dedicated

extensions/src/ACRE2Arma/common/dispatch.hpp
```diff
@@ -137,19 +137,19 @@ namespace acre {
...
                         bool callback_recieved = false; // call back buffer can only hold 100 messages a frame (doubt acre will ever hit this limit)
                         while (!_stop && !callback_recieved) {
-                            int callback_buffer_remaining = callbackFunc("ACRE_TR", _message.command.c_str(), ss.str().c_str());
+                            std::int32_t callback_buffer_remaining = callbackFunc("ACRE_TR", _message.command.c_str(), ss.str().c_str());
                             // LOG(TRACE) << "sending callback [id: " << _message.id << ", buffer: " << callback_buffer_remaining << "]";
                             if (callback_buffer_remaining > -1) {
                                 callback_recieved = true;
                             } else {
-                                sleep(5);
+                                std::this_thread::sleep_for(std::chrono::milliseconds(5));
```
TheMagnetar, line 152: The original call `sleep(5)` stops the process for 5 seconds.
```suggestion
                                std::this_thread::sleep_for(std::chrono::seconds(5));
```

TheMagnetar (changes requested):

extensions/src/ACRE2Arma/common/dispatch.hpp
PabstMirror: 5 ms is what I wanted originally
extensions push messages the the callback queue, which only has a max depth of 100
and then every in-game-frame it gets emptied
so all this is trying to do is wait for the next in-game frame

TheMagnetar: Ok. Then it is fine

TheMagnetar (approved):
All issues i had have been addressed.

jonpas (approved):
No issue after 2 session. Working just fine with 19 players.

PR #938, opened 2020-04-26
Debug - Add dumpPerformanceCounters func

PabstMirror:
from ace, which I think got it from acre 😄


jonpas:
Closes #643? :D (never finished it)

PR #940, opened 2020-04-27
Various - Fix undefined functions

PabstMirror:
- ace/components: edit: `GVAR(connectorsEnabled) = false;` so this isn't used
- fullDuplex: api func didn't exist, no bug because everything just uses `GVAR(fullDuplex)`
- rack: still seems to work fine
- handleRadioSpatialKeyPressed: doesn't seem to be used anywhere
- invalidGarbageCollect: was just for returning logging from clients to server

### DIFF
--- a/addons/sys_rack/XEH_postInit.sqf
+++ b/addons/sys_rack/XEH_postInit.sqf
@@ -120,7 +120,7 @@ if (vehicle acre_player != acre_player) then {
         false
     };
 
-    if ([_rackId] call FUNC(getMountedRackRadio) != "") exitWith {
+    if ([_rackId] call EFUNC(api,getMountedRackRadio) != "") exitWith {
         [QGVAR(logOnServer), format ["Rack ID %1 has already a radio mounted.", _rackId]] call CBA_fnc_serverEvent;
         false
     };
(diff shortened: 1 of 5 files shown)


addons/sys_rack/XEH_postInit.sqf
```diff
@@ -120,7 +120,7 @@ if (vehicle acre_player != acre_player) then {
         false
     };
 
-    if ([_rackId] call FUNC(getMountedRackRadio) != "") exitWith {
+    if ([_rackId] call EFUNC(api,getMountedRackRadio) != "") exitWith {
```
jonpas: API function just calls `FUNC(getMountedRadio)` if all tests pass, so this should probably be:
```suggestion
    if ([_rackId] call FUNC(getMountedRadio) != "") exitWith {
```
Ideally components should also never call API component functions, only the other way around.

jonpas (changes requested):
Component stuff is not used anyways, so that code was only partly tested during Rack development.

Otherwise nice finds.

jonpas (approved):

jonpas:
No issue after 2 session.

Issue #629, opened 2019-02-11
Custom PREP macros don't support compile cache


jonpas:
https://github.com/IDI-Systems/acre2/blob/8a48bc44e7b192ba0347438a9e419dea8f398807/addons/main/script_macros.hpp#L127-L130 (not `PREP_FOLDER`)

Should use something like this https://github.com/acemod/ACE3/blob/40f371c0634c44d3ec60b0e44a8351736d7ac53c/addons/main/script_macros.hpp#L99-L105

PR #944, opened 2020-04-28
Radios - Cache module functions (`PREP_MODULE`)

PabstMirror:
Ref #629
I'm not sure about the other macros, 
- many of them define multiple functions from within a single sqf file
- some are a mix of function definitions and `GVAR` setup and init code

combined cba+acre preInit XEH goes from 1900->900 ms
but also means you have to use `DISABLE_COMPILE_CACHE` to see changes on filepatching/dev
already on wiki https://acre2.idi-systems.com/wiki/development/building#disable-cba-function-caching


jonpas:
Yees! I wanted to do that a while ago but completely forgot about it. :(

jonpas (approved):
Can look into those other macros in another PR.

PabstMirror:
not sure if it will work but switching from `call compile` to `# include` might do it

PR #946, opened 2020-04-30
152/117F - Fix ZeroDivisor when adding spaces to channel names

ThymoNL:
- Fix a ZeroDivisor error when adding a space to a channel name
- Change maximum channel length to 12 to prevent overwriting the UI
- Presets "default2", "default3" and "default4" for the PRC-117F now use the proper scheme of "NET**XX**" instead of "FMLOSVOC**X**" for their channel names
- Fix an issue where the edit box would allow more characters to be added to a channel name than allowed.

Channel names are currently the length of whatever they are set to in the preset. This results in a ZeroDivisor if the name is not changed sequentially, for example when the user wants to insert a space by skipping a character in the menu.
This PR adds a check that extends the length of the name with spaces by the maximum channel name length the first time the menu is accessed.

**~~Potential issue:~~**
Considering the whole display width can be used to input a channel name this is problematic for displays where the full width is not available. Parts of the UI can be overwritten.
Perhaps it should be limited to what length the name can be on screen? I haven't included a fix for that in this PR because I don't know what limit the actual radios have and didn't want to pick an arbitrary one. I think to limit this you'd only have to change CHANNELNAME_MAX_LEN to what it should be, but I'm not sure if changes in other places are needed too.

_Overflow shown in PRC117F:_
PRC117F name overflow

_Overflow shown in PRC152:_
PRC152 name overflow

### DIFF
--- a/addons/sys_prc117f/farris_menus/PGM.sqf
+++ b/addons/sys_prc117f/farris_menus/PGM.sqf
@@ -19,6 +19,9 @@
 #define GET_RADIO_VALUE(x) [x] call FUNC(CURRENT_RADIO_VALUE)
 #define GET_CHANNEL_DATA() [] call FUNC(CURRENT_RADIO_CHANNEL);
 
+#define CHANNELNAME_MAX_LEN 20
+#define SPACE_CHAR 32
+
 DFUNC(CURRENT_RADIO_VALUE) = {
     private _channelNumber = ["getCurrentChannel"] call GUI_DATA_EVENT;
     private _channels = GET_STATE("channels");
@@ -399,8 +402,15 @@ GVAR(PGM_NORM_LOS) = ["PGM_NORM_LOS", "PGM_NORM_LOS", "",
             ],
             [
                 {
-                    private _value = nil;
-                    _value = GET_RADIO_VALUE("name");
+                    private _value = GET_RADIO_VALUE("name");
+                    private _valueLen = count _value;
+                    if (_valueLen < CHANNELNAME_MAX_LEN) then {
+                        TRACE_1("Resizing channel name", _valueLen);
+                        private _whitespace = [];
+                        _whitespace resize CHANNELNAME_MAX_LEN - _valueLen;
+                        _whitespace = _whitespace apply {SPACE_CHAR};
+                        _value = [_value, toString _whitespace] joinString "";
+                    };
                     SCRATCH_SET(GVAR(currentRadioId), "menuString", _value);
                 },
                 {
@@ -421,7 +431,7 @@ GVAR(PGM_NORM_LOS) = ["PGM_NORM_LOS", "PGM_NORM_LOS", "",
                 nil
             ],
             [
-                20,    // Digit count
+                CHANNELNAME_MAX_LEN,    // Digit count
                 [ROW_LARGE_3, 0, -1] // Highlighting cursor information
             ],
             "pgm_name"
--- a/addons/sys_prc117f/farris_menus/PGM.sqf
+++ b/addons/sys_prc117f/farris_menus/PGM.sqf
@@ -19,9 +19,6 @@
 #define GET_RADIO_VALUE(x) [x] call FUNC(CURRENT_RADIO_VALUE)
 #define GET_CHANNEL_DATA() [] call FUNC(CURRENT_RADIO_CHANNEL);
 
-#define CHANNELNAME_MAX_LEN 20
-#define SPACE_CHAR 32
-
 DFUNC(CURRENT_RADIO_VALUE) = {
     private _channelNumber = ["getCurrentChannel"] call GUI_DATA_EVENT;
     private _channels = GET_STATE("channels");
@@ -403,14 +400,7 @@ GVAR(PGM_NORM_LOS) = ["PGM_NORM_LOS", "PGM_NORM_LOS", "",
             [
                 {
                     private _value = GET_RADIO_VALUE("name");
-                    private _valueLen = count _value;
-                    if (_valueLen < CHANNELNAME_MAX_LEN) then {
-                        TRACE_1("Resizing channel name", _valueLen);
-                        private _whitespace = [];
-                        _whitespace resize CHANNELNAME_MAX_LEN - _valueLen;
-                        _whitespace = _whitespace apply {SPACE_CHAR};
-                        _value = [_value, toString _whitespace] joinString "";
-                    };
+                    _value = (_value + CHANNEL_PADDING_STRING) select [0,CHANNEL_NAME_MAX_LENGTH]; // Make sure we have something at each editindex
                     SCRATCH_SET(GVAR(currentRadioId), "menuString", _value);
                 },
                 {
@@ -431,7 +421,7 @@ GVAR(PGM_NORM_LOS) = ["PGM_NORM_LOS", "PGM_NORM_LOS", "",
                 nil
             ],
             [
-                CHANNELNAME_MAX_LEN,    // Digit count
+                CHANNEL_NAME_MAX_LENGTH, // Editable digits
                 [ROW_LARGE_3, 0, -1] // Highlighting cursor information
             ],
             "pgm_name"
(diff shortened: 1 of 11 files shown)


addons/sys_prc117f/farris_menus/PGM.sqf
```diff
@@ -399,8 +402,15 @@ GVAR(PGM_NORM_LOS) = ["PGM_NORM_LOS", "PGM_NORM_LOS", "",
...
-                    _value = GET_RADIO_VALUE("name");
+                    private _value = GET_RADIO_VALUE("name");
+                    private _valueLen = count _value;
+                    if (_valueLen < CHANNELNAME_MAX_LEN) then {
+                        TRACE_1("Resizing channel name", _valueLen);
+                        private _whitespace = [];
+                        _whitespace resize CHANNELNAME_MAX_LEN - _valueLen;
+                        _whitespace = _whitespace apply {SPACE_CHAR};
+                        _value = [_value, toString _whitespace] joinString "";
+                    };
```
Sniperhid: This approach seems pretty long winded, couldn't you just do `_value = _value select [0, 20]` to substring the string? This would then also remove the need to trim?

addons/sys_prc117f/fnc_preset_information.sqf
```diff
@@ -88,7 +88,7 @@ for "_i" from 0 to 99 do {
     HASH_SET(_channel,"power",20000);
     HASH_SET(_channel,"encryption",0);
     HASH_SET(_channel,"channelMode", "BASIC");
-    private _desc = format["FMLOSVOC%1",([_i+1, 2] call CBA_fnc_formatNumber)];
+    private _desc = format["NET%1",([_i+1, 2] call CBA_fnc_formatNumber)];
```
Sniperhid, line 91: Not sure we want to change the default channel names at this time.

Sniperhid (changes requested):

addons/sys_prc117f/farris_menus/PGM.sqf
ThymoNL: No, that wouldn't work. If you use a length greater than the length of the string the result just gets truncated to the length of the string. The whole idea is to extend the string so that when a character is inserted there actually is a spot to store it. If not there is either no array element there or the string will have nils in between.

For example if a channel is named "NET01" and I want to name it "NET01 A" you get a zero divisor because there is no more array:
[ N, E, T, 0, 1 ]   A
________________^

addons/sys_prc117f/fnc_preset_information.sqf
ThymoNL: Revert issue, had code there earlier. Nevertheless, FMLOSVOC is coming from the PRC-152, the "default" preset uses the correct name. So the other sides receive incorrect names meant for the PRC-152 where the format of "NETXX" is supposed to be used. Thoughts?

addons/sys_prc117f/farris_menus/PGM.sqf
Sniperhid: Good point! I wouldn't be surprised if something like this was much faster though (just appending 20 spaces and substring)
```
_value = (_value + "                    ") select [0,20]; // Append 20 spaces to pad end with spaces and substring to length 20
```

ThymoNL: 0.0033 ms over 0.0012 ms
It's a little magic number-y but it works. Changing the maximum channel length would then require hard coded changes in three places. Is that worth the 0.0021 ms the first time the menu is opened?

addons/sys_prc117f/fnc_preset_information.sqf
Sniperhid: Yeah, you're right. These should indeed be NETXX for the 117F, I was a little puzzled as it wasn't reference in the PR description.

This is definitely worth putting in the "When merged this pull request will:" section of the PR as a bullpoint as it should probably go in the changelog asa  fix as it might be that some legacy missions briefings reference these values.

ThymoNL: I've added it to the PR. My first attempt at fixing the name bug was to patch the name in the preset_information. Off course that doesn't work when someone doesn't use those defaults at all. After reverting those changes that was fixed, hence not in the original PR.

addons/sys_prc117f/farris_menus/PGM.sqf
Sniperhid: Max characters for channel names should be 12 for both 152 and 117F. And shall defer decision on which approach for padding the string to use to @jonpas

jonpas: > Changing the maximum channel length would then require hard coded changes in three places.

Not a problem, `#define` it in `script_component.hpp` of each of the 2 radios and you are done. All radio menus are designed as per real life, their parameters don't change dynamically, so there is no reason to waste performance or readability.

ThymoNL: You want the actual padding string defined too?

Sniperhid: If the channel name max length is going into a define, then it would probably make sense to define the padded string with it so that both declarations in the same place.

Otherwise the alternative would be to have a longer padded string that could handle a longer max length but that a negative of that is that it is still arbitrary. However the max length is probably never going to subsequently change once it is made 12 (which is why personally I wouldn't macro it). I am willing to accept macros though.

jonpas: All magic numbers should be macros.
Macros aren't there only for easy changing, but also for readability.

ThymoNL: Updated

Sniperhid (changes requested):
Nice, testing works okay,

Although this does highlight another bug when using the left arrow with the first digit selected it will allow the user to create a preset name 1 character longer than the maxmimum preset name size. (Effects both radios). I identified the source as follows:

In: https://github.com/IDI-Systems/acre2/blob/master/addons/sys_prc152/menus/types/Alphanumeric.sqf#L107

```
if (_editIndex > 0) then {
	_editIndex = _editIndex -1;
} else {
	_editIndex = _editDigits;
};
```

should be

```
if (_editIndex > 0) then {
	_editIndex = _editIndex -1;
} else {
	_editIndex = _editDigits -1;
};
```
Basically `_editDigits` stores the maximum number of digits for alphanumeric editor and `_editIndex` represents the index of the current selected character, but the index was going one higher than it should have been when going left and wrapping around.

ThymoNL:
Nice catch. Fixed.

Sniperhid:
Upon retesting came across another issue, on the 117F if you edit a preset name and then swap to another channel and attempt to edit its channel name it would not load the new channel name but keep the retain the last one you editing. My commit fixes that (making the var nil forces it to reload on next use)

Sniperhid (approved):

jonpas (approved):

addons/sys_prc117f/farris_menus/PGM.sqf
```diff
@@ -399,8 +400,8 @@ GVAR(PGM_NORM_LOS) = ["PGM_NORM_LOS", "PGM_NORM_LOS", "",
             ],
             [
                 {
-                    private _value = nil;
-                    _value = GET_RADIO_VALUE("name");
+                    private _value = GET_RADIO_VALUE("name");
+                    _value = (_value + CHANNEL_PADDING_STRING) select [0,CHANNEL_NAME_MAX_LENGTH]; // Make sure we have something at each editindex
```
TheMagnetar:
```suggestion
                    _value = (_value + CHANNEL_PADDING_STRING) select [0, CHANNEL_NAME_MAX_LENGTH]; // Make sure we have something at each editindex
```

addons/sys_prc152/farris_menus/PGM.sqf
```diff
@@ -86,13 +86,14 @@ GVAR(PGM) = ["PGM", "PGM", "PGM",
             [
                 {
                     private _value = GET_RADIO_VALUE("description");
+                    _value = (_value + CHANNEL_PADDING_STRING) select [0,CHANNEL_NAME_MAX_LENGTH]; // Make sure we have something at each editindex
```
TheMagnetar:
```suggestion
                    _value = (_value + CHANNEL_PADDING_STRING) select [0, CHANNEL_NAME_MAX_LENGTH]; // Make sure we have something at each editindex
```

TheMagnetar (changes requested):
Minor formatting

TheMagnetar (approved):

PR #951, opened 2020-05-07
Extensions - Fix freeze on start with Tobii EyeTracker

### LINKED ISSUE #861: Freeze on start with Tobii EyeTracker
- **Arma 3:** `1.96.146114 stable` [e.g. 1.00 stable, rc,  dev]
- **CBA:** `3.12.2.190909` [e.g. 3.0.0 stable, commit hash]
- **ACRE2:** `2.7.1.1016` [e.g. 2.0.0 stable, commit hash]
_Make sure to reproduce the issue with only CBA, ACRE2 (and optionally ACE3) on a newly created mission!_

**Description:**
With only CBA and ACRE 2 loaded, and the Tobii service (eye tracking hardware/software) I can't start ARMA 3, I just get stuck on a black screen.
Seems that no matter how long I leave it at this black screen ARMA won't start, and I have to kill the process. ARMA crash reported shows the following as the last enrty.
11:32:40 CallExtension loaded: acre (D:\Games\SteamLibrary\steamapps\common\Arma 3\!Workshop\@ACRE2\acre_x64.dll) [2.7.1.1016]

From the ACRE logs I only see the following on each start up.
2019-11-14 11:51:38,807-{INFO }- ACRE Loaded
2019-11-14 11:51:38,809-{INFO }- Generating PBO List

I tried reinstalling arma 3 and the mods and I'm still having the same issue.

**Steps to reproduce:**
- Install arma3
- Install drivers for Tobii eye tracking.
- Subscribe to CBA and ACRE in the steam workshop
- Start the arma 3 launcher.
- Select to load CBA and ACRE 2
- Click start with mods
- Wait for the initializing box to finish it's work
- Get to the black screen that is typical before the first splash screen
- Get stuck at this black screen until the process is killed.

**Expected behavior:**
Expected to get to splash screen and then in to main menu.

- Starting ARMA 3

**Log Files:**
-https://gist.github.com/Deathwatch187/e4f57d2f67dd1a1bde78bb3d8f06efbd(https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- https://gist.github.com/Deathwatch187/5fd9ea5c0bae13777092c174eacc9666

**Additional context:**
ACRE was able to install the TS plugin without issue
TeamSpeak version 3.3.2
A connected Tobii device is NOT needed, no such device was connected when I was testing this.

**Screenshots:**
N/A

eadelon:
The aim of this Pull Request is to fix the freeze on start with Tobii EyeTracker. The problem is due to a wrong usage of the `GetFinalPathNameByHandle` function when building the pbo file list. Indeed, the `GetFinalPathNameByHandle` function will hang if the specified handle is not of the type `FILE_TYPE_DISK`.

Close #861 _edit by @Jonpas_

### DIFF
--- a/extensions/src/ACRE2Arma/common/pbo/search.cpp
+++ b/extensions/src/ACRE2Arma/common/pbo/search.cpp
@@ -309,11 +309,17 @@ namespace acre {
                     std::string object_name(tmp_name.begin(), tmp_name.end());
                     //LOG(INFO) << "File: " << object_name;
                     if (object_type == "File" && object_name.find(".pbo") != object_name.npos) {
-                        char buffer[MAX_PATH];
-                        GetFinalPathNameByHandle(dupHandle, buffer, sizeof(buffer), VOLUME_NAME_DOS);
 
-                        //LOG(INFO) << "Pbo: " << buffer;
-                        _active_pbo_list.push_back(std::string(buffer));
+                        /* Do not try to get the path of objects which are not disk files. */
+                        /* The GetFinalPathNameByHandle function will hang if the specified handle is not of the type FILE_TYPE_DISK. */
+                        DWORD fileType = GetFileType(dupHandle);
+                        if (fileType == FILE_TYPE_DISK) {
+                            char buffer[MAX_PATH];
+                            GetFinalPathNameByHandle(dupHandle, buffer, sizeof(buffer), VOLUME_NAME_DOS);
+
+                            //LOG(INFO) << "Pbo: " << buffer;
+                            _active_pbo_list.push_back(std::string(buffer));
+                        }
                     }
                 }
 


TheMagnetar:
Awesome!

jonpas:
That is some nice work! Thanks a lot!

Sniperhid (approved):
So if I have understood correctly. It would seem that Tobii EyeTracker opens a named pipe and then ACRE using the API call `GetFinalPathNameByHandle` on the named pipeline handle as querying it will cause a block/freeze. Some interesting info here (particularly the first comment) -> http://0memory.blogspot.com/2008/08/getfinalpathnamebyhandle-api-hungs.html

This is some good detective work! Thanks for the contribution.

jonpas (approved):

PR #956, opened 2020-05-17
Extensions - Possible Fix for crashes with multiple connection tabs

### LINKED ISSUE #855: TeamSpeak crashes with multiple servers
**Description:**
TeamSpeak 3 will crash when attempting to connect to multiple servers with the ACRE plugin enabled.

**Steps to reproduce:**
- Enable ACRE
- Join a server
- Join a second server
- CTD

**Expected behavior:**
It doesn't crash

DeliciousJaffa:
- Possibly fix Teamspeak crashes on startup with multiple auto-connects (#855) 
See [this comment](https://github.com/IDI-Systems/acre2/issues/855#issuecomment-629826381) for further details.

This is **untested** as I do not have a build environment setup, but it is a simple enough change that someone with a build environment handy should be able to validate.

### DIFF
--- a/extensions/src/ACRE2TS/TSCallbacks_init.cpp
+++ b/extensions/src/ACRE2TS/TSCallbacks_init.cpp
@@ -90,7 +90,7 @@ void ts3plugin_currentServerConnectionChanged(uint64 serverConnectionHandlerID) 
     //TRACE("currentServerConnectionChanged %llu (%llu)", (long long unsigned int)serverConnectionHandlerID, (long long unsigned int)ts3Functions.getCurrentServerConnectionHandlerID());
 }
 
-void ts3plugin_onConnectStatusChangeEvent(uint64 id, int status, unsigned int err) {
+void ts3plugin_onConnectStatusChangeEvent(uint64 serverConnectionHandlerID, int status, unsigned int err) {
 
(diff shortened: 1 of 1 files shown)


ThymoNL:
Tested this. No regressions but also does not fix #855.

TheMagnetar:
Probably crashes it because it may not be initialised at the very begining. Thanks for testing this @Thymo-

DeliciousJaffa:
I'm not familiar with how the plugin works in entirety, but some calls to `ts3Functions.getCurrentServerConnectionHandlerID()` remain, possibly some involved in startup.

If `CTS3Client::setClientMetadata` or `CTS3Client::exPersistVersion` (which calls setClientMetadata) calls are part of the initialisation process, it will lead to another potentially crashing call here:
https://github.com/IDI-Systems/acre2/blob/0fda8b8adf5d4b9039663099b7292bcd611a0e85/extensions/src/ACRE2TS/TS3Client.cpp#L118-L138

Perhaps it might be a good idea for someone more familiar with the ACRE internals to have a pass over and see if connectionHandlers can be safely passed into these ACRE functions to avoid any reliance on being on the "correct" teamspeak tab?

jonpas:
No further activity, feel free to re-open.

Issue #685, opened 2019-06-03
Can't hear yourself through speakers on Radios


nameless:
**Arma 3 Version:** `1.92` (stable / rc / dev)
**CBA Version:** `3.11.2` (stable / dev + commit hash)
**ACRE2 Version:** ` 2.6.3.1003` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- acre2
- ACE
```

**Description:**
- When you enable speaker mode for any of the radios and you put it on the ground and try transmitting on that same frequency with a different radio, you cannot hear yourself from the radio that is on the ground that has speaker mode on.  

**Steps to reproduce:**
- Turn on speaker mode for any of the radios and drop it
- Pick up another radio and set channel to the same one that is dropped
- Start transmitting and you cant hear yourself from the radio that is on the ground

**Where did the issue occur?**
- Self-Hosted Multiplayer

Fynnoverse:
Just get me this clear.
You want TS3 to replay your Microphone. 
Because this is how its wrks or not, the ACRE2 Plugin tells TS3 how to "listen" on the others user, so how would it be in a simple way possible ??? you cannot hear yourself in TS3 ecept for the Mic-Test in the Config. 
Yeah it should be Physically possible, but dude rly?

NouberNou:
Most of this functionality is in the engine, but it was causing some very hard to track down crashes when it was first tested, so it kinda got shoved onto the back burner... years and years ago.

Fynnoverse:
Yeah thats what i mean with "simple way".
Sure its possible but i see this more like a feature request and not a bug.
My opinion ^^

jonpas:
Yeah, this isn't a bug. It is one of those things we would really want to do. I'll close this, because it's specific about something that's actually much larger (hearing yourself generally), we have it in our minds always.

PR #963, opened 2020-05-21
Add spectator radios integration for EG and ACE spectators

mharis001:
- Add spectator radios UI for focused entities in the EG and ACE spectator displays
- Radios are added/removed by clicking entries in the list
- Add keybind to clear spectators radios list while in spectator displays
- Unfortunately a dialog to display all current spectator radios doesn't seem viable since we are unable to prevent mouse scroll/drag moving camera and mouse click selecting entities

<details>
<summary>Images</summary>

acre_spectator_1

acre_spectator_2

</details>


TheMagnetar (dismissed):

TheMagnetar:
Once the conflicts are resolved it can be merged in my opinion

jonpas (commented):
I understand ACE Spectator will also automatically have that due to it extending EG Spectator?

addons/sys_spectator/stringtable.xml
```diff
@@ -0,0 +1,14 @@
...
+<Project name="ACRE2">
+    <Package name="Sys_Spectator">
+        <Key ID="STR_ACRE_Sys_Spectator_ClearRadios">
+            <English>Clear Spectator Radios</English>
+        </Key>
+        <Key ID="STR_ACRE_Sys_Spectator_ClearRadios_Description">
+            <English>Clears all radios from the spectator radios list.</English>
+        </Key>
+        <Key ID="STR_ACRE_Sys_Spectator_NoRadios">
+            <English>No Radios.</English>
```
jonpas:
```suggestion
            <English>No Radios</English>
```
Would look better I think.

mharis001:
No ACE spectator will not, it is based on EG Spectator but does not extend it. Will add support for it in another PR; should be fairly straight forward but need to add a load event on ACE's side.

jonpas:
Requires more research and figure out listening to radios that are currently transmitting (listening to outgoing transmissions).

BrettMayson:
> Outgoing transmissions issue continues

This would be the similar to #685 ?

jonpas:
It is the same cause I believe, but #685 is harder to solve, as we still care about signal processing, however we don't even want that for spectator.

jokoho48:
> Outgoing transmissions issue continues

I played around with that for my spectator and we don't add own radios here
https://github.com/IDI-Systems/acre2/blob/f79d12e5180b11380efe5fff2a699619712e953a/addons/sys_core/fnc_processRadioSpeaker.sqf#L116
I think adding a simple `ACRE_IS_SPECTATOR` check here could fix it. Sadly I didn't found anyone to test that hypothesis

jonpas:
A fix has already been conjured, but going to tag @Soldia1138 on above.

Soldia1138:
That won’t work. But as @jonpas said, there’s already a fix in the pipeline

jonpas:
Pushing back until ACE Spectator support exists.

mharis001:
Added ACE Spectator support. Requires next ACE update which will add the loaded event but don't need to wait for that to merge (not a hard dependency).

TheMagnetar (approved):
LGTM

jonpas (approved):

PR #964, opened 2020-05-21
GUI - Close inventory context menu after using open radio action

mharis001:
- title, was kept open because `acre_sys_radio_fnc_openRadio` returns `true`


jonpas:
Oh, so that's not "just how it works". Oops!

jonpas (approved):

PR #965, opened 2020-05-21
GUI - Use XEH for inventory display initialization

mharis001:
- Use display XEH to initialize `RscDisplayInventory` instead of waiting for display to not be null after `InventoryOpened` event
- Use IDC defines from base game `defineResincl.inc` file


jonpas (approved):
Great!

ThymoNL:
I can no longer build ACRE with this change due to `\a3\ui_f\hpp\defineResincl.inc` not existing.

jonpas:
It was added in #963.

ThymoNL:
That PR is not merged yet so merging this PR before that one breaks the master branch

jonpas:
And that's fine, just grab the include from that one for the time being. Things don't move as fast as we would want.

PR #971, opened 2020-05-29
Extensions - Fix freeze on start with Tobii EyeTracker (try 2)

eadelon:
The aim of this Pull Request is to fix the freeze on start with Tobii EyeTracker. I am sorry for the double post but the PR #951 does not fix entirely the freeze.

After a fresh install, I noticed that the function `NtQueryObject` can hang if the handle type is not supported. I found this post which explain the issue https://stackoverflow.com/questions/26867480/ntqueryobject-hangs-on-object-type-number-30-with-specific-access-mask

### DIFF
--- a/extensions/src/ACRE2Arma/common/pbo/search.cpp
+++ b/extensions/src/ACRE2Arma/common/pbo/search.cpp
@@ -241,6 +241,14 @@ namespace acre {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


TheMagnetar (approved):

Sniperhid (approved):

jonpas (approved):

jonpas:
Nice!

PR #973, opened 2020-06-02
Extension - Multi-thread init pbo/file scan

PabstMirror:
Moderate improvements to pre-start loading time

test results for 45GB of mods (CUP+RHS)

-fast ssd:
8067 ms spent in callExtension calling name: "acre", function: "init:"
6502 ms spent in callExtension calling name: "acre", function: "init:"
6541 ms spent in callExtension calling name: "acre", function: "init:"

2003 ms spent in callExtension calling name: "acre", function: "init:"
2386 ms spent in callExtension calling name: "acre", function: "init:"
1972 ms spent in callExtension calling name: "acre", function: "init:"

-ancient spinning rust HD:
19270 ms spent in callExtension calling name: "acre", function: "init:"

13282 ms spent in callExtension calling name: "acre", function: "init:"

PR #975, opened 2020-06-07
Extension - Limit number of pbo scan threads

PabstMirror:
Fix for #973

A client with `Windows 7 x64` experienced file read failures
The original PR just tried to open everything at once and I assume overloaded older versions of windows. It may also have had issues with thread limits on 32 bit windows.

This PR limits the number of read threads to 16 which should prevent this issue


NouberNou:
Suggest we use whatever [`std::thread::hardware_concurrency()`](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency) returns. Otherwise you'll just get thread contention and possibly slow stuff down.

PabstMirror:
These threads are just going to spend most of their time just waiting for hard drive data.
I think having more active data seeks should help overall read speed (i.e. should allow more efficient use of Native Command Queuing)

jonpas (approved):

PR #978, opened 2020-06-11
WIP: Write radio sound files from pbo to temp folder (drop b64)

PabstMirror:
Copy radio sounds directly from pbo instead of loading b64 files
I'm really not sure if this is a good idea or not

Plus:
- fix "b64" not allowed errors  (server config)
- easier to add custom radio sounds
- can cleanup all the b64 code

Negative:
- Might break something?
- breaks old custom b64 radio sounds


NouberNou:
So yea, there is no reason we have to use b64 still, since it was a relic of the pre-pbo code. Also I don't think there is a need to copy them to a temp folder. We are storing them and playing them from memory in the sound system if I remember right. We should be able to just pass a path from Arma to the extension and pull it into memory from the pbo like we do with the wrps.

NouberNou:
I think we should leave the old b64 functionality though imo to have backwards compat.

NouberNou (changes requested):
See previous comments.

extensions/src/ACRE2Arma/common/pbo/fileloader.hpp
```diff
@@ -1,22 +1,24 @@
...
+#pragma comment(lib, "shlwapi.lib")
 
 namespace acre {
     namespace pbo {
         class file_entry {
         public:
-            file_entry() {};
-            file_entry(file_p entry_) : _file(entry_) { 
+            file_entry() = delete;            
+            file_entry(file_p entry_) : _file(entry_) {
```
TheMagnetar, line 16:
```suggestion
            explicit file_entry(file_p entry_) : _file(entry_) {
```

extensions/src/ACRE2Arma/common/pbo/fileloader.hpp
```diff
@@ -1,22 +1,24 @@
...
+#include <shlwapi.h>
+#pragma comment(lib, "shlwapi.lib")
 
 namespace acre {
     namespace pbo {
         class file_entry {
         public:
-            file_entry() {};
-            file_entry(file_p entry_) : _file(entry_) { 
+            file_entry() = delete;
```
TheMagnetar, line 15: No need to delete. Simply delete the constructor.
```suggestion
            file_entry() = delete;            
```

TheMagnetar (changes requested):

Issue #983, opened 2020-06-28
No radio sounds (b64 whitelisted) 


tyl3r99:
**Mods (ACE, MLO, CBA, enhanced movement, shacktack, jsrs, cup weapons, rhs usaf and afrf, achillies, project opfor):**
- **Arma 3:** `Latest stable version as of 28th june 20` [e.g. 1.00 stable, rc,  dev]
- **CBA:** `cba_3.15.1.200418` [e.g. 3.0.0 stable, commit hash]
- **ACRE2:** `2.7.4.1027` [e.g. 2.0.0 stable, commit hash]
_Make sure to reproduce the issue with only CBA, ACRE2 (and optionally ACE3) on a newly created mission!_

**Description:**
I get the yellow box that shows when transmitting however no radio beeps or skwawking.
ACRE does connect to teamspeak and moves me into our operations channel just fine. 
I do not get any in game error messages i.e. pipe closed
I looked in my RPT and this is getting spammed. 

if (_name != "ACRE_TR") exitWith {};
(parseSimpleArray _data) params ["_id", ">
 9:26:49   Error position: <parseSimpleArray _data) params ["_id", ">
 9:26:49   Error Generic error in expression
------
9:25:45 Warning: 47324 ms spent in callExtension calling name: "ACRE2Steam", function: "2VER2.7.4.1027"
------
9:26:25 Wrong location draw style - "ACRE_RequiredDrawStyle"
------
9:26:43 [ACRE] (sys_core) WARNING: SquadLead has no TS3 ID 

**Steps to reproduce:**
- Go to ...
- Click ...
- See ...

**Expected behavior:**

**Where did the issue occur?**
- Dedicated
 (b64 was whitelisted in the extensions in the server.cfg)

**Log Files:**
- Link to ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- Link to ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the extension log file, which can be found at `<Arma 3 directory>\logs\acre_dll.log`.

Acre.dll log 
[acre_dll.log](https://github.com/IDI-Systems/acre2/files/4841933/acre_dll.log)
**Additional context:**

**Screenshots:**

PabstMirror:
It would be helpful to see full rpt

tyl3r99:
[...]

PR #984, opened 2020-06-28
Signal - Improve calculation when antennas are at the same position

PabstMirror:
Related to errors seen in #983 but signal calcs may just be unrelated

- Handle bad signals when antennas are at same position

Could reproduce error with model set to `LOS Simple`
```
[] spawn { 
    player addItem "ItemRadio"; 
    waitUntil {"ACRE_PRC343_ID_2" in (items player)}; 
    [2400, 5000, "ACRE_PRC343_ID_1", "ACRE_PRC343_ID_2"] call acre_sys_signal_fnc_getSignal; 
};
```

`parseSimpleArray` does not like `nan` returns from the extension
There were many places that would cause inf/-inf when positions were the exact same, 
so I figured this hacky offset fix was a lot simpler than handling it everywhere.  e.g.:

free space path loss:
```
const float32_t distance_3d = glm::distance(tx_pos_, rx_pos_);
... log10f(distance_3d) // fails
```
antenna gain:
```
glm::normalize(tx_pos - rx_pos); // fails
```

I also added a generic sanitizer as I also saw NaNs from LongRice but am not sure of the cause
```
{ERROR}- Signal was nan: 3,ACRE_PRC343_ID_2_ACRE_2HALFINCH_UHF_TNC_ACRE_PRC343_ID_1_ACRE_2HALFINCH_UHF_TNC,1701.47,5654.21,6.25005,0.0138816,0.951126,0.308492,ACRE_2HALFINCH_UHF_TNC,1601.18,5670.57,-2.10281,0.0315743,-0.0856666,-0.995823,ACRE_2HALFINCH_UHF_TNC,2400,5000,1,295.531,0,0,: -nan(ind),-nan(ind)
```

- Fix args from arrays
Needed to strip the [] from arrays
This caused problems with parsing first arg from arrays

### DIFF
--- a/extensions/src/ACRE2Arma/signal/antenna/antenna.cpp
+++ b/extensions/src/ACRE2Arma/signal/antenna/antenna.cpp
@@ -30,6 +30,9 @@ float32_t acre::signal::antenna::gain(const glm::vec3 dir_antenna_, const glm::v
     if ((f_ < _min_frequency) || ((f_ + _frequency_step) > _max_frequency)){
         return -1000.0f;
     }
+    if (glm::length(dir_signal_) <= 0.0f) {
+        return 0.0f; // cannot get direction if antennas have same position
+    }
 
     const glm::vec3 dir_antenna_v = glm::normalize(dir_antenna_);
     const float32_t elev_antenna = asinf(dir_antenna_v.z)*57.2957795f;
--- a/extensions/src/ACRE2Arma/common/arguments.hpp
+++ b/extensions/src/ACRE2Arma/common/arguments.hpp
@@ -48,11 +48,19 @@ namespace acre {
         arguments(const char** argv, std::int32_t argc) : _internal_index(0) {
             for (std::int32_t i = 0; i < argc; i++) {
                 std::string arg_string(argv[i]);
-                if (arg_string.size() > 2 && arg_string.at(0) == '\"' && arg_string.at(arg_string.size() - 1) == '\"') {
-                    arg_string = arg_string.substr(1, arg_string.size() - 2);  // callExtensionArgs will add quotes to strings
-                } 
-                std::vector<std::string> arg_split = acre::split(arg_string, ',');
-                _args.insert(_args.end(), arg_split.begin(), arg_split.end());
+                if (arg_string.size() > 2) {
+                    if (arg_string.at(0) == '\"' && arg_string.at(arg_string.size() - 1) == '\"') {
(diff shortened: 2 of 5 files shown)


TheMagnetar:
Hmmm... I am not sure. I can see normalized distance and natural logs being a problem, but that would not solve everything. A better and most reliable solution might be to return a 0 path loss for close antennas.

TheMagnetar:
With this solution you make the whole link budget, and not only the path loss, 0. What i meant is handling it internally in each model. For example, lines 13 to 16 in https://github.com/IDI-Systems/acre2/blob/master/extensions/src/ACRE2Arma/signal/models/arcade.cpp are handled as:

```
float32_t lossFreeSpace = 0.0F;
if (distance_3d > 0.001) {
    lossFreeSpace = -27.55f + 20.0f*log10f(frequency_Hz) + 20.0f*log10f(distance_3d); /* Free Space Path Loss model */
}
```

PabstMirror:
>Fix args from arrays: Needed to strip the [] from arrays

This is far more important than I realized because it is causing the first arg of an array to not be parsed correctly.
ie, 1st value will be interpreted as 0 
probably responsible for #986 (I bet it is much more noticeable on maps with hills on edge as opposed to islands)

extensions/src/ACRE2Arma/signal/antenna/antenna.cpp
```diff
@@ -30,6 +30,9 @@ float32_t acre::signal::antenna::gain(const glm::vec3 dir_antenna_, const glm::v
     if ((f_ < _min_frequency) || ((f_ + _frequency_step) > _max_frequency)){
         return -1000.0f;
     }
+    if (glm::length(dir_signal_) <= 0.0f) {
+        return 0.0f; // cannot get direction if antennas have same position
```
TheMagnetar, line 34: That is not "really true". You can have the same position but different angles but that would mean reworking some part of the code. However this is out of the scope of this PR.

TheMagnetar (approved):

PabstMirror:
New Summery
- Fixes a really huge (and really embarrassing) bug with my "optimization" where all arrays had first element interpreted as 0 because it was doing `atoi "[4.2"`
- `parseSimpleArray "NaN"` throws script error whereas call compile "NaN" would just give a nil
- Attempt to prevent any returns of NaN, mainly focused on cases where distance is 0

jonpas (approved):

TheMagnetar (approved):

PR #996, opened 2020-07-21
Enable unicode characters for TS channel and Arma server names

### LINKED ISSUE #579: Non-ASCII characters break channel switching
**Arma 3 Version:** `1.84` stable
**CBA Version:** `3.8.0` stable
**ACRE2 Version:** `2.6.1` stable

**Mods:**
```
- CBA_A3
- acre2
```

**Description:**
Non-ASCII characters in the server name breaks the channel switching system.

**Steps to reproduce:**
- [Sample server.cfg](https://files.slack.com/files-pri/T0J06UTRP-FCNEBG5A7/download/server.cfg)
- `hostName = "7.Kolordu Komutanlığı Teamspeak (ACRE) | 7KolorduTS | ";`

- Dedicated

**Placed Modules:**
- None

Issue created for max315 from Slack.

tbeswick96:
- Allow unicode characters in TeamSpeak channel names and Arma server names
- Change relevant TS3 methods for channel switching to use `wstring` instead of `string`
- Add 2 conversion methods to convert between `string` and `wstring`
- Stop the `TextMessage` `__isascii` check from running only for `setTs3ChannelDetails` procedure
- Fix #579 
- Tested locally using a TeamSpeak channel name containing `Grę` and a local Arma MP instance with server name `Grę`. Channel switching works as expected. All other functionality seems unaffected (Could use a smoke test if anyone is interested in building and testing)

---

Notes:
- The data coming from arma comes in as `char*`. This gets interpreted as ASCII, and the majority of data passed around inside ACRE is ASCII.
- The unicode conversion for channel switching only happens in the RPC function `setTs3ChannelDetails`. This means data is sent into it in the "wrong" format, is converted and used accordingly, and where needed passed back out in the "wrong" format. The out I think is only for channel passwords. So it may be the case that a channel password containing unicode doesn't work as expected, however the bytes remain the same so TS should interpret it correctly. I've not tested this.


TheMagnetar:
What about using `std::string test = u8"test"` instead of wstring?
https://en.cppreference.com/w/cpp/language/string_literal

tbeswick96:
> What about using `std::string test = u8"test"` instead of wstring?
> [en.cppreference.com/w/cpp/language/string_literal](https://en.cppreference.com/w/cpp/language/string_literal)

u8, as far as i can see, can only be used for literals, so that doesn't help much for getting the string from a variable

extensions/src/ACRE2Shared/StringConversions.cpp
```diff
@@ -0,0 +1,24 @@
+#include "compat.h"
+
+#include "StringConversions.h"
+#include <string>
+
+std::wstring StringConversions::stringToWstring(const std::string& str) {
+    int num_chars = MultiByteToWideChar(CP_UTF8, 0, str.c_str(), str.length(), NULL, 0);
```
TheMagnetar: Fixed integer sizes
```suggestion
    int32_t num_chars = MultiByteToWideChar(CP_UTF8, 0, str.c_str(), str.length(), NULL, 0);
```

extensions/src/ACRE2Shared/TextMessage.cpp
```diff
@@ -22,15 +22,19 @@ acre::Result CTextMessage::parse(char *const value, const size_t len) {
...
-            LOG("INVALID PACKET DETECTED l:%d", len);
-            return acre::Result::error;
-        }
-        if (value[x] == 0x00) {   // null terminate, bail
-            break;
+    // Don't check for ascii only characters for setTs3ChannelDetails procedure
+    std::string text(value);
+    if (text.find("setTs3ChannelDetails") == std::string::npos) {
+        // Check to make sure the entire chunk of data is a NULL terminated ascii string
+        for (size_t x = 0; x < len; x++) {
```
TheMagnetar:
```suggestion
        for (size_t x = 0U; x < len; x++) {
```

TheMagnetar (changes requested):
Great stuff! only minor changes. The PR looks good but I would like to see how the mumble plugin supports it, before merging this branch into master.

extensions/src/ACRE2Shared/TextMessage.cpp
tbeswick96: This line wasn't edited by me but have made the change anyway

tbeswick96:
@TheMagnetar Thanks, I don't have Mumble though so I can't test this with that, someone else with that all setup will need to lend a hand with that

TheMagnetar:
For sure, this is in our hands atm. Thanks for the PR @tbeswick96

jonpas:
Mumble compatibility branch here: https://github.com/IDI-Systems/acre2/tree/mumble-unicode

Tried with `Test lığı` as server name and it switched correctly on Mumble. However on TeamSpeak it moves me to the wrong channel (just uses first one that contains "ACRE" it seems). Could use another set of eyes.

jonpas:
Did anyone get a chance to test this with my example? @tbeswick96

tbeswick96:
I've not forgotten about this, just really busy. Intending to try and get this done in a few weeks

jonpas:
Most likely replaced by #1326. Please reopen if not the case.

PR #1000, opened 2020-08-04
Core - Add command to enable/disable ACRE globally

### LINKED ISSUE #862: Disable / Enable for the server
**Is your feature request related to a problem?**
Sometimes when doing training or testing we don't want to use ACRE.

**Solution you'd like:**
`#acre-disable` and `#acre-enable` that tell the addon to just chill.

This is slightly better than telling people to disable ACRE and then someone inevitably not, followed by them typing "why can't I hear anyone" in the game chat.

ThymoNL:
- Closes #862 
- Make ACRE "just chill"
- Add chat commands `#enable-acre` and `#disable-acre` for admins
- Add API functions to enable or disable ACRE for all players
- Block usage of radios
- Allow all people to hear each other, regardless of range

Could do with some more testing to see if this definitely works as intended. I didn't have anyone that was available and knew how to install my feature branch to test.


Sniperhid:
I think a simpler approach would be to use the override system, as seen in this if statement -> https://github.com/IDI-Systems/acre2/blob/master/addons/sys_core/XEH_postInit.sqf#L143

Instead of adding logic to all the handle speakers, because then there is also logic that needs to be added to range checks/mute checks and just avoids those edge cases.

The override has the added benefit of hearing everyone else in the Teamspeak channel as well.

jonpas:
We shouldn't be adding chat commands directly in ACRE2. Providing API functions is enough and users can select how to use them.

ThymoNL:
The whole point of this PR is to add an easy way to disable ACRE for all players in e.g. training missions, as was requested in the issue. The API functions were added as an extra. It would be weird to add this functionality and still require the mission maker to register a chat command to the API function. There won't be any conflicts with other mods either since the command contains 'acre' in it. I can't think of any reason against adding a chat command in ACRE.

@Sniperhid Sounds good, I'll make changes to use that instead.

jonpas:
What's the difference between typing the chat command or calling a function in debug console (to both which you have access as admin)?

You are right it wouldn't cause any conflicts of course, but a chat command feels like something more for the mission frameworks or missions themselves, not something we should supply.

ThymoNL:
I disagree, I don't see why this is something that should be limited to mission (frameworks). Would you also navigate to your Arma 3 directory to launch instead of having a desktop shortcut to do so because navigating there is "enough"?
Limiting it to the console makes the functionality more difficult to use. People running a training mission may not ever have touched a line of SQF in their live nor have a need to, this means they'll have to look up how to exactly call the function in the console. A chat command is easily memorised, the debug console should be used for debugging not intended functionality.

Sniperhid:
Regarding Chat Commands

A key thing to note is that CBA, ACE and ACRE are all intended as frameworks for a variety of communities to use and customise as needed. In essence the design paradigm of these is to provide generic and extensible systems. To date none of these mods register chat commands, as they leave it upto communities to implement as needed. So appreciate this is a first for ACRE2, and thus it's being met with questions surrounding if it is really necessarily.

Perhaps a way to fit in more within the existing paradigms is swap to a CBA setting an admin can change/mission can define.

addons/sys_core/fnc_isPaused.sqf
```diff
@@ -0,0 +1,19 @@
...
+ * Paused <BOOL>
+ *
+ * Example:
+ * [] call acre_sys_core_fnc_isPaused
+ *
+ * Public: No
+ */
+
+
+ACRE_IS_PAUSED
```
jonpas, line 19: This function is only used in API, we can use the global variable there directly as well.

addons/sys_core/XEH_postInit.sqf
```diff
@@ -93,6 +93,9 @@ if (!GVAR(aceLoaded)) then {
     [] call EFUNC(sys_components,toggleAntennaDir)
 }, [DIK_UPARROW, [false, true, true]]] call CBA_fnc_addKeybind;
 
+// Setup chat commands
+["acre-enable", { [false] call FUNC(setPaused); }] call CBA_fnc_registerChatCommand;
+["acre-disable", { [true] call FUNC(setPaused); }] call CBA_fnc_registerChatCommand;
```
jonpas, line 98: Two options
1. Move to documentation, so people can copy/paste it into their missions. Can also put it into one of the examples shipped with ACRE2.
2. Change into CBA Setting. (probably preferred)

I don't have preference of either, I think CBA Setting is easier as people can set it in-game, but CBA doesn't have a public setter (though many people use `CBA_settings_fnc_set`). Benefit over just documented API is also that it can be enabled in missions not created by you.

addons/sys_core/fnc_setPaused.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+
+params ["_pause"];
+
+TRACE_2("setPaused", ACRE_IS_PAUSED, _pause);
+if !(ACRE_IS_PAUSED isEqualTo _pause) then {
+    if (!_pause) then {
+        systemChat "ACRE Enabled";
+    } else {
+        systemChat "ACRE Disabled";
```
jonpas, line 26: Stringtable both.

addons/sys_core/fnc_setPaused.sqf
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [true] call acre_sys_core_fnc_setPaused
+ *
+ * Public: No
+ */
+
```
jonpas, line 17:
```suggestion
```

addons/sys_server/fnc_setPaused.sqf
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [true] call acre_sys_server_fnc_setPaused
+ *
+ * Public: No
+ */
+
```
jonpas, line 17:
```suggestion
```

jonpas (changes requested):
Sorry for taking longer on this, but the conclusion is we don't want chat commands in ACRE2 directly, alternatives given are CBA Settings or providing a copy/paste function for missions.

PR #1001, opened 2020-08-05
God Mode - Expand and Improve

jonpas:
- Add ability to set custom group names
- Improve remote control support
  - Use `profileName` instead of unit's name
  - Use `player` unit instead of `acre_player` unit to get target units in current chat (BI) channel
- Improve switch units support (different unit and original changes locality or gets deleted)
  - Add support for UID group targets (in addition to specific units)
  - Add support for dynamic targets (code that runs on PTT press and provides target units)
- Add custom display (eg. spectator) PTT keybind compatibility
  - Extension of `addDisplayPassthroughKeys`
- Remove `alive` target units filtering (dead players should also hear gods)
- Improve virtual entities support in Current Chat Channel (spectators and Zeuses without player unit)
  - Exclude headless clients (redundant target events)

---

Other smaller or regressed things:
- Fix group targets not being respected after change to RPC handling
  - Use TS3/VOIP ID
- Fix Zeus range checks and filtering (separate speaking type for Zeus)
- Fix `sendText` condition
- Cleanup variable names
- Show name of God who sent text message

God Mode


addons/sys_godmode/script_component.hpp
```diff
@@ -15,4 +15,20 @@
 #endif
 
 #include "\idi\acre\addons\main\script_macros.hpp"
-#include "script_acre_component_defines.hpp"
+
+
+// Group defines
+#define GODMODE_CURRENTCHANNEL   0
```
TheMagnetar: API functions use these defines. That is why they are declared in addons/sys_godmode/script_acre_component_defines.hpp

TheMagnetar (changes requested):

addons/sys_core/fnc_remoteStartSpeaking.sqf
```diff
@@ -143,9 +143,9 @@ private _result = false;
                 WARNING_1("Got start speaking event with non-existent radio id: %1",_radioId);
             };
         } else {
-            if (_onRadio == 5) then { // God Mode / Zeus
+            if (_onRadio == 5) then { // God Mode
```
TheMagnetar: 5 is a magic  number. I would define it so we can reuse it/adapt it in the future.

```suggestion
            if (_onRadio == REMOTE_SPEAKING_GOD_MODE) then { // God Mode
```

addons/sys_core/fnc_speaking.sqf
```diff
@@ -259,8 +259,12 @@ if !(GVAR(keyedMicRadios) isEqualTo []) then {
 
 {
     private _speakingId = _x;
-    private _params = ["g", _speakingId, 0, GVAR(godVolume)];
-    CALL_RPC("updateSpeakingData", _params);
+
+    // Check speaking gods to only allow hearing by the god's target group
+    if (EGVAR(sys_godmode,speakingGods) find _speakingId != -1) then {
```
TheMagnetar: Minor cosmetic suggestion.

```suggestion
    if ((EGVAR(sys_godmode,speakingGods) find _speakingId) != -1) then {
```

extensions/src/ACRE2Core/updateSpeakingData.h
```diff
@@ -33,8 +33,8 @@ RPC_FUNCTION(updateSpeakingData) {
     CEngine::getInstance()->getSoundEngine()->getSoundMixer()->lock();
     if (speaker) {
         LOCK(speaker);
-        if ((speakingType == "d") || (speakingType == "i")) {
-            if ((speaker->getInitType() != "d" && speaker->getInitType() != "i") || speakingType != speaker->getInitType()) {
+        if ((speakingType == "d") || (speakingType == "i") || (speakingType == "z")) {
+            if ((speaker->getInitType() != "d" && speaker->getInitType() != "i" && speaker->getInitType() != "z") || speakingType != speaker->getInitType()) {
```
TheMagnetar:
```suggestion
            if (((speaker->getInitType() != "d") && (speaker->getInitType() != "i") && (speaker->getInitType() != "z")) || speakingType != speaker->getInitType()) {
```

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -182,8 +182,10 @@ acre::Result CTS3Client::localStartSpeaking(const acre::Speaking speakingType_,
         }
     }
 
-    if (speakingType_ == acre::Speaking::god && !this->getVAD()) {
-        this->microphoneOpen(true);
+    if (speakingType_ == acre::Speaking::god || speakingType_ == acre::Speaking::zeus) {
```
TheMagnetar:
```suggestion
    if ((speakingType_ == acre::Speaking::god) || (speakingType_ == acre::Speaking::zeus)) {
```

extensions/src/ACRE2TS/TS3Client.cpp
```diff
@@ -199,6 +201,7 @@ acre::Result CTS3Client::localStopSpeaking(const acre::Speaking speakingType_) {
         case acre::Speaking::direct:
             break;
         case acre::Speaking::god:
+        case acre::Speaking::zeus:
```
TheMagnetar, line 205: This is a C++17 feature indicating that you intentionally let a case fall through. 
```suggestion
            [[fallthrough]]
        case acre::Speaking::zeus:
```

TheMagnetar (changes requested):

addons/sys_core/fnc_remoteStartSpeaking.sqf
jonpas: Other values here would also have to get added, not sure this fits this PR scope.

addons/sys_godmode/fnc_getUnitsBIChannel.sqf
```diff
@@ -15,30 +15,27 @@
...
-        private _side = side acre_player;
-        _units = allPlayers select {(_side == (side _x)) && {(alive _x) || {[_x] call acre_api_fnc_isSpectator}}};
+    case 1: { // Side
+        private _side = side player;
+        _units = _units select {(_side == (side _x)) || {_x in ACRE_SPECTATORS_LIST}};
     };
-    case 3: {                            // Group
-        _units = (units group acre_player) select {(alive _x) || {[_x] call acre_api_fnc_isSpectator}};
+    case 3: { // Group
+        _units = units (group player);
```
TheMagnetar, line 26: Are spectator units in a different group? What about players triggering spectator mode through the debug console?

docs/wiki/frameworks/god-mode.md
```diff
@@ -0,0 +1,80 @@
...
+
+{% include important.html content="Dev-build only" %}
+
+All [God Mode](/wiki/user/god-mode) configuration only applies to the local machine. This allows great flexibility in configuration and different group presets for each player.
+
+## Access
+
+Current Chat Channel and Groups have separate access rights.
+
+```js
```
TheMagnetar, line 13: Other parts of the docu are formatted using c++ iirc

TheMagnetar (changes requested):

docs/wiki/frameworks/god-mode.md
jonpas: Nope, it should use `js` because it fits better for SQF. C++ is used for config.

addons/sys_godmode/fnc_getUnitsBIChannel.sqf
jonpas: > Are spectator units in a different group?

That depends on the spectator system. There is no change here regarding spectators really, you just had to add it because you were filtering `alive`. But we now want dead units to hear as well, after some discussion it makes sense, since it is "admin talk".

> What about players triggering spectator mode through the debug console?

They'll just hear group. Again, no change from before. I don't think we can make this flexible enough, spectator systems can handle it if wanted.

TheMagnetar (approved):

docs/wiki/frameworks/events-list.md
```diff
@@ -6,7 +6,7 @@ title: Events List
 
 | Name  | Arguments | Type | Version |
 | ----- |---------- | ---- | ------- |
-| `acre_startedSpeaking`        | Unit <OBJECT>, On Radio <BOOL>, Radio ID <NUMBER> | Local | 2.7.0 |
+| `acre_startedSpeaking`        | Unit <OBJECT>, On Radio <BOOL>, Radio ID <NUMBER>, Speaking Type <NUMBER> | Local | 2.7.0 |
 | `acre_stoppedSpeaking`        | Unit <OBJECT>, On Radio <BOOL> | Local | 2.7.0 |
-| `acre_remoteStartedSpeaking`  | Unit <OBJECT>, On Radio <BOOL>, Radio ID <NUMBER> | Local | 2.7.2 |
+| `acre_remoteStartedSpeaking`  | Unit <OBJECT>, Speaking Type <NUMBER>, Radio ID <STRING> | Local | 2.7.2 |
```
jonpas, line 11: For reference, this was just wrong before.

PR #1002, opened 2020-08-05
Core - Cleanup debug

### LINKED ISSUE #638: Reserved variable error in processRadioSpeaker

jonpas:
- Cleanup debug from #909 
- Close #638 

Root cause is still unknown, but most likely [this](https://github.com/IDI-Systems/acre2/blob/c03262be2190cd54c3cb5f27d7ca3eddf7270506/addons/sys_core/fnc_remoteStartSpeaking.sqf#L107). This just cleans up the debug code.

### DIFF
--- a/addons/sys_core/fnc_processRadioSpeaker.sqf
+++ b/addons/sys_core/fnc_processRadioSpeaker.sqf
@@ -25,14 +25,8 @@ private ["_okRadios", "_functionName"];
 params ["_unit","_playerRadios"];
 TRACE_2("",_unit,_playerRadios);
 
-private _radioId = _unit getVariable [QGVAR(currentSpeakingRadio), ""];
-if (_radioId == "") exitWith { false };
-
-// Debug #638
-if (isNil "_radioId") exitWith {
-    WARNING_1("reserved variable debug 1 - report on GitHub! [%1]",_radioId);
-    false
-};
+private _radioId = _unit getVariable QGVAR(currentSpeakingRadio);
+if (isNil "_radioId" || {_radioId == ""}) exitWith { false }; // #638 _radioId is nil
 
 // @todo if Underwater Radios are implemented
 //if (ACRE_LISTENER_DIVE == 1) exitWith { false };
@@ -49,12 +43,6 @@ if (!GVAR(speaking_cache_valid)) then {
         END_COUNTER(okradio_check);
     #endif
     _okRadios = (_okRadios select 0) select 1;
-
-    // Debug #638
-    if (isNil "_radioId") exitWith {
-        WARNING_3("reserved variable debug 2 - report on GitHub! [%1-%2-%3]",_radioId,_okRadios,_playerRadios);
-        false
-    };
 
     private _transmittingRadioData = [_radioId, "getCurrentChannelData"] call EFUNC(sys_data,dataEvent);
     private _mode = HASH_GET(_transmittingRadioData, "mode");
--- a/addons/sys_core/fnc_processRadioSpeaker.sqf
+++ b/addons/sys_core/fnc_processRadioSpeaker.sqf
@@ -25,7 +25,7 @@ private ["_okRadios", "_functionName"];
 params ["_unit","_playerRadios"];
 TRACE_2("",_unit,_playerRadios);
 
-private _radioId = _unit getVariable QGVAR(currentSpeakingRadio);
+private _radioId = _unit getVariable [QGVAR(currentSpeakingRadio), ""];
 if (_radioId == "") exitWith { false };
 
 // Workaround #638 - _radioId is nil - TODO investigate further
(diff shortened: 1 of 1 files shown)


addons/sys_core/fnc_processRadioSpeaker.sqf
```diff
@@ -25,14 +25,8 @@ private ["_okRadios", "_functionName"];
...
-private _radioId = _unit getVariable [QGVAR(currentSpeakingRadio), ""];
-if (_radioId == "") exitWith { false };
-
-// Debug #638
-if (isNil "_radioId") exitWith {
-    WARNING_1("reserved variable debug 1 - report on GitHub! [%1]",_radioId);
-    false
-};
+private _radioId = _unit getVariable QGVAR(currentSpeakingRadio);
+if (isNil "_radioId" || {_radioId == ""}) exitWith { false }; // #638 _radioId is nil
```
TheMagnetar: Although this fixes it, i would be interested to know why it is "". Maybe add a comment that we should investigate further?

jonpas: Better question is why it is nil. Empty string is just the default value of getVariable.

TheMagnetar: True. I would expand the comments nevertheless:)

jonpas: Yes, for sure. :)

jonpas: Expanded it with a probable source (only place where it gets set). However the issue lies deeper than that link of course.

addons/sys_core/fnc_processRadioSpeaker.sqf
```diff
@@ -25,14 +25,12 @@ private ["_okRadios", "_functionName"];
 params ["_unit","_playerRadios"];
 TRACE_2("",_unit,_playerRadios);
 
-private _radioId = _unit getVariable [QGVAR(currentSpeakingRadio), ""];
+private _radioId = _unit getVariable QGVAR(currentSpeakingRadio);
```
TheMagnetar: Is a default no longer necessary? Why checking then on line 29 `if (_radioId == "") exitWith { false };` and later in line 33 `if (isNil "_radioId") exitWith { false };`

jonpas: Default value was added here https://github.com/IDI-Systems/acre2/commit/28cd75cd9ba176742a1c304423dad07b3fa6e3c3#diff-bebd4cbc667fc63e690cc00cdee69f6a, because that was my first hunch where the problem is. I don't know how I managed to remove this in the previous commit, better stays!

addons/sys_core/fnc_processRadioSpeaker.sqf
```diff
@@ -28,11 +28,9 @@ TRACE_2("",_unit,_playerRadios);
...
 if (_radioId == "") exitWith { false };
 
-// Debug #638
-if (isNil "_radioId") exitWith {
-    WARNING_1("reserved variable debug 1 - report on GitHub! [%1]",_radioId);
-    false
-};
+// Workaround #638 - _radioId is nil - TODO investigate further
+// Probable source: https://github.com/IDI-Systems/acre2/blob/c03262be2190cd54c3cb5f27d7ca3eddf7270506/addons/sys_core/fnc_remoteStartSpeaking.sqf#L107
+if (isNil "_radioId") exitWith { false };
```
tbeswick96, line 33: Wouldn't it be better to check `isNil` first before the `== ""` above it?

TheMagnetar (approved):

addons/sys_core/fnc_processRadioSpeaker.sqf
jonpas: Sure, but how rarely this happens it makes absolutely no difference.

PR #1003, opened 2020-08-07
Documentation - Fix C++ undefined behavior

muttleyxd:
Avoid guidelines which could lead to undefined behavior.

Give new developers cleaner and better guidelines to write C++ code.
I see 2 issues with current guidelines:
- Namespaces that are meant to hold internal or implementation specific code not exposed readily to end users should be prefixed with two (2) underscores `__`.

This is directly undefined behavior, as C++ standard states:
https://stackoverflow.com/a/228797

```
Each name that contains a double underscore (__) or begins with an underscore followed by an uppercase letter (2.11) is reserved to the implementation for any use.
```

I assume, we're not using C++17, so I propose a fix to prefix internal namespaces with `internal_`. Otherwise we could use `internal::` and simply nest it.

- All protected/private members and methods should be prefixed with a single `_` to provide easy identification that the value is not in the public scope.

This is not yet undefined behavior. But if someone makes a variable called `_Val`, then we'll have an undefined behavior. So to avoid this, we could revert to simpler syntax, simply call arguments same way as variable names and class member can have `_` suffix.

```
All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.
```

### DIFF
--- a/docs/wiki/development/coding-guidelines-cpp.md
+++ b/docs/wiki/development/coding-guidelines-cpp.md
@@ -61,11 +61,11 @@ void some_function() {
 
 ## Namespaces
 
-All namespaces should be descendants of `acre::`. Namespaces that are meant to hold internal or implementation specific code not exposed readily to end users should be prefixed with two (2) underscores `__`.
+All namespaces should be descendants of `acre::`. Namespaces that are meant to hold internal or implementation specific code not exposed readily to end users should be prefixed with `internal::`.
 
 ```c++
 namespace acre {
-    namespace __my_helpers {
+    namespace internal::my_helpers {
         // internal helper functions
     }
     namespace my_functionality {
@@ -80,7 +80,7 @@ All classes should strictly maintain separation between declarations and definit
 
 ### Protected/Private Members/Methods
 
-All protected/private members and methods should be prefixed with a single `_` to provide easy identification that the value is not in the public scope.
+All protected/private members and methods should use the same style as variable names.
 
 Example:
 
@@ -90,11 +90,11 @@ class test_class {
 public:
     test_class(uint32_t init_);
 protected:
-    uint32_t _val;
+    uint32_t val;
 };
 
 //test_class.cpp
-test_class::test_class(uint32_t init_) : _val(init_) {};
+test_class::test_class(uint32_t init_) : val(init_) {};
 ```
 
 ## Templates


TheMagnetar:
We can use C++17. The conversion has not taken place yet.

docs/wiki/development/coding-guidelines-cpp.md
```diff
@@ -23,16 +23,16 @@ std::string test_string = "abc";
...
 ### Arguments
 
-Arguments to any function, method, constructor, etc should be suffixed with a single `_`.
+Arguments to any function, method, constructor, etc should follow the same style as variable names.
 
 Examples:
 
 ```c++
-void some_function(uint32_t arg_);
+void some_function(uint32_t arg);
```
TheMagnetar: The suffixing (arg_) is a convention that will be adopted in future projects.

docs/wiki/development/coding-guidelines-cpp.md
```diff
@@ -80,21 +80,21 @@ All classes should strictly maintain separation between declarations and definit
...
 
 ```c++
 //test_class.hpp
 class test_class {
 public:
-    test_class(uint32_t init_);
+    test_class(uint32_t init);
 protected:
-    uint32_t _val;
+    uint32_t val_;
```
TheMagnetar: Same as my comment above but inverted.

TheMagnetar (commented):
Just a couple of comments regarding our internal discussions some time ago.  Thanks for going through this though :+1:

muttleyxd:
> We can use C++17. The conversion has not taken place yet.

I changed the doc to hint that we can use nested namespaces for this, looks clean and simple in my opinion.

docs/wiki/development/coding-guidelines-cpp.md
muttleyxd: I see a problem with this approach - while docs indicate that we should use underscore with lowercase text identifier, like `_val`, this could cause annoying problems.

Imagine following case:
```
// something.h
class something {
  public:
    void do_something_with_val();
  private:
    int _Val;
};

// something.cpp
// #include "something.h"
#include <vector> // doesn't define _Val, but compiler devs can change it

#define COMPILER_UPDATED // let's say that COMPILER_UPDATED means, that vector include has changed, and including it results in defining _Val as false
#ifdef COMPILER_UPDATED
#define _Val false
#endif

void something::do_something_with_val()
{
  if (_Val != 1024) {
     // do_something_important
  }
}

int main()
{
    something s;
    // some internal operations that may change _Val to 1024
    s.do_something_with_val();
}
```

If there's a bigger review, then it's really easy to let a private variable like `_Val` to slip through, into the code. While it's unlikely case, it always can happen - compiler devs have the right to define anything they want, within the limits described by C++ standard.

TheMagnetar: Only arguments are suffixed. Members are without prefixes or suffixes

muttleyxd: Now it should be fine.

TheMagnetar (approved):
Looks fine. Poking @jonpas as our review master for a second opinion.

jonpas (approved):
Looks good to me, changes are in-line with our style, just using newer functionality and getting rid of undefined behaviours. :+1:

PR #1004, opened 2020-08-07
Extensions - C++ fixes

muttleyxd:
Improve C++ code quality and bring it a little closer to C++ standard.

I'm slowly working on making this compileable from Linux, don't expect any miracles but at least I should increase code quality a little.

Commit description:
1. Extensions - Properly access base class members 
As isocpp states - we have to pass `this->` to access base class members, when we are deriving from template class.
https://godbolt.org/z/31rqT1
https://isocpp.org/wiki/faq/templates#nondependent-name-lookup-members

2. Extensions - Use forward slashes for includes. 
Pretty obvious, this works anywhere, instead of only MSVC.

3. Extensions - Replace MS boolean by standard C++ values
Pretty self-explanatory too.

4. Extensions - Make macros compatible with GCC.
Additional `##` were unnecessary, it doesn't change compilation output on MSVC.
https://godbolt.org/z/xEdMxW
https://godbolt.org/z/1f477q


extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
```diff
@@ -192,7 +192,7 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function)
                 // allows up to 4096 char string (+1 to ensure NUL terminated), like initial version
 
                 //DEBUG("Read from pipe [%d]\n", hPipe);
-                const BOOL ret = ReadFile(readHandle, (LPVOID)value, sizeof(char) * (read_length - 1U), &cbRead, NULL);
+                const bool ret = ReadFile(readHandle, (LPVOID)value, sizeof(char) * (read_length - 1U), &cbRead, NULL);
```
TheMagnetar: This is unfortunately not related to C++ standard, but in correctly using the Windows specific functions:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile. As an example, this one uses BOOL instead of bool for a return type.

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
```diff
@@ -365,24 +365,24 @@ void ClosePipe() {
...
+    readConnected = false;
 }
 
 void Cleanup(void) {
 
 }
 
 
-BOOL APIENTRY DllMain( HMODULE hModule,
+int APIENTRY DllMain( HMODULE hModule,
```
TheMagnetar: Why return int? Documentation states clearly BOOL
https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain

TheMagnetar (changes requested):

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
PabstMirror: As long as it isn't redefined I think you can repeat the typedef multiple times: e.g.:
```
typedef int BOOL;
typedef int BOOL;
```
so we could just add `typedef int BOOL;` at the top, which should make it portable without having to change type

muttleyxd:
You're right, I removed changes related to `DllMain`, here's the full diff I made now

```
diff --git a/extensions/src/ACRE2Arma/acre/acre_dllmain.cpp b/extensions/src/ACRE2Arma/acre/acre_dllmain.cpp
index fd4a081e..0b27d043 100644
--- a/extensions/src/ACRE2Arma/acre/acre_dllmain.cpp
+++ b/extensions/src/ACRE2Arma/acre/acre_dllmain.cpp
@@ -4,10 +4,10 @@
 #include "controller.hpp"
 INITIALIZE_EASYLOGGINGPP
 
-int WINAPI DllMain(IN HINSTANCE hDllHandle,
+BOOLEAN WINAPI DllMain(IN HINSTANCE hDllHandle,
     IN DWORD     nReason,
     IN LPVOID    Reserved) {
-    bool bSuccess = true;
+    BOOLEAN bSuccess = TRUE;
     el::Configurations conf;
 
     switch (nReason) {
diff --git a/extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp b/extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
index 7701e0f7..64d81f6c 100644
--- a/extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
+++ b/extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
@@ -382,7 +382,7 @@ void Cleanup(void) {
 }
 
 
-int APIENTRY DllMain( HMODULE hModule,
+BOOL APIENTRY DllMain( HMODULE hModule,
                        DWORD  ul_reason_for_call,
                        LPVOID lpReserved
                      )
@@ -398,5 +398,5 @@ int APIENTRY DllMain( HMODULE hModule,
         Cleanup();
         break;
     }
-    return true;
+    return TRUE;
 }
diff --git a/extensions/src/ACRE2Arma/dynload/acre_dynload_dllmain.cpp b/extensions/src/ACRE2Arma/dynload/acre_dynload_dllmain.cpp
index ef88f54b..be6a94b4 100644
--- a/extensions/src/ACRE2Arma/dynload/acre_dynload_dllmain.cpp
+++ b/extensions/src/ACRE2Arma/dynload/acre_dynload_dllmain.cpp
@@ -4,10 +4,10 @@
 
 INITIALIZE_EASYLOGGINGPP
 
-int WINAPI DllMain(IN HINSTANCE hDllHandle,
+BOOLEAN WINAPI DllMain(IN HINSTANCE hDllHandle,
     IN DWORD     nReason,
     IN LPVOID    Reserved) {
-    bool bSuccess = true;
+    BOOLEAN bSuccess = TRUE;
     el::Configurations conf;
 
     switch (nReason) {
diff --git a/extensions/src/ACRE2Steam/CallExt_DllMain.cpp b/extensions/src/ACRE2Steam/CallExt_DllMain.cpp
index d1d8441f..6817ce0d 100644
--- a/extensions/src/ACRE2Steam/CallExt_DllMain.cpp
+++ b/extensions/src/ACRE2Steam/CallExt_DllMain.cpp
@@ -551,7 +551,7 @@ void Cleanup(void) {
 }
 
 
-int APIENTRY DllMain( HMODULE hModule,
+BOOL APIENTRY DllMain( HMODULE hModule,
                        DWORD  ul_reason_for_call,
                        LPVOID lpReserved
                      )
@@ -567,5 +567,5 @@ int APIENTRY DllMain( HMODULE hModule,
         Cleanup();
         break;
     }
-    return true;
+    return TRUE;
 }
```

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
muttleyxd: Fixed

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
muttleyxd: @PabstMirror you're right, I think it could be moved to a file called `compat_linux.h` or something, to be decided :grinning: 

@TheMagnetar fixed

TheMagnetar: What is fixed? I mean, why change `BOOL` to  `bool`? There is  no gain since the majority of that code is Windows specific and the API states BOOL instead of bool. Or am I seeing some unpublished code :)

The same applies in my opinion to a `compat_linux.h` file. The only solution I saw that was viable without a bunch of compats was using SDL2 for having the same code on windows and on linux. However, not everything can be done in that way (loading PBOs for example)

muttleyxd: > There is no gain since the majority of that code is Windows specific and the API states BOOL instead of bool. Or am I seeing some unpublished code :)

Currently no gain from this, you're right. I changed the commit to replace some of TRUE/FALSE directives, doesn't change behavior, but for cross-platform code that's one less constant to have.

muttleyxd:
Added a commit with QtCreator gitignore file and build directories

jonpas:
It is easier to review if you do not force push. There is no need for it.

extensions/src/ACRE2TS/TsCallbacks_static.cpp
```diff
@@ -32,7 +32,7 @@ void ts3plugin_infoData(uint64 serverConnectionHandlerID, uint64 id, enum Plugin
     type = type;
     id = id;*/
 
-    BOOL noAcre = FALSE;
+    BOOL noAcre = false;
```
TheMagnetar, line 35: If a variable is declared as BOOL, I expect it to be TRUE/FALSE, not true/false

extensions/src/ACRE2Core/Wave.cpp
```diff
@@ -33,14 +33,12 @@ CWave::~CWave(void)
...
-    BOOL bResult = FALSE;
-
-
+    BOOL bResult = false;
 
     std::ifstream file;
     file.open(wavFile, std::fstream::in | std::fstream::binary);
     if (file.bad() || !file.is_open())
-        return FALSE;
+        return false;
```
TheMagnetar, line 41: If a function returns BOOL, I would expect it to return TRUE/FALSE and not true/false.
```suggestion
        return FALSE;
```

extensions/src/ACRE2Arma/common/directx/d3d_display.cpp
```diff
@@ -227,7 +227,7 @@ namespace acre {
             // Create window
             _hInst = (HINSTANCE)GetCurrentProcess();
             RECT rc = { 0, 0, (LONG)width, (LONG)height };
-            AdjustWindowRect(&rc, WS_OVERLAPPEDWINDOW, FALSE);
+            AdjustWindowRect(&rc, WS_OVERLAPPEDWINDOW, false);
```
TheMagnetar, line 230: It has BOOL as a parameter, therefore  it should be FALSE
```suggestion
            AdjustWindowRect(&rc, WS_OVERLAPPEDWINDOW, FALSE);
```

extensions/src/ACRE2Arma/common/directx/d3d_display.cpp
```diff
@@ -107,7 +107,7 @@ namespace acre {
             sd.OutputWindow = _hWnd;
             sd.SampleDesc.Count = 1;
             sd.SampleDesc.Quality = 0;
-            sd.Windowed = _fullscreen ? FALSE : TRUE;
+            sd.Windowed = !_fullscreen;
```
TheMagnetar, line 110: Windowed is expecting BOOL, therefore no changes are needed in here

extensions/src/ACRE2Arma/arma2ts/simplepipe_win32.cpp
```diff
@@ -131,7 +131,7 @@ bool PipeWin::OpenServer(HANDLE pipe, bool connect) {
 
 bool PipeWin::Write(const void* buf, size_t sz) {
     DWORD written = 0;
-    return (TRUE == ::WriteFile(pipe_, buf, sz, &written, NULL));
+    return (true == ::WriteFile(pipe_, buf, sz, &written, NULL));
```
TheMagnetar, line 134: WriteFile returns BOOL, therefore it should be compared against TRUE
```suggestion
    return (TRUE == ::WriteFile(pipe_, buf, sz, &written, NULL));
```

TheMagnetar (commented):
No need to change TRUE/FALSE. In all the cases, the variable is declared as BOOL, the argument expects BOOL or the function returns BOOL.

jonpas:
No further activity, feel free to re-open.

PR #1009, opened 2020-08-10
Create .clang-format

PabstMirror:
I think this should roughly match the style in the formatting guide (I'm no expert with this file-format)
Both VS and VScode w/ C++ Ext support it automatically

I don't really want to do anything extreme like format the entire codebase or enforce commits via validators
but I think it can be helpful to be able to auto-format sections.


PabstMirror:
replaced by #1010

PR #1010, opened 2020-08-11
Tools - Add Clang

TheMagnetar:
- Add Clang-format
- Add Clang-tidy


jonpas (approved):
Are we going to get commits with massive changes after this is merged? :D

extensions/src/.clang-format
```diff
@@ -0,0 +1,55 @@
+﻿---
```
jonpas: > `./extensions/src/.clang-format`

Contains BOM, should be removed.

PabstMirror (approved):

PR #1013, opened 2020-08-15
God Mode - Fix API function argument check

jonpas:
- API function argument check not allowing to set another code as dynamic group target function.


Sniperhid (approved):
Nice Fix!

Issue #574, opened 2018-08-18
Using Basic Mission Setup module to set default radios won't make any difference


baranelitez:
**Arma 3 Version:** `x.xx` (stable / rc / dev)
**CBA Version:** `3.8.0` (stable / dev + commit hash)
**ACRE2 Version:** `2.6.1.992` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- acre2
- ACE
- RHS
```

**Description:**
- Using Basic Mission Setup module to set default radios won't make any difference. The units all spawn with 343's regardless of their rank or type of unit (rhs or vanilla). Trying to put other radios into their inventories and launching the game won't make difference either. They still spawn with 343's.

**Steps to reproduce:**
- Add any unit in editor.
- Add Basic Mission Setup module into the game
- Change the value of "Default Radio 1" from ACRE_PRC343  to ACRE_PRC152
- Notice how units still spawn with 343's.

**Where did the issue occur?**
- Self-Hosted Multiplayer 

**Placed Modules:**
- Basic Mission Setup

TheMagnetar:
The modules were deprecated some versions ago and with 2.6 you should not see them in the editor.

 In addition we cannot speak for rhs units since they are not compatible with cba.

Maybe we can provide a setting for changing the basic radio. I think this will make a lot of sense. 

Nevertheless we would also appreciate you fill the template correctly so that we know what are your acre and cba versions

baranelitez:
I've updated the template. Not sure about the ACRE version though. It states nowhere in game, in launcher or in the workshop.

It's the latest version from the workshop though. And workshop states the latest change to this date (26 June 2018) was "2.6.1.992". Yet I'm still seeing the modules within the editor.

I would appreciate if there's a setting for changing the basic radio. Thank you for your interest and assistance.

jonpas:
Modules should be hidden on any new mission. If you already had it placed in a mission, it will appear visible. We will not fully remove them to not break backwards compatibility, but they will stay hidden. If they are indeed not hidden on a fresh mission, then that's a bug, let us know please.

https://github.com/IDI-Systems/acre2/blob/master/addons/api/fnc_setItemRadioReplacement.sqf#L12
This API function should let you change the default radio, just call it early in the mission. `initPlayerLocal.sqf` should do the trick perfectly.

We could make it a CBA setting, but that would lower the use-case for mission frameworks.

baranelitez:
thanks for the API function @jonpas . 

[...]

PR #1015, opened 2020-08-18
Radio - Replace inventory PFH with CBA loadout handler

Freddo3000:
- Change inventory monitoring PFH to instead use CBA loadout player eventhandler
- ~~Add setting for setting the replacement radio~~ #1283
- ~~Add option for simply removing the radio instead without replacement~~ #1283

image

Tested in vanilla arsenal


addons/sys_radio/initSettings.sqf
```diff
@@ -0,0 +1,19 @@
+private _allRadios = [] call EFUNC(api,getAllRadios);
+private _radioClasses = [""] + _allRadios # 0;
+private _radioNames = ["None"] + _allRadios # 1;
```
PabstMirror:
```suggestion
private _radioNames = [localize "str_a3_cfgglasses_none0"] + _allRadios # 1;
```

addons/sys_radio/XEH_preInit.sqf
```diff
@@ -6,6 +6,8 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
```
PabstMirror: remove lines 39-41 (var set by cba settings)

addons/sys_radio/initSettings.sqf
```diff
@@ -0,0 +1,19 @@
+private _allRadios = [] call EFUNC(api,getAllRadios);
```
PabstMirror: I don't think it will be a problem, but I'm not sure if order is guaranteed by this function between clients.
I think it is just based on mod load order, so could be different

PabstMirror:
maybe make a note in `acre_api_fnc_setItemRadioReplacement` that it's use is deprecated
I think manually changing variables that are cba_settings isn't good

jonpas:
Does this interfere with Basic Mission Setup module? This particular functionality of it is broken atm (ref. https://github.com/IDI-Systems/acre2/issues/574), but we should still think about it.

addons/api/CfgFunctions.hpp
```diff
@@ -29,7 +29,6 @@ class CfgFunctions {
 
             PATHTO_FNC(filterUnitLoadout);
 
-            PATHTO_FNC(setItemRadioReplacement);
```
jonpas, line 32: Should use https://github.com/IDI-Systems/acre2/blob/1f363d002b337c8451f6a09ec6fc5b355aca9e5c/addons/main/script_macros.hpp#L151.

addons/sys_radio/XEH_preInit.sqf
```diff
@@ -33,10 +35,6 @@ if (hasInterface) then {
 
     // this isn't used anymore i do not think?
```
jonpas: Remove the 2 lines of comments too.

addons/sys_radio/initSettings.sqf
```diff
@@ -0,0 +1,21 @@
+[{!isNil EFUNC(api,getAllRadios)},{
```
jonpas: Why would that ever not be available?

Ideally that function should be in `sys_core` and `api` should just be a wrapper with extra arguments checks. But this is not the only bad example of this in the current code base.

jonpas (changes requested):

addons/sys_radio/initSettings.sqf
```diff
@@ -0,0 +1,20 @@
+
```
jonpas:
```suggestion
```

addons/api/XEH_PREP.hpp
```diff
@@ -1,2 +1,5 @@
 // Helpers
 PREP(mapChannelFieldName);
+
+// Obsolete
+OBSOLETE_SYS(FUNC(setItemRadioReplacement),{params ['_radioType']; EGVAR(sys_radio,defaultItemRadioType) = _radioType;});
```
jonpas: Should use `CBA_settings_fnc_set`, otherwise it will get reverted saving any other settings.

jonpas (changes requested):
How does this work in combination with Basic Setup module? I think default radios in the module are kind of broken currently anyways.

jonpas:
> How does this work in combination with Basic Setup module? I think default radios in the module are kind of broken currently anyways.

This question remains and should be looked at. It kind of partly replaces that part of the module as far as I understand.

---

@Sniperhid can you look at recent changes regarding replacement of PFH with EH?

addons/sys_core/fnc_arsenalClose.sqf
```diff
@@ -31,3 +31,5 @@ private _weapons = [acre_player] call EFUNC(sys_core,getGear);
 ACRE_ARSENAL_RADIOS = [];
 GVAR(arsenalRadios) = [];
 GVAR(arsenalOpen) = false;
+
+[] call EFUNC(sys_radio,monitorRadiosHandler);
```
jonpas, line 37: This doesn't catch closing of ACE Arsenal or the likes.

Freddo3000:
Just as a heads up I haven't tested the last few commits in game

addons/ace_interact/XEH_postInitClient.sqf
```diff
@@ -8,6 +8,7 @@ if (!hasInterface) exitWith {};
 
 ["ace_arsenal_displayClosed", {
     EGVAR(sys_core,arsenalOpen) = false;
+    [] call DFUNC(monitorRadiosHandler);
```
jonpas, line 13: Leaving a note here: We have to think about API for this, copy-pasting those lines is far from ideal.

addons/api/fnc_basicMissionSetup.sqf
```diff
@@ -82,14 +82,15 @@ private _addRadios = {
     _player setVariable [QGVAR(basicMissionSetup), true, true];
 
     private _cleanRadioList = [];
-    if !("ACRE_PRC343" in _defaultRadios) then {
+    private _defaultRadio = QEGVAR(sys_radio,defaultItemRadioType) call CBA_settings_fnc_get;
```
jonpas: That would be just `EGVAR(sys_radio,defaultItemRadioType)`, you don't need to call `get`.

jonpas:
Will probably take a look at finishing this for 2.10.0.

jonpas:
This is almost done, but I don't want to risk it release in v2.11.0, will be part of dev-build after though.

jonpas:
> Add setting for setting the replacement radio
> Add option for simply removing the radio instead without replacement

Split that off into #1283 so we can merge the important part and leave below for further testing.

> Change inventory monitoring PFH to instead use CBA loadout player eventhandler

As noted in a comment above, problem is we are potentially not catching all cases of radio replacement with the handler. If we need a special call to handle it for Arsenal, then who knows how many 3rd party systems would need to do the same - that's not reliable. System needs to catch it all without a new API requiring this. PFH as it is shouldn't be expensive in most cases anyways.

PR #1033, opened 2020-10-12
API - Optimize `setPTTDelay`

jharmsen845:

- Allow us to now reach the statement on line 22, that was previously blocked by 
`if (!( _delay isEqualType 0)) exitWith { false };`.

- Valid `_delay` values will now proceed to line 24, where `ACRE_PTT_RELEASE_DELAY` will be set accordingly.

### DIFF
--- a/addons/api/fnc_setPTTDelay.sqf
+++ b/addons/api/fnc_setPTTDelay.sqf
@@ -19,8 +19,6 @@ params [
     ["_delay", 0, [0]]
 ];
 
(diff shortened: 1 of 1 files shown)


jonpas:
For reference `params` was updated in https://github.com/IDI-Systems/acre2/pull/588, since then this line was actually not necessary anymore as `params` takes care of doing type check already.

jonpas (approved):

PR #1034, opened 2020-10-18
API - Fix scheduler issue in `getRadioByType`

jokoho48:
- Fix a Issue with getRadioByType when Unit does not have a Radio of that type and is called from a Scheduled Enviroment

### DIFF
--- a/addons/api/fnc_getRadioByType.sqf
+++ b/addons/api/fnc_getRadioByType.sqf
@@ -40,6 +40,7 @@ if (_array isEqualType objNull) then {
     };
 } forEach _array;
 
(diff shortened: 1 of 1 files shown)


addons/api/fnc_getRadioByType.sqf
```diff
@@ -40,4 +40,7 @@ if (_array isEqualType objNull) then {
     };
 } forEach _array;
 
+// this prevents script errors in the Schedular when the unit does not have a radio and the _ret is nil
```
jonpas:
```suggestion
// Prevent script errors in the scheduler when the unit does not have a radio and `_ret = nil`
```

jonpas (approved):

PR #1035, opened 2020-11-07
Racks - Add an additional rack to helicopters

Mike-MF:
- Adds Rack 2 to helicopters
- Stringtabled Rack 2 on Planes.

### DIFF
--- a/addons/sys_rack/CfgVehicles.hpp
+++ b/addons/sys_rack/CfgVehicles.hpp
@@ -119,6 +119,16 @@ class CfgVehicles {
                 allowedPositions[] = {"driver", "gunner", "copilot"}; // Attack helicopters have "gunner" copilots usually
                 mountedRadio = "ACRE_PRC117F";
             };
+
+            class Rack_2: Rack_1 {
+                displayName = CSTRING(dash2);
+                shortName = CSTRING(dashShort);
+            };
+
+            class Rack_3: Rack_1 {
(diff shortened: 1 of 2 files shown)


jonpas:
We've discussed this one some issue before, but planes and helicopters should have at minimum 2 racks across the board. I am not sure on 3, since it's harder to delete one than add one, but we might as well?

BrettMayson:
Internally I've always had at least two on any military vehicle at large events. I think multiple for aircraft is a good step but I think 3 is unrealistic, most aircraft have COM1 and COM2. Not against per se, just a note.

jonpas:
A lot have 3 (A-10C, Huey, Gazelle) etc, but I tend to agree 2 is better overall.

BrettMayson:
Ah yeah good point, I'm fine with either though

jonpas (approved):

TheMagnetar (changes requested):
There is already upper dash and lower dash. Why add another term (Rack 2) just for helicopters?

jonpas:
> Why add another term (Rack 2) just for helicopters?

It doesn't. "Rack 2" already existed for planes as you can see in the config, it was just not stringtabled.

1/2 or Upper/Lower, doesn't make a difference to me, so whichever.

TheMagnetar (approved):
My bad. Mobile is not the best platform

Issue #555, opened 2018-06-30
Add function to set speaker mode on compatible radios


Sweedn:
Hello, could we have the possibilty to : 
- Set radios in speaker mode without any player action
- Add the possibilty to "lock" radio in speaker mode, interesting in TvT scenarios
- Perfect for designing mission when there isn't headset yet or too expensive for soldiers to afford (WW2 era or low-tech armies)

TheMagnetar:
Making an API function for configuring the speaker mode makes sense, on the other hand, locking a radio in speaker mode makes very little sense to me at the moment.

xetra11:
Yea speakermode would be pretty cool!

Ginganinja82:
This would be awesome !, even just in the init to set the speaker on load ? Idea is to have a "general [AI]" in base to broadcast what's going on in the field when new players join.. could be a nice feature for immersion ?

TheMagnetar:
I do not think that what you propose is possible

Jason-Terry:
I know this is only 23 days old, but isn't this feature working on the 152 and the 148?

If so, I wouldn't mind taking a stab at extending it to the 177f and 77, with some small direction I could also work on this and make a PR.

TheMagnetar:
There is no speaker directly on the 77 or the 117f. They would need to be implemented either as a component or through racks

Jason-Terry:
> There is no speaker directly on the 77 or the 117f. They would need to be implemented either as a component or through racks

I see. I know the AN/PRC-117 we used the in USMC had a speak attachement that you could use in replacement of the handset.

So we would need to implement a simple component, that can be linked to the 177f, and it just needs to replace the origin point of the 177f sounds?

I am going to dig around at some of the components already in game, see if I can't gather how they work as.

milivojm:
https://wiki.clearbackblast.com/guides:acrespeakers

148 and 152 have speakers but they don't work well as speakers. We've just tested it and people can hear transmission like it's in their head, not like speaker audio coming from the receiver. Don't know if to open a new issue or just let it all here.

jonpas:
Done in #1038.

PR #1038, opened 2020-11-11
API - Add Audio Source Control

Soldia1138:
- Add two API Functions for controling the Radio Audio Source for available Radios

**Currently four Radios have the possibility to switch the audio source:**
 * Possible Values for ACRE_PRC148: ["INTAUDIO", "TOPAUDIO", "TOPSIDETON", "SIDEAUDIO", "SIDESIDETON"]
 * Possible Values for ACRE_PRC152: ["INTAUDIO", "TOPAUDIO"]
 * Possible Values for ACRE_SEM52SL: ["INTSPEAKER", "HEADSET"]
 * Possible Values for ACRE_SEM70: ["INTSPEAKER", "HEADSET"]

### DIFF
--- a/addons/api/fnc_getRadioAudioSource.sqf
+++ b/addons/api/fnc_getRadioAudioSource.sqf
@@ -10,7 +10,7 @@
  * Audio Source <STRING>
  *
  * Example:
- * _audioSource = ["ACRE_PRC148_ID_123"] call acre_api_fnc_getRadioAudioSource;
+ * _audioSource = ["ACRE_PRC148_ID_123"] call acre_api_fnc_getRadioAudioSource
  *
  * Public: Yes
  */
@@ -19,6 +19,4 @@ params [
     ["_radioId", "", [""]]
 ];
 
-private _audioSource = [_radioId, "getState", "audioPath"] call EFUNC(sys_data,dataEvent);
-
-_audioSource
+[_radioId, "getState", "audioPath"] call EFUNC(sys_data,dataEvent)
--- a/addons/api/fnc_getRadioAudioSource.sqf
+++ b/addons/api/fnc_getRadioAudioSource.sqf
@@ -19,8 +19,6 @@ params [
     ["_radioId", "", [""]]
 ];
 
-if ( !(_radioId isEqualType "")) exitWith { -1 };
-
 private _audioSource = [_radioId, "getState", "audioPath"] call EFUNC(sys_data,dataEvent);
 
 _audioSource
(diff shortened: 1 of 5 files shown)


addons/api/fnc_setRadioAudioSource.sqf
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Public: Yes
+ */
+
+params [
+    ["_radioId", "", [""]],
+    ["_audioSource", "", [""]]
+];
+
+if ( !(_radioId isEqualType "")) exitWith { -1 };
```
jonpas: Can be removed as `params` takes care it.

addons/api/fnc_getRadioAudioSource.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * _audioSource = ["ACRE_PRC148_ID_123"] call acre_api_fnc_getRadioAudioSource;
+ *
+ * Public: Yes
+ */
+
+params [
+    ["_radioId", "", [""]]
+];
+
+if ( !(_radioId isEqualType "")) exitWith { -1 };
```
jonpas: Can be removed as `params` takes care it.

jonpas (dismissed):

jonpas:
Closes #555 without adding the ability to lock it.

addons/api/fnc_setRadioAudioSource.sqf
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+params [
+    ["_radioId", "", [""]],
+    ["_audioSource", "", [""]]
+];
+
+private _radioType = [_radioId] call acre_api_fnc_getBaseRadio;
+private _success = false;
+
```
PabstMirror, line 29: This would help to make the API easier to use by just letting users just pass a `true` to set speaker
but also adds clutter to the function and might not be needed

```
if (_audioSource isEqualTo true) then {
    _audioSource = switch (_radioType) do { 
        case "ACRE_PRC148";
        case "ACRE_PRC152": {"INTAUDIO"};
        case "ACRE_SEM70";
        case "ACRE_SEM52SL": {"INTSPEAKER"};
        default {"#"};
    };
};
```

jonpas: That should be a separate API function (wrapper of this one for example).

Soldia1138: I will add another function to enable/disable the speaker

jonpas: @Soldia1138 Added, please review.

PabstMirror (approved):

PR #1044, opened 2020-11-29
Core - Change Babel default keybind to fix ACE Interact conflict

Ceebox:
I have modified the default keybind for the Babel system, as it conflicts with ACE INTERACT. This mod is built with CBA3 already, which means that you know there is a high likelihood that the user is going to use ACE. By changing this, it removes the conflict.


jonpas (commented):
Good with me, but would like input from other @IDI-Systems/acre-managers.

PabstMirror (approved):

TheMagnetar (approved):

PR #1051, opened 2021-01-08
Tools - Use hemtt in build.py if installed

PabstMirror:
new version of makepbo can't build pbos with non-binerized p3ds

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -9,6 +9,18 @@ MAINPREFIX = "idi"
 PREFIX = "acre_"
 ##########################
 
+def tryHemttBuild(projectpath):
+    hemttExe = os.path.join(projectpath, "hemtt.exe")
+    if os.path.isfile(hemttExe):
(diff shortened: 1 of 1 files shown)


jonpas (approved):
There is already `build.bat` in root of project, which we should promote to use instead of `build.py`. But this works as an intermediary solution until we are 200% HEMTT is completely ok.

PabstMirror:
This just runs `hemtt.exe pack` which takes less than a second on my machine
build.bat runs `hemtt.exe build` which binerizes p3ds and takes about 18 sec (but instant if no changes have been made)

Just plain old packing is what build.py has done in the past, so this is just meant to mirror that

jonpas:
Right, fairly sure I wrote that `build.bat` before there was a proper `hemtt pack`. It should probably use that as the release (coming CI) calls HEMTT directly.

jonpas:
Feel free to merge this though!

PR #1053, opened 2021-01-11
Add ACE Arsenal Stats

### LINKED ISSUE #525: Add additional ACE Arsenal Stats
- Power
  - [General City and Perfect Ranges (?)](http://acre2.idi-systems.com/wiki/user/basic-concepts#radios-and-classes)
- Frequency Range
- Amount of channels (?)

Any other ideas, bring them forward!

PabstMirror:
Close #525

### DIFF
--- a/addons/sys_prc117f/CfgWeapons.hpp
+++ b/addons/sys_prc117f/CfgWeapons.hpp
@@ -20,6 +20,11 @@ class CfgWeapons {
         class Library {
             libTextDesc = QUOTE(NAME_PRC117F);
         };
+        
+        EGVAR(arsenalStats,frequencyMin) = 30e6;
+        EGVAR(arsenalStats,frequencyMax) = 512e6;
+        EGVAR(arsenalStats,transmitPower) = 20000;
+        EGVAR(arsenalStats,effectiveRange) = "20+km (10-20km)";
     };
 
     RADIO_ID_LIST(ACRE_PRC117F)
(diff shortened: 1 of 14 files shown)


addons/sys_prc117f/CfgWeapons.hpp
```diff
@@ -20,6 +20,11 @@ class CfgWeapons {
         class Library {
             libTextDesc = QUOTE(NAME_PRC117F);
         };
+        
+        EGVAR(arsenalStats,frequencyMin) = 30e6;
+        EGVAR(arsenalStats,frequencyMax) = 512e6;
+        EGVAR(arsenalStats,transmitPower) = 20000;
+        EGVAR(arsenalStats,effectiveRange) = "20+km (10-20km)";
```
jonpas, line 27: I am not sure I like placing those arbitrary numbers here. It's clear it's for the Arsenal, so that's good, but can we get those from actual radio data that's already defined somehow? @Sniperhid @Soldia1138

PabstMirror: 148 seems to validate data with ` if (_newVal > 512 || {_newVal < 30}) then {`
but 152 allows you to enter anything 
image
117 same
image

could add new (non-arsenal) configs for min/max frequency for user input
but I don't know how the real radios handle this

jonpas: @IDI-Systems/acre-managers ?

jonpas: I think the way to go about this is make those not Arsenal-specific, but actually use those config entries to limit radio operation (like 148 does).

jonpas (approved):
This is all just internal, so should we change it at some point as discussed above, there won't be any issues.

PR #1060, opened 2021-02-06
Configs - Convert all true/false in configs to 0/1 

PabstMirror:
and minor cleanup of units/weapons[]


TheMagnetar (approved):

jonpas (approved):

jonpas:
Even better for HEMTT - next build!

PR #1062, opened 2021-02-12
CI - Add release and development build workflow

jonpas:
- Add full release build CI system
  - Extensions and Binarization included!
  - Builds published to FTP, Steam Workshop and GitHub

Updated documentation will follow in a separate PR after this gets merged and tested for confirmed functionality. This does not change the release process itself, so the current documentation is still accurate, however it does use HEMTT which we will prioritize over Mikero Tools after fully confirmed working release builds with this PR.


TheMagnetar (approved):
Looks good!

PR #1072, opened 2021-03-05
SEM70 - Fix duplicated knob picture

PabstMirror:
5 was a duplicate of 4
used 3's outer (outer shells just alternate odd/even)
and then rotated the inner thing 22.5 deg from 4's


jonpas (approved):
Lovely.

PR #1091, opened 2021-04-26
API - Improve and Fix some API functions

jokoho48:
- make use of select CODE in getAllRadiosByType
- make use of findIf in getRadioByType, hasBaseRadio, hasKindOfradio, hasRadio, initVehicleRacks and setMultiPushToTalkAssignment
- optimise isRadio my removing code that cant get reached
- fix wrong return type in setRadioChannel
- fix setMultiPushToTalkAssignment checking current assignment and not the new one
- fix tests in api

### DIFF
--- a/addons/api/fnc_getAllRadiosByType.sqf
+++ b/addons/api/fnc_getAllRadiosByType.sqf
@@ -23,8 +23,6 @@ params [
     ["_array", [], [[], objNull]]
 ];
 
-private _ret = [];
-
 if (_array isEqualType objNull) then {
     _array = _array call EFUNC(sys_core,getGear);
 } else {
@@ -33,11 +31,4 @@ if (_array isEqualType objNull) then {
     };
 };
 
-{
-    private _radioId = _x;
-    if ([_radioId, _radioType] call FUNC(isKindOf)) then {
-        _ret pushBackUnique _radioId;
-    };
-} forEach _array;
-
-_ret
+_array select {[_x, _radioType] call FUNC(isKindOf);};
--- a/addons/api/fnc_getAllRadiosByType.sqf
+++ b/addons/api/fnc_getAllRadiosByType.sqf
@@ -31,4 +31,5 @@ if (_array isEqualType objNull) then {
     };
 };
 
-_array select {[_x, _radioType] call FUNC(isKindOf)}
+_array = _array select {[_x, _radioType] call FUNC(isKindOf)};
+_array arrayIntersect _array;
--- a/addons/api/fnc_getAllRadiosByType.sqf
+++ b/addons/api/fnc_getAllRadiosByType.sqf
@@ -31,4 +31,4 @@ if (_array isEqualType objNull) then {
     };
 };
 
-_array select {[_x, _radioType] call FUNC(isKindOf);};
+_array select {[_x, _radioType] call FUNC(isKindOf)}
--- a/addons/api/fnc_getRadioByType.sqf
+++ b/addons/api/fnc_getRadioByType.sqf
@@ -23,8 +23,6 @@ params [
     ["_array", [], [[], objNull]]
 ];
 
-private _ret = nil;
-
 if (_array isEqualType objNull) then {
     _array = _array call EFUNC(sys_core,getGear);
 } else {
@@ -33,14 +31,8 @@ if (_array isEqualType objNull) then {
     };
 };
 
-{
-    private _radioId = _x;
-    if ([_radioId, _radioType] call FUNC(isKindOf)) exitWith {
-        _ret = _radioId;
-    };
-} forEach _array;
-
-// Prevent script errors in the scheduler when the unit does not have a radio and `_ret = nil`
-if (isNil "_ret") exitWith { nil };
-
-_ret
+private _index = _array findIf {
+    [_x, _radioType] call FUNC(isKindOf);
+};
+if (_index == -1) exitWith {nil};
+_array select _index;
--- a/addons/api/fnc_getRadioByType.sqf
+++ b/addons/api/fnc_getRadioByType.sqf
@@ -35,4 +35,4 @@ private _index = _array findIf {
     [_x, _radioType] call FUNC(isKindOf);
 };
 if (_index == -1) exitWith {nil};
-_array select _index;
+_array select _index
--- a/addons/api/fnc_hasBaseRadio.sqf
+++ b/addons/api/fnc_hasBaseRadio.sqf
@@ -22,6 +22,6 @@ params [
 if (isNull _unit) exitWith {false};
 
 
-private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isBaseRadio);};
+private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isBaseRadio)};
 
-_index != -1;
+_index != -1
--- a/addons/api/fnc_hasBaseRadio.sqf
+++ b/addons/api/fnc_hasBaseRadio.sqf
@@ -21,12 +21,7 @@ params [
 
 if (isNull _unit) exitWith {false};
 
-private _ret = false;
 
-{
-    private _weapon = _x;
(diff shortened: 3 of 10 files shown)


addons/api/fnc_setRadioChannel.sqf
```diff
@@ -21,7 +21,7 @@ params [
     ["_channelNumber", 0, [0]]
 ];
 
-if ( !(_radioId isEqualType "")) exitWith { -1 };
+if (!(_radioId isEqualType "")) exitWith { false };
```
jonpas:
```suggestion
if !(_radioId isEqualType "") exitWith { false };
```

addons/api/fnc_hasRadio.sqf
```diff
@@ -21,12 +21,6 @@ params [
...
+private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isRadio);};
 
-{
-    private _weapon = _x;
-    _ret = [_weapon] call FUNC(isRadio);
-    if (_ret) exitWith { };
-} foreach ([_unit] call EFUNC(sys_core,getGear));
-
-_ret
+_index != -1;
```
jonpas:
```suggestion
_index != -1
```

addons/api/fnc_hasKindOfRadio.sqf
```diff
@@ -17,27 +17,20 @@
...
-        _ret = [_weapon, _type] call FUNC(isKindOf);
-        if (_ret) exitWith { };
-    } foreach _weaponArray;
-};
+if (_type in _array) exitWith {true};
+
+private _index = _array findIf {[_x, _type] call FUNC(isKindOf)};
 
-_ret
+_index != -1;
```
jonpas:
```suggestion
_index != -1
```

addons/api/fnc_hasBaseRadio.sqf
```diff
@@ -21,12 +21,7 @@ params [
...
 
-{
-    private _weapon = _x;
-    _ret = [_weapon] call FUNC(isBaseRadio);
-    if (_ret) exitWith { _ret };
-} foreach ([_unit] call EFUNC(sys_core,getGear));
+private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isBaseRadio);};
 
-_ret
+_index != -1;
```
jonpas:
```suggestion
_index != -1
```

addons/api/fnc_hasBaseRadio.sqf
```diff
@@ -21,12 +21,7 @@ params [
...
 if (isNull _unit) exitWith {false};
 
-private _ret = false;
 
-{
-    private _weapon = _x;
-    _ret = [_weapon] call FUNC(isBaseRadio);
-    if (_ret) exitWith { _ret };
-} foreach ([_unit] call EFUNC(sys_core,getGear));
+private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isBaseRadio);};
```
jonpas:
```suggestion
private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isBaseRadio)};
```

addons/api/fnc_hasRadio.sqf
```diff
@@ -21,12 +21,6 @@ params [
 
 if (isNull _unit) exitWith {false};
 
-private _ret = false;
+private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isRadio);};
```
jonpas:
```suggestion
private _index = ([_unit] call EFUNC(sys_core,getGear)) findIf {_x call FUNC(isRadio)};
```

addons/api/fnc_getRadioByType.sqf
```diff
@@ -33,14 +31,8 @@ if (_array isEqualType objNull) then {
...
-
-// Prevent script errors in the scheduler when the unit does not have a radio and `_ret = nil`
-if (isNil "_ret") exitWith { nil };
-
-_ret
+private _index = _array findIf {
+    [_x, _radioType] call FUNC(isKindOf);
+};
+if (_index == -1) exitWith {nil};
+_array select _index;
```
jonpas:
```suggestion
_array select _index
```

addons/api/fnc_getAllRadiosByType.sqf
```diff
@@ -33,11 +31,4 @@ if (_array isEqualType objNull) then {
...
 
-{
-    private _radioId = _x;
-    if ([_radioId, _radioType] call FUNC(isKindOf)) then {
-        _ret pushBackUnique _radioId;
-    };
-} forEach _array;
-
-_ret
+_array select {[_x, _radioType] call FUNC(isKindOf);};
```
jonpas:
```suggestion
_array select {[_x, _radioType] call FUNC(isKindOf)}
```

jonpas (changes requested):

addons/api/fnc_setRadioChannel.sqf
jokoho48: I miss read the coding guidelines... 🤦

addons/api/fnc_getAllRadiosByType.sqf
```diff
@@ -33,11 +31,4 @@ if (_array isEqualType objNull) then {
...
 
-{
-    private _radioId = _x;
-    if ([_radioId, _radioType] call FUNC(isKindOf)) then {
-        _ret pushBackUnique _radioId;
-    };
-} forEach _array;
-
-_ret
+_array select {[_x, _radioType] call FUNC(isKindOf)}
```
jonpas: This doesn't filter for unique items, but I guess this is fine as that shouldn't be possible to be had (without other issues).

addons/api/fnc_isRadio.sqf
```diff
@@ -16,14 +16,10 @@
...
 
 params [
-    ["_weapon", "", [""]]
+    ["_radio", "", [""]]
 ];
 
-// check that it at least inherits from ACRE_BaseRadio *AND* has a unique ID
-// If its scoped 1, that means its either a base or not a radio.
-private _flag = getNumber (configFile >> "CfgWeapons" >> _weapon >> "acre_isRadio");
-private _id = getNumber (configFile >> "CfgWeapons" >> _weapon >> "acre_uniqueId");
```
jonpas, line 25: Where did that go? I'd also keep the comment above.

addons/api/fnc_getAllRadiosByType.sqf
jokoho48: this can only happen if a user adds in duplicated items in the beginning.
we could run an arrayIntersect on the return, but I don't know if that is really worth it.
but we may want to remove the `unique` out of the description.

jonpas: I think running `arrayIntersect` would be fine.

addons/api/fnc_isRadio.sqf
jokoho48: the comment above looks like from a version where the code worked differently because
1. `getNumber` never returns nil so the `if isNil exitWith` would never trigger
2. the comment says that if it is scoped 1 it is a base radio or not a radio what is not reflected in the code below because there  it actually got check if it was one

so that comment basically contradicts what the code did before and also the Function Description says.

the code now returns true on Base Class Radios for example an ACRE_PRC117F but also on an ACRE_PRC117F_id_1.

before it only returned true on an ACRE_PRC117F_id_1 but not on an ACRE_PRC117F

jonpas: Thinking about this I think this function should only return `true` for actual unique ID'd radios. Only those are actual "radios", base classes are just unusable items in ACRE's world. So I would modify the header to reflect that and keep the logic the same for that part. 

Besides, there is another API function to check if it's a base radio `isBaseRadio`, which is the opposite of what I would expect this function to do.

I guess scope part is redundant.

jonpas (approved):

PabstMirror (approved):

PR #1092, opened 2021-04-26
General - Make use of `configOf`

jokoho48:
- Makes use of ConfigOf

### DIFF
--- a/addons/sys_core/fnc_findOcclusion.sqf
+++ b/addons/sys_core/fnc_findOcclusion.sqf
@@ -74,12 +74,11 @@ private _intersectObjects = lineIntersectsObjs [_startPos, _endPos, _vehicleUnit
 
 
 {
-    private _typeString = typeOf _x;
-    if (_typeString != "") then {
+    if !(isNull _x) then {
 
         call {
-            if (_typeString isKindOf "House") exitWith { _thicknessFactor = _thicknessFactor + 0.3; };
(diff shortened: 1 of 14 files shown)


addons/sys_core/fnc_findOcclusion.sqf
```diff
@@ -74,12 +74,11 @@ private _intersectObjects = lineIntersectsObjs [_startPos, _endPos, _vehicleUnit
 
 
 {
-    private _typeString = typeOf _x;
-    if (_typeString != "") then {
+    if !(isNull _x) then {
```
jonpas:
```suggestion
    if (!isNull _x) then {
```

jonpas (changes requested):
Awesome!

jonpas (approved):

addons/sys_core/fnc_findOcclusion.sqf
```diff
@@ -74,12 +74,11 @@ private _intersectObjects = lineIntersectsObjs [_startPos, _endPos, _vehicleUnit
 
 
 {
-    private _typeString = typeOf _x;
-    if (_typeString != "") then {
+    if (!isNull _x) then {
```
PabstMirror: I think we want to revert this change
it's for detecting map objects that don't have a type

jokoho48:
```suggestion
    if ((typeOf _x) != "") then {
```

PabstMirror (approved):

PR #1103, opened 2021-05-12
Extension - Fix potential named pipe buffer overflow

iljavs:
There's no room allocated for a terminating 0-byte at the end of mBuffer. This could lead to an off-by-one buffer overflow on line 268. This is fixed by adding room (+1) for a 0-byte on line 215 when performing the buffer allocation.


PabstMirror:
I think we'd need to change line 270 as well

jonpas:
I am somewhat unsure of this.

jonpas:
@PabstMirror Why line 270?

I digged into this, this will probably never ever be triggered because buffer size is so much bigger (10240) and usual read length is low (under 1000).

However I would instead prefer `- 1` on L260. See #1150.

jonpas:
Alternative merged. Thank you for your contribution either way, and for spotting the potential error!

PR #1105, opened 2021-05-20
WIP: Headless broadcast

PabstMirror:
Had a feature request to allow playing music via radios.
This allows an AI unit to talk or speak on radio with the use of 
some external ts client that is outputting sound e.g.:  https://github.com/Splamy/TS3AudioBot

https://www.youtube.com/watch?v=MTFx7xuhRjE

Example use:
```
[q2, "Botty", 0, 0, ""] call acre_sys_headless_fnc_start // has AI unit q2 start speaking with "Botty"'s Audio 
```

Ultimately, I think this feature is just too niche (and kind of hackey) to add publicly.
But it does seem to be a useful tool in debugging.

also fun bug:
acre just expects mono microphone data, but TS is capable of stereo
without the mono fix, everything was pitch shifted lower because the sound data is interleaved [left, right, left]


jonpas:
> Ultimately, I think this feature is just too niche (and kind of hackey) to add publicly.

Hackey part yes, otherwise I would say this fits public API just fine.

> also fun bug:

Maybe we can split that off to include in next release? Unless you intend to finish this.

JordannDev:
Is there any intention of this being merged and put into release? I think this could be very useful for scenarios where game masters want ambient radio traffic over usable radios (as an example use case).

Jsmuk:
Are there any plans to revive this and get it in - even as a dev / undocumented feature? It'll be useful for developing, but I can see the gameplay benefits of it as well. 

I've resolved the merge conflicts in my fork [here](https://github.com/Jsmuk/acre2/tree/headlessBroadcast) if that helps at all.

addons/sys_headless/fnc_start.sqf
```diff
@@ -0,0 +1,96 @@
...
+ * Public: No
+ */
+
+if (!hasInterface) exitWith {};
+
+params ["_unit", "_tsDisplayName", "_languageId", "_speakingType", "_radioId"];
+TRACE_4("start",_unit,_tsName,_languageId,_speakingType,_radioId);
+
+if (!alive _unit) exitWith { ERROR_1("bad unit",_this);  false };
+if (!isNil {_unit getVariable [QGVAR(keepRunning), nil]}) exitWith {  ERROR_1("unit is already active",_this); false };
```
rautamiekka, line 30: As of 'Arma 3' 2.18, the nil check could be simplified:
```suggestion
if !(_unit isNil QGVAR(keepRunning)) exitWith {  ERROR_1("unit is already active",_this); false };
```

addons/sys_headless/fnc_stop.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * [q2] call acre_sys_headless_fnc_stop
+ *
+ * Public: No
+ */
+if (!hasInterface) exitWith {};
+
+params ["_unit"];
+TRACE_1("stop"_unit);
+
+if (isNil {_unit getVariable [QGVAR(keepRunning), nil]}) exitWith { ERROR_1("unit is not active",_this); };
```
rautamiekka, line 22: Likewise this could be simplified:
```suggestion
if (_unit isNil QGVAR(keepRunning)) exitWith { ERROR_1("unit is not active",_this); };
```

VelonacepsCalyxEggs:
Yeah, this PR opens a lot of possibilities for a wide variety of scenarios, like stations that people can listen to to get important mission related information by deciphering the signal or similar. Or just straight up playing radio on a specific frequency.
I really hope this gets merged, this is going to be a game changer in my opinion.

PR #1108, opened 2021-05-23
Plugin - Update to latest Mumble API spec

Krzmbrzl:
- Update the used Mumble API to the latest (stable) plugin API v1.0.x

This work is WIP but I wanted to have someone actually knowing the ACRE source code to have a glimps at my doings early on in order to not develop in the complete wrong direction.
Also I want to borrow your CI since I don't like developing on a Windows machine xD

Related: #980, #577


extensions/src/ACRE2Mumble/MumbleCallbacks_init.cpp
```diff
@@ -9,20 +9,17 @@
 #define FROM_PIPENAME "\\\\.\\pipe\\acre_comm_pipe_fromTS"
 #define TO_PIPENAME   "\\\\.\\pipe\\acre_comm_pipe_toTS"
 
-extern MumbleAPI mumAPI;
+extern MumbleAPI_v_1_0_x mumAPI;
+// TODO: Should these special values be a named variable?
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleCallbacks_init.cpp
```diff
@@ -38,10 +35,11 @@ void mumble_registerAPIFunctions(struct MumbleAPI api) {
...
 
-mumble_error_t mumble_init(mumble_connection_t connection) {
-    if (connection != -1) {
-        activeConnection = connection;
-    }
+mumble_error_t mumble_init(mumble_plugin_id_t id) {
+	pluginID = id;
+	
+	// TODO: Is initializing the active server connection required at this point?
+	// If so it must be done via an API call
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleCallbacks_init.cpp
```diff
@@ -51,6 +49,7 @@ void mumble_onServerSynchronized(mumble_connection_t connection) {
 
     // set ID on every new connection
     acre::id_t clientId = 0;
+	// TODO: Missing error handling of API call
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleCallbacks_pluginEvents.cpp
```diff
@@ -6,8 +6,9 @@
 //
 // Handle a command event
 //
-bool mumble_onReceiveData(mumble_connection_t connection, mumble_userid_t sender, const char *data, size_t dataLength, const char *dataID) {
+bool mumble_onReceiveData(mumble_connection_t connection, mumble_userid_t sender, const uint8_t *data, size_t dataLength, const char *dataID) {
     if ((dataLength > 0U) && CEngine::getInstance()->getExternalServer()) {
+		// TODO: Cast is probably unnecessary
         CEngine::getInstance()->getExternalServer()->handleMessage((unsigned char *) data, dataLength);
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleCallbacks_sound.cpp
```diff
@@ -16,7 +16,10 @@
 
 using LIMITER = std::numeric_limits<int16_t>;
 
-bool mumble_onAudioSourceFetched(float *outputPCM, uint32_t sampleCount, uint16_t channelCount, bool isSpeech, mumble_userid_t userID) {
+bool mumble_onAudioSourceFetched(float *outputPCM, uint32_t sampleCount, uint16_t channelCount, uint32_t sampleRate, bool isSpeech, mumble_userid_t userID) {
+	// TODO: Do you need this for something?
+	(void) sampleRate;
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleCallbacks_speaking.cpp
```diff
@@ -40,6 +40,7 @@ void mumble_onUserTalkingStateChanged(mumble_connection_t connection, mumble_use
             return;
         }
 
+		// TODO: Is the TalkingState enum defined somewhere in ACRE? 'cause it is not defined in Mumble
         if (status == TalkingState::PASSIVE || status == TalkingState::INVALID) {
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleClient.cpp
```diff
@@ -71,12 +71,13 @@ acre::Result CMumbleClient::start(const acre::id_t id_) {
...
 bool CMumbleClient::getVAD() {
-    transmission_mode_t transmitMode;
+    mumble_transmission_mode_t transmitMode;
     const mumble_error_t err = mumAPI.getLocalUserTransmissionMode(pluginID, &transmitMode);
     if (err != ErrorCode::EC_OK) {
         return false;
     }
 
+	// TODO: Is this enum defined somewhere in ACRE?
     return transmitMode == TransmissionMode::TM_VOICE_ACTIVATION;
```
Krzmbrzl: :point_up:

extensions/src/ACRE2Mumble/MumbleClient.cpp
```diff
@@ -324,6 +325,7 @@ uint64_t CMumbleClient::findChannelByNames(std::vector<std::string> details_) {
             char *channelName = nullptr;
 
             if (mumAPI.getChannelName(pluginID, activeConnection, channelId, &channelName) == STATUS_OK) {
+				// TODO: Does std::string take care of deleting the allocated name?
                 std::string channelNameString(channelName);
```
Krzmbrzl: :point_up:

Krzmbrzl (commented):
I have a few questions that came up when working on this. I have marked them in my review and would appreciate if someone could help me figure these out :)

extensions/src/ACRE2Mumble/MumbleCallbacks_init.cpp
jonpas: More error handling can probably be done in multiple API call locations.

extensions/src/ACRE2Mumble/MumbleClient.cpp
jonpas: Yes, its destructor will deallocate it.

extensions/src/ACRE2Mumble/MumbleClient.cpp
jonpas: Mumble includes from that time: https://github.com/IDI-Systems/acre2/blob/a0572a4dd4fe871b8b896e544300605f4cd0d87c/extensions/src/ACRE2Mumble/mumble_includes/PluginComponents.h#L55-L62

extensions/src/ACRE2Mumble/MumbleCallbacks_speaking.cpp
jonpas: Mumble includes from that time: https://github.com/IDI-Systems/acre2/blob/a0572a4dd4fe871b8b896e544300605f4cd0d87c/extensions/src/ACRE2Mumble/mumble_includes/PluginComponents.h#L46-L53

jonpas:
Please do not force push at all here, it makes reviewing anything impossible to do in a conceivable amount of time.

> Also I want to borrow your CI since I don't like developing on a Windows machine xD

We don't currently build extensions in pull requests CI due to certain dependencies.

veteran29:
@Krzmbrzl does Mumble support "server-side muting" (voice data not being sent for locally muted clients), from what I remember it was the biggest "deal breaker" when compared to TS, this might matter for clients with weaker connections when playing bigger operations with a lot of players.

Sorry for hijacking the PR but ACRE has moved it's slack to channel on ACE slack and I can not find you there ;(

extensions/src/ACRE2Mumble/MumbleCallbacks_speaking.cpp
Krzmbrzl: But that is not an enum class either so the way it is referenced should have not worked back then as well :eyes:

Krzmbrzl:
@veteran29 Yeah I only joined the ACE slack after you showed me that it might still be beneficial to do so :D 

To answer the question: No Mumble currently does not support that but I would assume this to be a feature that could be (easily) added in the future.

extensions/src/ACRE2Mumble/MumbleCallbacks_speaking.cpp
jonpas: Fairly sure the reason it works is because it's NOT an `enum class`, which would throw an error. With regular `enum` it works, but it is not a suggested use IIRC.

extensions/src/ACRE2Mumble/MumbleClient.cpp
Krzmbrzl: @jonpas are you sure? From what I see at https://www.cplusplus.com/reference/string/string/string/ it seems that this constructor only copies the given C-string.
In general I think std::string can't even properly discard this String as it needs to be done via Mumble's API anyway... I think that's a bug :eyes:

Krzmbrzl: I changed the implementation to what _I_ think is correct at this point. If that is not what is wanted/needed here, we can talk about it again in the following review :)

Krzmbrzl:
Given that this compiles now, I _think_ this is all that is needed from a purely API update point of view :thinking:

TheMagnetar:
Not really... compile does not mean ready. Only when it compiles and it works as expected can we label it as ready.

Krzmbrzl:
> all that is needed **from a purely API update point of view**

That is all I said. And what I wanted to achieve with this statement is to state the the PR is ready enough for someone with knowledge of the ACRE plugin architecture to have a look at the open questions that I have asked here :)

extensions/src/ACRE2Mumble/mumble_includes/MumbleAPI_v_1_0_x.h
```diff
@@ -17,9 +17,9 @@
 #define MUMBLE_PLUGIN_API_MINOR_MACRO 0
 #define MUMBLE_PLUGIN_API_PATCH_MACRO 0
 
-constexpr int32_t MUMBLE_PLUGIN_API_MAJOR        = MUMBLE_PLUGIN_API_MAJOR_MACRO;
-constexpr int32_t MUMBLE_PLUGIN_API_MINOR        = MUMBLE_PLUGIN_API_MINOR_MACRO;
-constexpr int32_t MUMBLE_PLUGIN_API_PATCH        = MUMBLE_PLUGIN_API_PATCH_MACRO;
+const int32_t MUMBLE_PLUGIN_API_MAJOR            = MUMBLE_PLUGIN_API_MAJOR_MACRO;
```
TheMagnetar: why do you remove the constexpr?

Krzmbrzl: Because the API is supposed to the C-compatible but `constexpr` is cpp-only. Therefore having it in there prevented using the header files in C

EDIT: To be more clear on this: This has been changed in the upstream header files along with the name-prefixing and the change you are seeing here is merely the result of replacing the "old" headers withe the new ones.
Thus the changes to the Mumble header files as such are unrelated to the ACRE code.

Krzmbrzl:
As I wrote on Slack already I will not be able to test these changes within Arma in the near future. Therefore I would appreciate it if someone else could try it out for me and report back here whether it worked or not.

The ACRE build from this PR including the mumble plugin can be downloaded from [here](http://178.254.32.29/nextcloud/index.php/s/tCmAXsi6NBrf7KN).

Note that this is an older version than the latest ACRE release due to the Mumble branch being behind the current master branch.

The steps to test this would be the following:
1. Download and install the latest Mumble **server** snapshot from https://www.mumble.info/downloads/#development-snapshots - if you had an earlier version of Mumble installed already, uninstall it first. For the client installer, please see my [post below](https://github.com/IDI-Systems/acre2/pull/1108#issuecomment-873599571).
2. Install the ACRE2 mod and start Arma with it as usual
3. The above step should have auto-installed the Mumble plugin. If this was the case, you will find an ACRE2 entry in Mumble's settings under Settings > Plugins. If this is not the case, click on `Install plugin...` and select the respective DLL from the ACRE2 mod directory under `plugin/mumble` and install it that way.
4. Finally you need to set up a local server. A remote server works as well but I am assuming you don't have a Mumble server running already and therefore the local version is easiest. In order to start the server, start `Mumble Server` (or maybe in this version it is still called `Murmur`) from the start menu. The server will run in the background and you can now connect to it from Mumble.
This can be done by opening the server browser in Mumble and either the local server is directly listed there or you'll have to add a new server with address `localhost`.

After connecting to your new server (which will cause a warning about a self-signed certificate that you can safely accept) you'll have a server with a single channel in it.

If you require more channels for your test, you have to connect to the server using the `SuperUser` username. This will prompt for a password. This password you can set beforehand by starting the server via the command line and feeding it the `-supw <your password>` argument.
Thus you can set the password to e.g. 123:
```
murmur.exe -supw 123
```
With the SuperUser account you can create channels using the context menu but for actually testing ACRE, you'll have to switch to a different username again.

Note: Switching the username means reconnecting to the server using a different username. If you connected via an entry in your favorites list, you'll have to edit the entry in order to change the username.

Krzmbrzl:
As has been reported by @veteran29 using the Mumble ACRE plugin with the latest Mumble snapshot will cause a crash on Mumble's side as soon as ACRE connects to Arma.

This is due to a bug in Mumble related to the order of parameters in an audio callback function of the plugin.

I have fixed the bug and you can find an installer for the fixed version [here](https://www.dropbox.com/s/ir31dekdtguintp/mumble_client-1.5.0.0-x64.msi?dl=1).
Note however that this is a snapshot of Mumble 1.5.0 and therefore you want to uninstall this again once you are done with testing, if you plan on installing Mumble 1.4.0 which should be released in the near future.

Krzmbrzl:
Testing performed so far shows that the plugin seems to be working with the fixed Mumble build.

However there was a combined freeze of Arma and Mumble observed at some point. My best guess so far is that the plugin is producing a deadlock by calling a Mumble API function which are now synchronized and blocking.
I'll have to check the individual API calls again to see if I can spot an issue...

Krzmbrzl:
Just as an update here: I identified the deadlock and traced its cause down but am still considering what the best fix for it may be (see also my post on the ACRE slack).

Krzmbrzl:
Okay I think I found a good solution for the deadlock. The Mumble plugin is now using its own event loop that processes all events asynchronously and thus Mumble's main thread does not get locked for that.
There is an edge case when the plugin is shut down. Potentially there could still be functions running asynchronously but we can't just let them running because using the API (which they may) is UB once the shutdown function returns control to Mumble.

Therefore I have tweaked Mumble's side a bit such that deadlocks like the one we saw here should in general become impossible. Instead there will be a noticeable delay and after that the function will return a timeout error.

----

Here are the updated builds of both the Mumble client and ACRE (note that for the server you can still use the 1.4.0 snapshot server without issues):
- [ACRE](http://178.254.32.29/nextcloud/index.php/s/G5MgMCoYSNsrZjq)
- [Mumble client](http://178.254.32.29/nextcloud/index.php/s/tCmAXsi6NBrf7KN)

Krzmbrzl:
Okay I finally got to perform a small test yesterday and I ended up with freezing Arma + Mumble. This looked like some sort of deadlock again.

I'll have to investigate more in order to try to find the root cause of this.

Krzmbrzl:
So, I'm back from the dead :wave: 
As promised, I investigated the observed deadlock and I think I got rid of it now (see commit message of b72aca68160f349820cb02f619f6201f5f6025cc for details). I am now trying to find a few victims that I can test this a bit more thoroughly with and report back here, once I have the results.

Note for self: Somewhere in the code, there is an API call to `getAllChannels`, that does not call `mumble_releaseResource` as Mumble currently reports this on shutdown, when ACRE was active:
```
[ERROR]: Plugin with ID 92 leaked memory from a call to API function "getAllChannels"
```

EDIT: I just verified that Mumble does in fact not allow deadlocks to occur by calling an API function. Thus presumably the deadlock was in fact created by an event attempting to schedule another event. Since the event loop then blocked itself, Mumble will lock as well the next time a callback is called that attempts to schedule an event.

Krzmbrzl:
I did some more work today and figured out that most crashes observed in the last testing round that I did was due to using an unstable Mumble build - no idea what corrupted it but using a freshly built version prevents most crashes.

By repeated loading and unloading the plugin, I was still able to crash Mumble though. Turned out this was due to an unhandled exception thrown from the plugin. This has now been fixed.

Additionally I found that disconnecting from a server while the plugin is connected will cause a new deadlock. However, I think I already know where this is coming from. Resolving it will require some input from the main devs though (happening on Slack), so I will only fix it once I have chatted with them.

extensions/src/ACRE2Mumble/MumbleCallbacks_static.cpp
```diff
@@ -29,3 +29,29 @@ mumble_version_t mumble_getAPIVersion() {
 mumble_version_t mumble_getVersion() {
     return mumble_version_t{ACRE_VERSION_MAJOR, ACRE_VERSION_MINOR, ACRE_VERSION_SUBMINOR};
 }
+
+struct MumbleStringWrapper mumble_getAuthor() {
+    static const char* author = "ACRE2 developers";
```
jonpas: Should return `ACRE_TEAM_URL` as in TeamSpeak.
https://github.com/IDI-Systems/acre2/blob/a0572a4dd4fe871b8b896e544300605f4cd0d87c/extensions/src/ACRE2TS/TSCallbacks_init.cpp#L43
or `ACRE_TEAM` if more fitting.

Macro source: https://github.com/IDI-Systems/acre2/blob/74b5b20fa028da206543e8100057019cae083f77/extensions/src/ACRE2Shared/_CONSTANTS.h

extensions/src/ACRE2Mumble/MumbleCallbacks_static.cpp
```diff
@@ -29,3 +29,29 @@ mumble_version_t mumble_getAPIVersion() {
...
+    MumbleStringWrapper wrapper;
+    wrapper.data = author;
+    wrapper.size = strlen(author);
+    wrapper.needsReleasing = false;
+
+    return wrapper;
+}
+
+struct MumbleStringWrapper mumble_getDescription() {
+    static const char* description = "ACRE2 is a full fledged communications suite for Arma 3, allowing realistic radio and voice communications.";
```
jonpas: Should return `ACRE_DESC` as in TeamSpeak.
https://github.com/IDI-Systems/acre2/blob/a0572a4dd4fe871b8b896e544300605f4cd0d87c/extensions/src/ACRE2TS/TSCallbacks_init.cpp#L46

jonpas (changes requested):

Krzmbrzl:
Okay so my latest test indicates that this seems to be working now. Thus, I would say that we could merge this into your `mumble-plugin` branch and then I can continue based on that.

Krzmbrzl:
@jonpas thanks for fixing that for me!
(You don't have something like a clang-format config that I overlooked, do you?)

jonpas (approved):

jonpas:
There is a [clang-format](https://github.com/IDI-Systems/acre2/blob/master/extensions/src/.clang-format), but I believe it was added after last updates to Mumble plugin branch. Also not all source comply to it yet, so don't worry about it too much.

PR #1114, opened 2021-05-28
PRC117F/PRC152 - Add missing hashes to `getChannelData` event

mjc4wilton:
- Title

Need to pull deviation in a mod I'm working on and I'm having to get hash data in a separate way specifically for it. This makes it easier as its just one event call.


jonpas (approved):
@Sniperhid @Soldia1138 Assuming this is good.

PR #1117, opened 2021-06-17
Documentation - Add Antennas and Gain Pattern Videos

jonpas:
**When merged this pull request will:**
- Move Class Names page from User to Frameworks section in sidebar (URL remains the same)
- Add Antenna class names
- Add Antennas list to Basic Concepts page
- Add Antenna Gain Pattern videos by @Drofseh 
- Rework Basic Concepts page structure/order

image

PR #1124, opened 2021-07-23
Antenna - Make 1.23M VHF TNC compatible with AN/PRC-152

jonpas:
- Make 1.23M VHF TNC compatible with AN/PRC-152

[`Harris_123cm_Whip.nec`](https://github.com/IDI-Systems/acre2/blob/29e20ea74051eb1274735c8d28a04fa55eaf2ba7/extras/antennas/Harris_123cm_Whip.nec#L1) states this antenna is for both 152 and 117F, so I believe this was an oversight (likely because this system isn't actually exposed to the user). This was found while @Drofseh was making videos for #1117.

### DIFF
--- a/addons/sys_antenna/CfgAcreComponents.hpp
+++ b/addons/sys_antenna/CfgAcreComponents.hpp
@@ -86,7 +86,7 @@ class CfgAcreComponents {
         connector = ACRE_CONNECTOR_TNC;
         height = 1.23;
         binaryGainFile = QPATHTOF(binary\Harris_123cm_Whip_gain.aba);
-        compatibleRadios[] = {"ACRE_PRC117F"};
+        compatibleRadios[] = {"ACRE_PRC152", "ACRE_PRC117F"};
     };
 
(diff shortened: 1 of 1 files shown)


PabstMirror (approved):

TheMagnetar (approved):

TheMagnetar:
Note that this does not change anything since the antennas are not swappable. We do not have any object representing an antenna in the  inventory. Therefore, what do you try to achieve in here? Do you want to add a variation of the PRC152 with this antenna and change the default component https://github.com/IDI-Systems/acre2/blob/master/addons/sys_prc152/CfgAcreRadios.hpp?

jonpas:
> Note that this does not change anything since the antennas are not swappable. We do not have any object representing an antenna in the inventory.

As said above. :P

> Therefore, what do you try to achieve in here?

Even if it's not used, it is a system that might be exposed to the user in the future. This caused a confusion when creating the gain pattern videos, thus I opened this PR. `.nec` comment and config were not aligned.

PR #1147, opened 2021-10-09
Core - Fix radio, intercom and sound effects not affected by global volume

### LINKED ISSUE #1080: `acre_api_fnc_setGlobalVolume` doesn't seem to impact radios
- **Arma 3:** `2.02`
- **CBA:** `3.15.3`
- **ACRE2:** `2.8.0.1040` [dev]

**Description:**
Passing various values to acre_api_fnc_setGlobalVolume doesn't seem to change radio volumes, direct speech works as expected.

**Steps to reproduce:**
- Dedi server with 2 people, 
- Call `[0.5] call acre_api_fnc_setGlobalVolume` locally on your client, direct speech volume is 1/2'd, radio traffic remains normal volume.

**Expected behavior:**
Should impact all volumes (direct + radio)

- Dedicated

**Log Files:**
- Link to ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- Link to ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the extension log file, which can be found at `<Arma 3 directory>\logs\acre_dll.log`.

**Additional context:**
I can see it referenced here https://github.com/IDI-Systems/acre2/blob/deb0a5dab7f7e687b1f4decf86d5de4a8319cd59/addons/sys_core/fnc_processRadioSpeaker.sqf#L93 but it doesn't seem to be taking effect. Raised due to the ACE unconcious effects impacting direct direct speech but not radios.

### LINKED ISSUE #6392: ACE Hearing - Unconscious Volume doesn't work for incoming incoming radio transmissions
**Arma 3 Version:** `1.82.144710` (stable)
**CBA Version:** `3.7.1.180604` (stable)
**ACE3 Version:** `3.12.2` (stable)

**Mods:**
```
- CBA_A3
- ace
- ACRE2 (or TFAR)
```

**Description:**
ACE Hearing: Unconscious Volume doesn't work for incoming incoming radio transmissions

**Steps to reproduce:**
- Set Unconscious Volume to 0
- Get shot
- Hear all the radio transmissions you want but not the person next to you talking in direct speech

- Dedicated

**Placed Modules:**
- None

jonpas:
- Fix #1080 
  - Remove use of `GVAR(globalVolume)` from `fnc_processRadioSpeaker` to avoid double-setting (don't think `_radioVolume` parameter in this function is even used, as for radios the radio volume property is used)
  - Fix https://github.com/acemod/ACE3/issues/6392 properly for radios and all other ACRE2 sounds as well
  - Fix radio speakers not affected by global volume
  - Fix intercom and infantry phone not affected by global volume
- Fix an unstringified config entry
- Fix sound effects not affected by global volume
- Fix sound effects not following dB dropoff
  - Streamline radio sounds to dB scale, ignoring God Mode

Could use another set of ears to make sure levels are about the same and not completely off. Radio speakers should also be a bit more manageable now and not 3x louder (more like 2x as intended).

_Was testing without ACE, then added ACE and completely forgot ACE calls this function every frame, thought I broke something as it kept resetting in front of my eyes..._

### DIFF
--- a/addons/sys_prc343/functions/fnc_onVolumeKnobPress.sqf
+++ b/addons/sys_prc343/functions/fnc_onVolumeKnobPress.sqf
@@ -32,7 +32,7 @@ private _currentVolume = GET_STATE("volume"); //["getState", "volume"] call GUI_
 private _newVolume = ((_currentVolume + _currentDirection) max 0) min 1;
 
 if (_currentVolume != _newVolume) then {
-    ["Acre_GenericClick", [0, 0, 0], [0, 0, 0], _newVolume^3, false] call EFUNC(sys_sounds,playSound);
+    ["Acre_GenericClick", [0, 0, 0], [0, 0, 0], _newVolume, false] call EFUNC(sys_sounds,playSound);
     ["setVolume", _newVolume] call GUI_DATA_EVENT;
 
 
--- a/addons/sys_core/CfgSounds.hpp
+++ b/addons/sys_core/CfgSounds.hpp
@@ -1,7 +1,7 @@
 class CfgSounds  {
     class acre_lip_sound {
         name = "";
-        sound[] = {QPATHTOF(sounds\mouth.wss), db-40, 1.0};
+        sound[] = {QPATHTOF(sounds\mouth.wss), "db-40", 1.0};
         titles[] = {};
     };
 };
--- a/addons/sys_core/fnc_processDirectSpeaker.sqf
+++ b/addons/sys_core/fnc_processDirectSpeaker.sqf
@@ -82,7 +82,7 @@ private _underwater = (ACRE_LISTENER_DIVE == 1) || {_emitterHeight < -0.2};
 
 if (_isIntercomAttenuate) then {
     _speakingType = "i";
-    _directVolume = _directVolume * [_unit] call EFUNC(sys_intercom,getVolumeIntercomUnit);
+    _directVolume = _directVolume * ([_unit] call EFUNC(sys_intercom,getVolumeIntercomUnit));
     _underwater = false;
 };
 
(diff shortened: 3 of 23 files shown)


jonpas:
Alright, this is ready for review, tested with @TyroneMF and everything is in order now and nicely adjusts.

addons/sys_prc343/functions/fnc_onVolumeKnobPress.sqf
```diff
@@ -32,7 +32,7 @@ private _currentVolume = GET_STATE("volume"); //["getState", "volume"] call GUI_
 private _newVolume = ((_currentVolume + _currentDirection) max 0) min 1;
 
 if (_currentVolume != _newVolume) then {
-    ["Acre_GenericClick", [0, 0, 0], [0, 0, 0], _newVolume^3, false] call EFUNC(sys_sounds,playSound);
+    ["Acre_GenericClick", [0, 0, 0], [0, 0, 0], _newVolume, false] call EFUNC(sys_sounds,playSound);
```
PabstMirror, line 35: Might be outside the scope of this PR but 
Do we actually want the click to be scaled by the new volume of the radio?

PabstMirror (approved):
looks ok

addons/sys_prc343/functions/fnc_onVolumeKnobPress.sqf
jonpas: Yes it is played by the radio, those clicks are generally routed to same output as receiving audio.

jonpas: I probably misunderstood the first time, but in addition, this is also a click played by the radio firmware, not necessarily the physical button press. I guess that can be discussed, 152 for example doesn't take volume into account for its volume buttons, but 343 and others do.

343 in addition does it for volume knob but not for channel knob. I guess this might be to easily get feedback of the currently set volume.

PR #1159, opened 2021-11-01
Radio - Allow blocking all radio transmissions on a unit

PabstMirror:
`acre_sys_core_isDisabled` - Blocks radios and ptt voice
`acre_sys_core_isDisabledRadio` - Will just block radios

### DIFF
--- a/addons/sys_radio/fnc_canUnitTransmit.sqf
+++ b/addons/sys_radio/fnc_canUnitTransmit.sqf
@@ -23,6 +23,9 @@ private _vehicle = vehicle acre_player;
 if (acre_player getVariable [QEGVAR(sys_core,isDisabled), false]) exitWith {
     false
 };
+if (acre_player getVariable [QEGVAR(sys_core,unableToUseRadio), false]) exitWith {
+    false
+};
 
 if (_vehicle != acre_player) then {
     if ((_radioId in ACRE_ACCESSIBLE_RACK_RADIOS || {_radioId in ACRE_HEARABLE_RACK_RADIOS}) && {[_radioId, acre_player] call EFUNC(sys_rack,isRadioHearable)}) then {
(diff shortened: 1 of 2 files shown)


jonpas:
`acre_sys_core_isDisabledRadio` would make more sense from the existing naming perspective.

Ideally, we should make API functions for this (out of scope of this PR probably), with 3 ways to disable ACRE: `0 - default, 1 - radio disabled, 2 - direct disabled, 3 - all disabled` or something like it.

PabstMirror:
`unableToUseRadio` was the same wording as a another mod
but I agree `isDisabledRadio` is probably more consistent with acre's naming

We could also overload the existing var and handle both bool and scalar `false/true/0/1/2/3`
but I'm not sure I like that

also, an entirely different way of fixing this problem would be to
```
if ace exists then 
   check ace's canInteractWith func
```
but I don't really like that either

addons/sys_radio/fnc_canUnitTransmit.sqf
```diff
@@ -23,6 +23,9 @@ private _vehicle = vehicle acre_player;
 if (acre_player getVariable [QEGVAR(sys_core,isDisabled), false]) exitWith {
     false
 };
+if (acre_player getVariable [QEGVAR(sys_core,unableToUseRadio), false]) exitWith {
```
PabstMirror:
```suggestion
if (acre_player getVariable [QEGVAR(sys_core,isDisabledRadio), false]) exitWith {
```

jonpas:
Yeah, I think separate variables internally is completely fine, makes debugging/testing easier too.

API functions that set those things could be considered in the future.

---

We should also reset it here probably https://github.com/IDI-Systems/acre2/blob/74b5b20fa028da206543e8100057019cae083f77/addons/sys_core/fnc_onPlayerKilled.sqf#L29.

jonpas (approved):

PR #1180, opened 2022-01-05
Extensions - Add effect for internal radio speakers

### LINKED ISSUE #1173: Improve Audio Quality of 148/152 Speaker Functionality - INT AUDIO & Radio Speakers
**Is your enhancement related to a problem?**
It's more of a question rather than a specific request/problem: is it possible to improve the audio distortion used for the [148/152 speaker functionality](https://wiki.clearbackblast.com/guides:acrespeakers)? It's a really neat/useful feature, however, it kinda sound a bit more like people around you plugged their headsets directly into your radio rather than a true speaker system. Especially in direct comparion to the same feature in TFAR. Would it be possible to update the audio distortion / sound model used for speakers to make then sound a bit more tinny and "speaker-like". I know that's not a good description of the enhancement but unfortunately I am not an audio-tech guy. :)

**Solution you'd like:**
Pretty much what I described above, if the speaker system could sound a bit more like TFARs speakers or other RL equivalents it would improve the experience a lot.

**Alternatives you've considered:**
Switching to TFAR, but why would I do that when I <3 ACRE2 so much?!

**Additional context:**
Nope, that's pretty much it. Thank you for your time and efforts!

PabstMirror:
Close #1173

Small speakers generally struggle with lower frequencies. e.g. frequency response of some phones:
Pixel3-comp_575px

This attempts to emulate that, and should make it more apparent to players that they are listening to a radio's speaker

### DIFF
--- a/extensions/src/ACRE2Core/FilterRadio.cpp
+++ b/extensions/src/ACRE2Core/FilterRadio.cpp
@@ -8,7 +8,7 @@
 
 #define BASE_DISTORTION_LEVEL 0.43f
 
-acre::Result CFilterRadio::process(short* samples, int sampleCount, int channels, acre::volume_t value, bool noise) {
+acre::Result CFilterRadio::process(short* samples, int sampleCount, int channels, acre::volume_t value, bool noise, const bool isLoudSpeaker) {
     float buffer[4096], *floatPointer[1];
     short *shortPointer[1];
 
@@ -43,6 +43,11 @@ acre::Result CFilterRadio::process(short* samples, int sampleCount, int channels
         // shelf it
         this->m_LowPass.process(sampleCount*channels, floatPointer);
         this->m_HighPass.process(sampleCount*channels, floatPointer);
+
+        // Drop low-end if on speakerphone
+        if (isLoudSpeaker) { 
+            this->m_lowShelf.process(sampleCount * channels, floatPointer);
+        }
 
         // Convert it back to shorts
         for (int i = 0; i < sampleCount; ++i) {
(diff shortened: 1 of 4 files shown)


extensions/src/ACRE2Core/FilterRadio.cpp
```diff
@@ -89,6 +94,7 @@ CFilterRadio::CFilterRadio(void)
     srand((unsigned int)time(0));
     this->m_HighPass.setup (TS_SAMPLE_RATE, 750, 0.97);
     this->m_LowPass.setup (TS_SAMPLE_RATE, 4000, 2.0);
+    this->m_lowShelf.setup(TS_SAMPLE_RATE, 1000, -20, 1);
```
PabstMirror: -20db matches the video, 
but I think it would be better to make the effect more subtle so we don't have to add a setting for this
probably 10-15

jonpas: I think -20db is fine honestly, but we can set it to -15db so it doesn't become "useless" on certain audio setups.

Drofseh: Can it be a setting? Not sure how that would work for the extension

jonpas: I don't see a need for a setting. This is more realistic but not that different.

PabstMirror:
```suggestion
    this->m_lowShelf.setup(TS_SAMPLE_RATE, 1000, -10, 1);
```
I think 10db will be enough to allow listeners to tell that it's a speaker

jonpas (approved):

PR #1187, opened 2022-02-11
Attenuate - Add config value for attenuation between people inside and turned out

Timi007:
- Add config value for attenution between people inside and turned out in the same vehicle
- Replace hardcoded `0.5` value with config value allowing modders to change it at will
- Keep `0.5 * acreAttenuation` as default
- Try to make code more readable

I'm still uncertain if `acreTurnedOutAttenuation` is a good name. If you have a better name, I'm open to change it.

### DIFF
--- a/addons/sys_attenuate/CfgSoundEffects.hpp
+++ b/addons/sys_attenuate/CfgSoundEffects.hpp
@@ -1,68 +1,68 @@
 /*
  * acreAttenuation: Use this to determine the attenution between people inside and outside vehicles.
- * acreTurnedOutAttenuation: Use this to determine the attenution between people inside (turned in) and turned out in a vehicle (refers to isTurnedOut command).
+ * acreAttenuationTurnedOut: Use this to determine the attenution between people inside (turned in) and turned out in a vehicle (refers to isTurnedOut command).
  */
 
 class CfgSoundEffects {
     class AttenuationsEffects {
         acreDefaultAttenuation = 0;
-        acreDefaultTurnedOutAttenuation = 0;
+        acreDefaultAttenuationTurnedOut = 0;
 
         class CarAttenuation {
             acreAttenuation = 0.5;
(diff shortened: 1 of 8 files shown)


Drofseh:
I might switch it to `acreAttenuationTurnedOut` and `acreDefaultAttenuationTurnedOut`

addons/sys_attenuate/CfgSoundEffects.hpp
```diff
@@ -1,49 +1,68 @@
-/* Use this to determine the attenution between people inside and outside vehicles. */
+/*
+ * acreAttenuation: Use this to determine the attenution between people inside and outside vehicles.
+ * acreAttenuationTurnedOut: Use this to determine the attenution between people inside (turned in) and turned out in a vehicle (refers to isTurnedOut command).
```
jonpas, line 3: If I understand correctly, this is when both players are in the vehicle, one turned in, one turned out. If either of them is outside the vehicle, `acreAttenuation` gets used instead? In that case, turned out is still treated like he was outside the vehicle?

Timi007: Yes, this is for when a player is turned in and one is turned out (this was previously hardcoded to `0.5 * acreAttenuation`). No, `acreAttenuation` is used for turned in players and outside players. If a player is turned out and one is outside, there is no `acreAttenuation` (`acreAttenuation == 0`).

jonpas: Sorry, yes that's what I meant. Good, just wanted to confirm logic didn't change for that.

jonpas (approved):
Looks good to me. @Timi007 has this been tested in a multiplayer environment?

Timi007:
No. I will test it sometime this or next week.

Timi007:
Tested it just now.

PR #1194, opened 2022-03-27
Attenuate - Add config value for attenuation turned out in different compartments

Timi007:
- Add config value for attenuation between players inside and turned out in different comparments
- Add documentation for all attenuation systems

Currently player 1 turned out in compartment A can hear player 2 turned in (inside) in compartment B with an attenuation value of  `acreAttenuationTurnedOut` (or before #1187 with a value of `acreAttenuation * 0.5`). This PR will allow the configuration of `acreAttenuationTurnedOut` between different compartments.

As default I propose the same values as for the attenuation between compartments for players turned in (inside). This will change the current default behavoir but makes more sense. 

Example case:
acre_attenuation

We have 2 compartments in an IFV. In Comp A sits the driver and gunner and in B the infantry.  
Currently the *red* attenuation can be configure in `CfgVehicles >> Tank >> ACRE >> attenuation`.  The default is that they cannot hear each other.
The *brown* attenuation is configured in `CfgSoundEffects >> TankAttenuation >> acreAttenuationTurnedOut` (Default `0.3`).   
The *yellow* attenuation is currently the same as the *brown* meaning that the driver turned out can hear the turned in infantry in the back although he cannot hear them if he himself is turned in because they're in different compartments.  
The purpose of this PR is to add a config value `CfgVehicles >> Tank >> ACRE >> attenuationTurnedOut` for the *yellow* attenuation.   

I'm open to change the default values for `CfgVehicles >> Tank >> ACRE >> attenuationTurnedOut` to be close to the current default behavior (`attenuationTurnedOut`).


Timi007:
This has already been tested in a multiplayer environment.

jonpas:
Damn, that image is awesome! :D

Would you consider expanding documentation on attenuation, especially this, with this or another drawing as a visual aid on how compartments function? Would be really neat.

> As default I propose the same values as for the attenuation between compartments for players turned in (inside). This will change the current default behavoir but makes more sense.

Agreed.

Timi007:
Sure!

So that everyone knows what we're talking about, some definitions:
- **Attenuation**: A value 0-1 defining how much the voice volume is lowered. 0 means the volume is not lowered an behaves the same as you would outside the vehicle. 1 means that the voice is muted and cannot be heard by others.
- **Compartment**: A virtual space where vehicle seats can be grouped together. A vehicle can have multiple compartments. BI already provides compartments. Each compartment is called `CompartmentX` where `X` is a number. It is up to the developer of the vehicle to use compartments.

First let's keep it simple and look at a vehicle with only one compartment. The picture represents all possible scenarios a player can talk to another one. A connection of the same color is the same scenario. For easier explanation, I will consider the vehicle an IFV or tank but the concepts are applicable to other types of vehicles too.
acre_attenuation_one_comp_names

Let's start with the dude outside the vehicle named Bob that talks to Tom *(violet)*. Because Tom is turned out, the attenuation is 0. If Bob talks to Kim or vice versa *(pink)* the attenuation value is `acreAttenuation` which is defined in the config `CfgSoundEffects >> AttenuationsEffects >> TankAttenuation >> acreAttenuation`. In our case this value is `0.6`.

Because everyone is in one compartment, there's no attenuation between Chad, Jim and Kim *(blue)*.

Tom and Hans are turned out. All players turned out have no attenuation between them *(green)*.
If they talk to someone inside e.g. Jim *(brown)* the attenuation value is `acreAttenuationTurnedOut` defined in `CfgSoundEffects >> AttenuationsEffects >> TankAttenuation >> acreAttenuationTurnedOut`, so `0.3`.

Now let's split the vehicle in two compartments. Hans and Chad will be in `Compartment1` and Jim, Kim, Tom will be in `Compartment2`. The scenarios for Bob are not changed, so we leave him out.
acre_attenuation_two_comp_names

Chad talks to Jim *(red)*. Because they are in different compartments the attenuation value is looked up in the config `CfgVehicles >> Tank >> ACRE >> attenuation >> Compartment1 >> Compartment2` (from the perspective of Jim) which is 1 meaning that Jim cannot hear Chad. 

If Hans however talks to Jim *(yellow)*, Jim can hear him with an attenuation value of `0.3` (`acreAttenuationTurnedOut`) because the current system does differentiate between compartments if one is turned out.

This PR adds the same red system for the yellow scenario. Meaning that if now Hans talks to Jim, Jim will not hear him because the config attenuation value is `CfgVehicles >> Tank >> ACRE >> attenuationTurnedOut >> Compartment1 >> Compartment2 == 1`.

Timi007:
Ups, ignore that I closed it. I miss clicked and accidentally commented and closed.

jonpas:
We should put that here! https://acre2.idi-systems.com/wiki/frameworks/vehicle-attenuation :)

Timi007:
> We should put that here! https://acre2.idi-systems.com/wiki/frameworks/vehicle-attenuation :)

In this PR or separate?

jonpas:
Either would be fine.

Timi007:
Added documentation for all attenuation systems in this PR

jonpas (approved):

PR #1205, opened 2022-05-24
Core - Filter radios for CBA Extended Loadouts

BrettMayson:
- Call `acre_api_fnc_filterUnitLoadout` on `CBA_loadoutGet`
- Requires https://github.com/CBATeam/CBA_A3/pull/1548

### DIFF
--- a/addons/sys_core/XEH_postInit.sqf
+++ b/addons/sys_core/XEH_postInit.sqf
@@ -3,6 +3,11 @@
 
 [QGVAR(onRevealUnit), { _this call FUNC(onRevealUnit) }] call CALLSTACK(CBA_fnc_addEventHandler);
 
+["CBA_loadoutGet", {
+    params ["", "_loadout", ""];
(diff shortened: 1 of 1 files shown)


BrettMayson:
Should be good to go now that the CBA bug fix is out

addons/sys_core/XEH_postInit.sqf
```diff
@@ -151,4 +151,9 @@ if (getClientStateNumber < 10) then { // Check before game has started (in brief
     }, 0, []] call CBA_fnc_addPerFrameHandler;
 };
 
+["CBA_loadoutGet", {
```
Vdauphin: Can you move this before the `!hasInterface` so when `CBA_fnc_getLoadout` is called server side, it catch the correct loadout?

jonpas: @BrettMayson ?

jonpas (approved):

PR #1213, opened 2022-07-15
Optionals - Add `meta.cpp`

### LINKED ISSUE #1209: Add meta.cpp files to optionals
**Is your enhancement related to a problem?**
Currently, the acre2 optional's published on the steam workshop are missing meta.cpp and mod.cpp. This means that the mods cannot be auto-downloaded through the Arma 3 launcher server browser.

**Solution you'd like:**
Update respective optional mods so they have at least the meta.cpp file.

https://community.bistudio.com/wiki/Arma_3:_meta.cpp

https://steamcommunity.com/sharedfiles/filedetails/?id=2450980898&searchtext=acre2+sog
EXAMPLE:
/// Version of meta.cpp content
protocol = 1;

/// Steam Workshop item id
publishedid = 2450980898;

/// Steam Workshop item name
name = "ACRE2 S.O.G. Prairie Fire Compatibility";

/// Time of last update of the Steam Workshop Item (in .NET Framework time format)
timestamp = 9876432101234567890;

/// Mod hash that is send by the server to the client instead of calculated hash)
hashOverride = 1234567890;

PabstMirror:
close #1209


PabstMirror:
@BrettMayson  - adding a meta.cpp with `publishedid = 2450980898;` makes hemtt angry
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow }', F:\DEV\HEMTT\target\release\build\armake2-ed023e89ff1c9404\out/config_grammar.rs:91:36
```
it only happens on id's greater than 2^31 (so it only effects more recent mods 😄 )

optionals/sys_csla/meta.cpp
```diff
@@ -0,0 +1,2 @@
+protocol = 1;
+publishedid = 2649217268;
```
PabstMirror, line 2:
```suggestion
publishedid = "2649217268";
```

optionals/sys_sog/meta.cpp
```diff
@@ -0,0 +1,2 @@
+protocol = 1;
+publishedid = 2450980898;
```
PabstMirror, line 2:
```suggestion
publishedid = "2450980898";
```

optionals/sys_ws/meta.cpp
```diff
@@ -0,0 +1,2 @@
+protocol = 1;
+publishedid = 2657862776;
```
PabstMirror, line 2:
```suggestion
publishedid = "2657862776";
```

optionals/sys_csla/meta.cpp
PabstMirror: not sure if quotes will work, because I think it's the a3 launcher reading these files?

Dahlgren:
> it only happens on id's greater than 2^31 (so it only effects more recent mods 😄 )

Issue exists in armake2 grammar logic, lib would need to be fixed and then updated in HEMTT released branch. This must be causing problems with any other config using floats/integers larger than 32 bit.

https://github.com/BrettMayson/armake2/blob/d164a525b7f43147af87f48cade2e52dafcd2255/src/grammars/config_grammar.rustpeg#L8-L21

I wonder why it even cares about the meta.cpp files 🤔

optionals/sys_csla/meta.cpp
Dahlgren, line 2:
```suggestion
publishedid = 2649217268;
name = "ACRE2 CSLA Iron Curtain Compatibility";
```

optionals/sys_gm/meta.cpp
```diff
@@ -0,0 +1,2 @@
+protocol = 1;
+publishedid = 1740990569;
```
Dahlgren, line 2:
```suggestion
publishedid = 1740990569;
name = "ACRE2 Global Mobilization Compatibility";
```

optionals/sys_sog/meta.cpp
Dahlgren, line 2:
```suggestion
publishedid = 2450980898;
name = "ACRE2 S.O.G. Prairie Fire Compatibility";
```

optionals/sys_ws/meta.cpp
Dahlgren, line 2:
```suggestion
publishedid = 2657862776;
name = "ACRE2 Western Sahara Compatibility";
```

Dahlgren (commented):
`name` is a supported property as well, https://community.bohemia.net/wiki/Arma_3:_meta.cpp

This  is used by some tools for a more human readable presentation

optionals/sys_csla/meta.cpp
jonpas: This should be able to be tested locally by just adding `meta.cpp` to already downloaded mod and opening Arma 3 launcher. It should display the Workshop item name instead of the `@mod` folder name.

jonpas:
> This is used by some tools for a more human readable presentation

Doesn't Arma 3 Launcher automatically get that name from the Workshop via the ID?

jonpas:
Shouldn't be a problem with HEMTT v1, will check after #1241.

jonpas:
Works with HEMTT v1, will merge after #1241.

jonpas (approved):

PR #1215, opened 2022-07-23
Extensions - Add Tracy support

Krzmbrzl:
- Add Tracy support in order to be able to investigate the performance issues encountered in the Mumble plugin. Note that Tracy needs to be activated via `-DUSE_TRACY=ON` when calling cmake
- Fix a small issue in the Mumble plugin where the returned config path would be inaccessible (due to lying in programs directory)
- Add CMake option to enable tracing (OFF by default)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "extensions/src/ACRE2Profiling/tracy"]
+	path = extensions/src/ACRE2Profiling/tracy
+	url = https://github.com/wolfpld/tracy.git
--- a/extensions/CMakeLists.txt
+++ b/extensions/CMakeLists.txt
@@ -58,6 +58,9 @@ if(MSVC)
     set(GLOBAL_SOURCES ${GLOBAL_RC})
 endif()
 
+
+add_subdirectory(src/ACRE2Profiling)
+
 include_directories(src/ACRE2Shared)
 include_directories(src/ACRE2Core)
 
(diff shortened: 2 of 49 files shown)


Krzmbrzl:
@jonpas I think this is now ready for review/merging.
Had I known in advance that I would keep tinkering with this for so long, I would have marked the PR as draft :eyes:

jonpas:
Self-note: Need to add that to docs and CI, as it's a new extension module.

jonpas:
> needs to be activated via `-DUSE_TRACY=ON` when calling cmake

I don't think that's correct, did you mean `-DENABLE_TRACING=ON`?

Also can you submodule Tracy on a specific tag instead of a hard-to-find-version commit?

---

As for CI, I don't think we need to add `ACRE2\ACRE2Profiling` in https://github.com/IDI-Systems/acre2/blob/eed369203df942e1afc7b8e637c0ae43b8ce6ebd/.github/workflows/release.yml#L97 as it should get automatically pulled in by `ACRE2Core` like `ACRE2Core` itself gets pulled into other modules.

I do want to add profiling options to the docs though, if/when Mumble ever gets merged - or maybe cherry-picked out into the main branch at least.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "extensions/src/ACRE2Profiling/tracy"]
```
PabstMirror, line 1: instead of submodules maybe just use cmake fetch

```
include(FetchContent)
FetchContent_Declare(
    tracy	
    GIT_REPOSITORY https://github.com/wolfpld/tracy
    GIT_TAG v0.9.1
)
FetchContent_MakeAvailable(tracy)
```

PR #1218, opened 2022-08-28
Extensions - Add Linux support for the Mumble plugin

DylanFrese:
* Make some syntactic changes for other compilers

* Update CMakelists for Linux compatibility
  
  This allows ACRE2Core, ACRE2Shared, and ACRE2Wine to be built under Linux toolchains.

* Extend compat.h to provide Linux compat types
  
  Some Windows types and functions have been aliased for compilation on Linux.

* Provide FAudio/x3daudio compatibility on Linux
  
  FilterPosition relies on x3daudio to do positional transformations. On Linux, we can use FAudio as an (almost) drop-in replacement.
  
* Alias tbb concurrency modules on Linux
  
  Intel TBB is available as a library on Linux, under the `tbb` namespace, rather than the `concurrency` namespace as on Windows.
  
* Update filesystem paths for building on Linux

  Use paths in ~ for ACRE config & temp files

* Do not load AcreSettings in constructor
  
  If the default path baked into the constructor is not the same as one that is later loaded, the user would end up with two config files!

* Reimplement NamedPipeServer on Linux
  
  This is the most involved change in this changeset.
  
  On Linux, Arma runs under WINE, but Mumble is native. There's no (known to me) way to do IPC between WINE processes and native processes.  

  This solution uses a simple duplex network socket listening on localhost, with a default port of 19141. The Mumble side is responsible for setting up the server, and the Arma side connects.

* Modify arma2ts to use sockets if running in Wine
  
  When running in Wine, ACRE2Arma.dll will mirror the changes to NamedPipeServer.cpp for Linux; instead of using named pipes, it will connect to the socket opened on the other side.

* Add a Linux build job to CI
  
  Also, publish the resulting Linux/Mumble plugin

This PR adds support for building the ACRE Mumble plugin for native Linux, and modifies both it and the arma2ts DLL (when running in WINE) to communicate over network sockets instead of named pipes.

I've polished the commit log so that this changeset may be reviewed in a developmental manner on the command line with
`git log -w -p --reverse origin/mumble-plugin..HEAD`, rather than as one big diff (if you prefer).

I've also added and tested CI support (although I haven't tested the publishing steps, for obvious reasons) for the Linux plugin. CI has built [this workshop item](https://steamcommunity.com/sharedfiles/filedetails/?id=2855182350), if you'd like to test it.


DylanFrese:
This is currently marked as WIP because **ACRE2Steam** currently does not search for the Linux version of Mumble to copy plugins to. Wine, by default, maps `Z:` to `/`, so automatically adding/updating the plugin is certainly doable, I just have to get my head around what `voip_plugin` is doing.

The Linux plugin will also build as 32-bit, but there's no 32-bit image of Ubuntu, and Ubuntu 22 doesn't appear to have all the multiarch packages necessary.

You can view a successful CI run here: https://github.com/DylanFrese/acre2/actions/runs/2942274717

The *scratch* commit in that CI run has modified `release.yml` to compile on a non-release branch, and also trimmed out things like publishing and signing; it is not included in this PR.

rautamiekka:
> Ubuntu 22 doesn't appear to have all the multiarch packages necessary.

As in, doesn't exist at all ? I seem to recall reading some being missing.

DylanFrese:
> > Ubuntu 22 doesn't appear to have all the multiarch packages necessary.
> 
> As in, doesn't exist at all ? I seem to recall reading some being missing.

I'm not 100% sure, I don't have a Ubuntu system myself to check on. I know Ubuntu 19 dropped 32-bit support to an extent. I'd went through the steps to enable multiarch in CI, and modify `sources.list` to grab i386 packages, but `apt` reported they weren't present.

I'm almost certain, though, that anyone running Arma under Wine/Proton is using 64-bit.

rautamiekka:
> > As in, doesn't exist at all ? I seem to recall reading some being missing.
> 
> modify `sources.list` to grab i386 packages, but `apt` reported they weren't present.
> 
> I'm almost certain, though, that anyone running Arma under Wine/Proton is using 64-bit.

Ya, x86_32 support ended after *Ubuntu 18.04 LTS "Bionic Beaver" (supported until April 2023) except for a very rare few derivatives, so those packages straight up don't exist.

^ According to [this 2021 comment](https://askubuntu.com/questions/1356712/running-32-bit-programs-on-21-04#comment2322843_1356712), likely way too many 32 multiarch libs just don't exist cuz ppl don't wanna do the work, and you likely are right about 64 'Arma' under Wine/Proton mostly, so indeed it's far better to make a 64 version 1st and try to make a 32 one if it ain't too much trouble, otherwise only a 64.

DylanFrese:
Again, to be clear, the 32-bit build works. I've built it on my system, running Debian. If there are any 32-bit users out there (which is more rare in the Linux desktop space than in the Windows space), they'd be able to build it themselves (which they're also more likely to know how to do than in the Windows space).

jonpas:
I don't see a reason to support 32-bit Linux.

extensions/src/ACRE2Arma/arma2ts/CallExt_DllMain.cpp
```diff
@@ -9,23 +9,34 @@
...
+#pragma comment(lib, "Ws2_32.lib")
 
 #define PIPE_COMMAND_OPEN    0
 #define PIPE_COMMAND_CLOSE    1
 #define PIPE_COMMAND_WRITE    2
 #define PIPE_COMMAND_READ    3
 #define PIPE_COMMAND_RESET    4
+#define COMMAND_IS_WINE 99
+
+#define WINE_SOCKET_PORT "19141"
```
DylanFrese, line 32: Arbitrary; not on the https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports

extensions/src/ACRE2Core/Engine.cpp
```diff
@@ -29,21 +29,34 @@
...
+#ifdef WIN32
         std::array<char, MAX_PATH> appDataPath{""};
         if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, 0, appDataPath.data()))) {
             acrePluginLog = std::string(appDataPath.data()) + "\\Arma 3\\" + acrePluginLog;
         }
+#else
+        if (getenv("XDG_DATA_HOME")) {
+            acrePluginLog = std::string(getenv("XDG_DATA_HOME")) + "/" + acrePluginLog;
+        } else {
+            acrePluginLog = std::string(getenv("HOME")) + "/.local/share/" + acrePluginLog;
```
DylanFrese, line 52: Can crash if `$HOME` is unset, but that should never happen, and the user probably has bigger problems if that happens.

extensions/src/ACRE2Core/ping.h
```diff
@@ -10,8 +10,14 @@
...
 RPC_FUNCTION(ping) {
+#ifdef WIN32
     g_pingTime = clock() / CLOCKS_PER_SEC;
-    vServer->sendMessage(CTextMessage::formatNewMessage("pong", "%f,", g_pingTime));
+#else
+    timespec t;
+    clock_gettime(CLOCK_MONOTONIC_RAW, &t);
+    g_pingTime = t.tv_sec;
+#endif
+    vServer->sendMessage(CTextMessage::formatNewMessage("pong", "%f,", (float) g_pingTime));
```
DylanFrese, line 20: `MONOTIME` is more 'correct' for computing time differences, but does not track walltime. I don't think it actually matters here.

DylanFrese:
I've pushed code that now copies over the Linux .so to the user's Mumble Plugins directory when running under Wine. I depends on the environment variable `WINEHOMEDIR` being set -- the home directory of the Linux user as mapped in Wine. This isn't documented anywhere, but Wine sets it and has for a long time (concluded after doing some git archaeology); there isn't really any other way to get this information. It also isn't guaranteed that the user's home directory is even mapped inside of Wine, but it almost always will be.

As for the rest of the PR, I've taken the approach of trying to alter as little behaviour as possible when it comes to the Windows builds of the extensions, mostly because I don't use Windows and am not really able to verify the correctness of any changes.

I don't have any substantial experience with C++, and so what I've written I've tried to blend in with surrounding code, but I'm sure a lot of it isn't idiomatic. I welcome a harsh review to improve it to a higher C++ standard; I simply don't know when there are better approaches available in the language or libraries.

@jonpas This is ready for review.

DylanFrese:
The Ubuntu runner builds against libtbb2, which is kind of old for newer distributions. It's not available in Debian Bookworm or Manjaro. I need to update the Ubuntu image or otherwise try to link against libtbb12

PR #1221, opened 2022-10-01
Inventory - Show channel/frequency information for radios

mharis001:
- Like ACE interaction actions, show radio channel/frequency information in the inventory UI item lists

### DIFF
--- a/addons/sys_gui/fnc_initDisplayInventory.sqf
+++ b/addons/sys_gui/fnc_initDisplayInventory.sqf
@@ -46,3 +46,53 @@ _ctrlContainerItems ctrlAddEventHandler ["MouseButtonUp", {
     // Make sure any moved ItemRadio is actually fully moved before resetting hold off
     [FUNC(inventoryListMouseUp), _this, 1] call CBA_fnc_waitAndExecute;
 }];
+
+// Show channel/frequency information for radios
+private _fnc_update = {
+    params ["_display"];
+
+    private _pttAssign = [] call EFUNC(api,getMultiPushToTalkAssignment);
+
+    {
+        private _ctrl = _display displayCtrl _x;
+
+        for "_i" from 0 to (lbSize _ctrl - 1) do {
+            private _item = _ctrl lbData _i;
+
+            if (_item call EFUNC(api,isRadio)) then {
+                private _name = _item call EFUNC(api,getDisplayName);
+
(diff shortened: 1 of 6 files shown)


addons/sys_gui/fnc_initDisplayInventory.sqf
```diff
@@ -46,3 +46,53 @@ _ctrlContainerItems ctrlAddEventHandler ["MouseButtonUp", {
...
+                // Display frequency for single-channel radios (e.g. AN/PRC-77)
+                private _maxChannels = [_item, "getState", "channels"] call EFUNC(sys_data,dataEvent);
+
+                private _text = if (isNil "_maxChannels") then {
+                    private _txData = [_item, "getCurrentChannelData"] call EFUNC(sys_data,dataEvent);
+                    private _currentFreq = HASH_GET(_txData, "frequencyTX");
+                    format ["%1 %2 MHz", _name, _currentFreq];
+                } else {
+                    format [LELSTRING(ace_interact,channelShort), _name, _item call EFUNC(api,getRadioChannel)]
+                };
```
jonpas: We could put that in a core function as both ACE Interact and this uses the exact same logic.

addons/sys_gui/fnc_initDisplayInventory.sqf
```diff
@@ -46,3 +46,53 @@ _ctrlContainerItems ctrlAddEventHandler ["MouseButtonUp", {
...
+
+                // Display radio keys in front of bound radios
+                // Case-insentive since PTT API function returns radio class in lower case while
+                // inventory display provides it in config case
+                private _radioKey = (_pttAssign findIf {_x == _item}) + 1;
+
+                if (_radioKey > 0 && {_radioKey < 4}) then {
+                    _text = format ["%1: %2", _radioKey, _text];
+                };
```
jonpas: Same for this, could take `_radioKey` as an argument.

jonpas (approved):
Looks good and can be merged despite the reviews (can be done at a later time).

BrettMayson:
This method could replace what I was trying in https://github.com/CBATeam/CBA_A3/pull/1329, if this works better than that. I couldn't get that PR to work on corpses

mharis001:
Added a function to `sys_core` to remove the duplicated code. Also, use this now for the ACE Arsenal name changing.

PR #1226, opened 2022-11-27
BF-888S - Add Arsenal Stats

TACHarsis:
- Add arsenal stat entries for min/max frequency, power and range to BF888S config

Only value I am absolute certain is power. Frequency I might have misunderstood and the range is from a quick google search.

### DIFF
--- a/addons/sys_bf888s/CfgWeapons.hpp
+++ b/addons/sys_bf888s/CfgWeapons.hpp
@@ -20,6 +20,11 @@ class CfgWeapons {
         class Library {
             libTextDesc = QUOTE(NAME_BF888S);
         };
+
+        EGVAR(arsenalStats,frequencyMin) = 400; // internet says 400-470MHz
+        EGVAR(arsenalStats,frequencyMax) = 400.62; // code implies 63 channels 0.01 apart?
+        EGVAR(arsenalStats,transmitPower) = 5000;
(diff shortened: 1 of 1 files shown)


jonpas:
_Answer for question on Discord to keep it easily referencable._

Range is just an approximation anyways, our Wiki lists 4-6 km perfect, 2-4km city, so what you put down is good. If you want to make it the same as other radios, then I suggest `"4-6km (2-4km)"`.

Frequency range of a BF-888S is 400-470 MHz (UHF band), You can just look up the spec sheet of the real radio _(mind, there are 2 versions IIRC, 3W and 5W)_. Presets do not cover the full capability of the radio. You can also reference the antenna, but in this case the NA-701 simulation model covers a wider range than the radio does.

TACHarsis:
Frequencies then are 400-470Mhz.
As for range I just assumed the best and used the highest estimate, since it's using 5 watts ingame, even though some reviews say that the actual sending power is closer to 2 watts. Considering that it is set to 50 times (5000) the sending power of the 343 (100), even with a weaker antenna, that should probably be close to accurate. 
Didn't have time to actually test distances precisely, but we used it in a wooded area today and it seemed to check out.
If someone has more precise experimental values or the power gets adjusted I'll lower the upper estimate.

jonpas:
> 5 watts ingame, even though some reviews say that the actual sending power is closer to 2 watts.

I think there are 2 versions of the radio depending on local law. They do the same with their other models.

jonpas (approved):

PR #1234, opened 2023-03-10
Data - Fix data sync causing hitching with large data due to logging

### LINKED ISSUE #1233: [sys_data] Data sync logging is causing hitching with large dataset
**Description:**
The act of logging this diagnostic information is causing hitching:
```
2023/03/05, 22:21:31 [ACRE] (sys_data) INFO: Data Sync for acre_player: PLAYER_NAME [2:680563] - 566.91KB
```
The code that handles this is here:
https://github.com/IDI-Systems/acre2/blob/master/addons/sys_data/fnc_serverHandleJip.sqf#L23
https://github.com/IDI-Systems/acre2/blob/master/addons/sys_data/fnc_clientHandleJipData.sqf

The act of calculating the size is horribly expensive if there is a lot of data to sync, likely more expensive than the act of actually syncing the data. 

Recommended fix is to setup a boolean that can be set on the fly `ACRE_SYS_DATA_LOG_SYNC_SIZE` which when false will skip the expensive `(count (toArray (str _jipData)))/1024` calculation and just log the other data.

JonBons:
- Adds a `ACRE_DEBUG_DATA_SYNC` variable that can be set to 1 to re-enable this kind of log
- When disabled we still log on sync but we skip calculating the dataset size
- Fixes https://github.com/IDI-Systems/acre2/issues/1233


PabstMirror (approved):

jonpas:
Awesome, good find and fix!

jonpas (approved):

PR #1240, opened 2023-04-11
Extension - Handle stereo audio from TeamSpeak (cherry pick)

### LINKED ISSUE #1237: Audio artifacts in TS channel with high(est) bitrate
Game and mods (from workshop) are latest versions at the moment. No ACE.
Teamspeak server version: 3.13.7 on Linux
client: 3.5.6
If ACRE's channel Audio codec is set to Opus Music with Quaility 10 for some clients sound is pitched down + crackling.
My input and output devices are all set to 24bit 48khz. I did not had any problems, but other client with 16 bit 44.1khz "Speakers" heard my voice with mentioned artifacts. No problems with Opus Voice Quality 10

I may add recorded example later.

PabstMirror:
fix #1237
Cherypick the stero fix from #1105
> acre just expects mono microphone data, but TS is capable of stereo
without the mono fix, everything was pitch shifted lower because the sound data is interleaved [left, right, left]

### DIFF
--- a/extensions/src/ACRE2Core/FilterPosition.cpp
+++ b/extensions/src/ACRE2Core/FilterPosition.cpp
@@ -8,6 +8,7 @@
 
 #include <cmath>
 #pragma comment(lib, "x3daudio.lib")
+#pragma comment(lib, "xaudio2.lib")
 
 #define MAX_FALLOFF_DISTANCE    75
 #define MAX_FALLOFF_RANGE        150
--- a/extensions/src/ACRE2Core/SoundEngine.cpp
+++ b/extensions/src/ACRE2Core/SoundEngine.cpp
@@ -37,7 +37,7 @@ acre::Result CSoundEngine::onEditPlaybackVoiceDataEvent(acre::id_t id, short* sa
             for (size_t i = 0; i < player->channels.size(); ++i) {
                 if (player->channels[i]) {
                     player->channels[i]->lock();
-                    player->channels[i]->In(samples, sampleCount);
+                    player->channels[i]->In(samples, sampleCount, channels);
                     player->channels[i]->unlock();
                 }
             }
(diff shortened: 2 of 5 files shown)


extensions/src/ACRE2Core/FilterPosition.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include <cmath>
 #pragma comment(lib, "x3daudio.lib")
+#pragma comment(lib, "xaudio2.lib")
```
PabstMirror, line 11: not fully sure why this is now required, it's unrelated to pr but need to build on VS17.6

jonpas: Interesting.

jonpas: Didn't need that on VS17.5. :thinking:

PabstMirror: error
```
2>ACRE2Core.lib(FilterPosition.obj) : error LNK2019: unresolved external symbol __imp_X3DAudioInitialize referenced in function "public: enum acre::Result __cdecl CFilterPosition::process(short *,int,int,unsigned int,class CSoundMixdownEffect *)" (?process@CFilterPosition@@QEAA?AW4Result@acre@@PEAFHHIPEAVCSoundMixdownEffect@@@Z)
2>ACRE2Core.lib(FilterPosition.obj) : error LNK2019: unresolved external symbol __imp_X3DAudioCalculate referenced in function "public: enum acre::Result __cdecl CFilterPosition::process(short *,int,int,unsigned int,class CSoundMixdownEffect *)" (?process@CFilterPosition@@QEAA?AW4Result@acre@@PEAFHHIPEAVCSoundMixdownEffect@@@Z)
```

it might just be something on my machine
we could just remove the line from the pr and let it build via github automation
or fix via cmake config, but we already had a existing `#pragma comment(lib`

jonpas: It also builds with it for me, forgot to say that. So I am fine with leaving it in.

jonpas (approved):

PR #1241, opened 2023-04-13
Tools - Update HEMTT to v1

jonpas:
- Update tooling to use HEMTT v1

TODO:
- [x] Fix all the remaining config issues (mostly quoting)
- [x] Figure out what `Dlg_DIALOGHGT` is supposed to be
  - Is undefined (probably since move to Arma 3) and resulted in `0.0` in config, so replacing with that value.
- (Maybe some other time) ~~Set versions in docs, `README.md` and extension before building release~~
  - _Possibly change the way we do this a bit, so docs and `README.md` gets committed - extension and game can be set at 0.0.0_
- [x] Get full version in CI
- [x] Test everything
- [x] Update Tools package and documentation


jonpas:
**Requires HEMTT v1.4.0 or higher.**

PR #1245, opened 2023-05-14
I/O - Don't show pipe disconnected warning in Eden Preview

Mike-MF:
- Stops showing "ACRE is not connected to teamspeak" when in the editor.

### DIFF
--- a/addons/sys_io/fnc_server.sqf
+++ b/addons/sys_io/fnc_server.sqf
@@ -27,7 +27,7 @@ DFUNC(connectionFnc) = {
             // acre_player sideChat format["RESULT: %1", GVAR(pipeCode)];
             if (GVAR(pipeCode) != "1") then {
                 if (time > 15) then {
-                    if (isMultiplayer) then {
+                    if (isMultiplayer && {!is3DENPreview}) then {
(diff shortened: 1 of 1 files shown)


addons/sys_io/fnc_server.sqf
```diff
@@ -27,7 +27,7 @@ DFUNC(connectionFnc) = {
             // acre_player sideChat format["RESULT: %1", GVAR(pipeCode)];
             if (GVAR(pipeCode) != "1") then {
                 if (time > 15) then {
-                    if (isMultiplayer) then {
+                    if (isMultiplayer && {!is3DENPreview}) then {
```
jonpas:
```suggestion
#ifndef DEBUG_MODE_FULL
                    if (!is3DENPreview) then {
#endif
```
Wrap the closing bracket of this section the same way as well.

I removed the `isMultiplayer` check here as it's already checked before starting the function anyways. Remove the other `isMultiplayer` on L48 as well please.

jonpas (approved):

addons/sys_io/fnc_server.sqf
BrettMayson: Is there a reason not to indent the macros?

PR #1249, opened 2023-06-03
Inventory - Use per frame handler to set descriptive radio names

mharis001:
- title, as reported by some users the names could flicker on the inventory screen
- Could not reproduce but my best guess is that this change would eliminate any potential issues with the radio names not being set every frame


jonpas:
Can confirm this is fixed. I could easily reproduce the original issue with a local dedicated server as well.

jonpas (approved):

PabstMirror:
Still getting flicker 
https://clips.twitch.tv/MagnificentDoubtfulPenguinMoreCowbell-vsUNRScr0utJV2cL (@ 17 sec)
but I can't reproduce in local hosted mp

Cyruz143:
I've had reports of this too, no useful evidence like Pabst though. Dedi MP.

PR #1253, opened 2023-06-21
Compat SOG - Improve Racks and Attenuation configs

veteran29:
- title


optionals/sys_sog/subConfig/sys_sog_vehicles/config.cpp
```diff
@@ -0,0 +1,81 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        addonRootClass = QUOTE(COMPONENT);
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"acre_main", "loadorder_f_vietnam"};
```
jonpas: Why not just add this to the main config?

veteran29: I recalled that somebody somewhere complained that the compat required something more than just ACRE addons as they wanted it always loaded, but I guess it does not matter with `skipWhenMissingDependencies`, will move it.

jonpas: I think when skipping is in stable, we'll add it to all compats and put them in the base anyways.

Or maybe not due to size? Something to decide on.

jonpas:
@veteran29 I merged this and and made some changes to conform to how ACE3 is doing this. Let me know when this is ready.

veteran29:
If you're fine with the vehicles config being in a subcomponent it's good to go.

It could probably use M41, T-54, PT76 configs for attenuation but I do not have time.

veteran29:
Not sure about this:
https://github.com/IDI-Systems/acre2/pull/1253/files#diff-caaa1043f65e21860ae4a01035dbb2b3af41926c284d39a78e996add32dc3727R18

jonpas:
> If you're fine with the vehicles config being in a subcomponent it's good to go.
> 
> It could probably use M41, T-54, PT76 configs for attenuation but I do not have time.

What's the reason it's in a subconfig instead of just normal config?

> Not sure about this: [#1253 (files)](https://github.com/IDI-Systems/acre2/pull/1253/files#diff-caaa1043f65e21860ae4a01035dbb2b3af41926c284d39a78e996add32dc3727R18)

`vn_car_attenuation` is the same as `CarAttenuation`. `vn_tank_attenuation` is only slightly different from `TankAttenuation`. Just use ACRE's existing attenuation classes? If we want specific, then that one being in the compat addon is fine.

veteran29:
> What's the reason it's in a subconfig instead of just normal config?

No good reason since `skipWhenMissingDependencies`.

> just use ACRE's existing attenuation classes?

I've seen you have dedicated classes for CUP and RHS so have done it the same way:
https://github.com/IDI-Systems/acre2/blob/master/addons/sys_attenuate/CfgSoundEffects.hpp#L55-L66

jonpas:
> No good reason since `skipWhenMissingDependencies`.

Let's put it into main config then.

> I've seen you have dedicated classes for CUP and RHS so have done it the same way: [`master`/addons/sys_attenuate/CfgSoundEffects.hpp#L55-L66](https://github.com/IDI-Systems/acre2/blob/master/addons/sys_attenuate/CfgSoundEffects.hpp?rgh-link-date=2023-09-06T14%3A53%3A33Z#L55-L66)

Ah, I see. That is from before ACRE2 was even on GitHub. Use the default ones if they fit.

jonpas:
You weren't even using the new attenuation types anywhere.

I merged the subconfig up now. Let me know if this is all.

jonpas (approved):

Issue #854, opened 2019-10-17
Compatibility with Contact DLC Radio Backpacks


TheNightstalk3r:
Currently, the Contact DLC Radio backpacks are not compatible with the LRRs ACRE2.0 offers (either because of a setting or their tiny inventory space - I blame BI). It would be a great addition to the game to have these antenna-rich backpacks function with ACRE for those who want radio backpacks in the camouflage within the standard ArmAverse and its DLCs.

BrettMayson:
```hpp
class CfgVehicles {
	class Bag_Base;
	class B_RadioBag_01_base_F: Bag_Base {
		maximumLoad = 150; // 120 for 117F
	};
};
```
I've been using this to allow for a 117F and a small amount of extra gear

jonpas:
We probably shouldn't change maximum load in ACRE as that is a global change. What os the default value anyways? Sounds like those backpacks should have much more space to begin with.

Amd we can't change radio weights as that will change carrying capacity for everything else. 

I think best course of action here os separate mod that increases maximum load of those backpacks as Brett provided.

BrettMayson:
An optional perhaps? Not sure what the best course would be for the steam workshop however

BrettMayson:
A more interesting approach would be to create a new item `Integrated 117F`. It would have 0 mass. We could add it to all radio backpacks. This would get around the "All radios are items" issue without rewriting anything existing. The only issue would be the less than ideal method to prevent it from being moved out of the backpack. Still would be an interesting experiment.

TheNightstalk3r:
Would there not be an option to make this Integrated 117F only compatible with those backpacks? It's not ideal for mod support, but it's something corresponding mod makers could do something with.

jonpas:
`Integrated 117F` just sounds like a start of a great confusion to me. You should be able to put in a normal 117F.

I think changed mass in a separate mod is the best option. I am not sure that's something we would want in ACRE2, we'll discuss it internally.

jonpas:
If anyone wants to add backpacks that inherit from Contact DLC ones but have more space to allow a radio to fit, we can put them into ACRE2 as optional and also distribute it that way. However, we have no plans of doing this ourselves and distribution might be easier/faster if done on your own, so I am closing this request.

PR #1261, opened 2023-07-17
Add new Optional to fit AN/PRC-117F into Contact Radiopacks

mrschick:
As explained in #854, this PR implements [this config](https://github.com/IDI-Systems/acre2/issues/854#issuecomment-543489897) in a new Optional, overriding the max load of Radiopacks added by the Contact DLC and allowing them to carry an AN/PRC-117F with room to spare.

### DIFF
--- a/optionals/sys_radpacks/$PBOPREFIX$
+++ b/optionals/sys_radpacks/$PBOPREFIX$
@@ -1,0 +1,1 @@
+idi\acre\addons\sys_radpacks
--- a/optionals/sys_radpacks/README.md
+++ b/optionals/sys_radpacks/README.md
@@ -1,0 +1,4 @@
+sys_radpacks
+===
+
+Compatibility with Contact DLC radio packs, increasing their max load so that they can carry an AN/PRC-117F and a small amount of additional gear.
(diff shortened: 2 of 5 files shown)


jonpas:
[`setMaxLoad`](https://community.bistudio.com/wiki/setMaxLoad) exists now, which allows you to do it all in a mission or some framework someone is possibly using. This is ultimately easier to implement and use as needed.

```sqf
// initPlayerLocal.sqf
["loadout", {
    params ["_unit"];
    if (backpack _unit == "B_RadioBag_01_base_F") then {
        ["tag_setMaxLoad", [backpackContainer _unit, 150]] call CBA_fnc_serverEvent;
    };
}, true] call CBA_fnc_addPlayerEventHandler;

// initServer.sqf
["tag_setMaxLoad", {(_this select 0) setMaxLoad (_this select 1)}] call CBA_fnc_addEventHandler;
```

If we still want a solution in ACRE, we should implement it as an API function, but I really don't see the point when it's that simple and would possibly want to be further modified for a specific use case.

mrschick:
Thanks for the info, I didn't know about `setMaxLoad` and presumed that packaging the Config override detailed by Brett as its own mod and running it on every client was the best available way to increase radiopack volume.
I agree that the short script is more user friendly than an Optional would be, no real need for an in-ACRE solution 👍🏻

PR #1265, opened 2023-07-27
Racks/Intercom - Enable without ACE Interaction Menu (requires Intercom to access racked radios)

veteran29:
- title

Reasoning:
ACRE intercom menu allows you to use the radios in tanks without ACE Interact menu, you can just set the "Work" to the radio you want to use and open the configuration menu via keybinds.

Vehicles without the ~~rack~~intercom would need some other way (scroll menu?) but that could be done in separate PR IMHO.

### DIFF
--- a/.hemtt/project.toml
+++ b/.hemtt/project.toml
@@ -1,4 +1,5 @@
 name = "ACRE2"
+mainprefix = "idi"
 prefix = "acre"
 author = "ACRE2 Team"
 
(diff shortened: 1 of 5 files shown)


Drofseh:
> Vehicles without the rack would need some other way

Do you mean 'vehicles without the intercom...' ?

veteran29:
Yes

jonpas (approved):

PR #1270, opened 2023-08-07
Move compatibilities to main mod and conditionally load

Mike-MF:
- Moves CSLA, GM, SOG & WS compats into main.
- Utilises `skipWhenMissingDependencies`
- Uses `QPATHTOF` for WRP file.


addons/compat_csla/config.cpp
```diff
@@ -11,11 +11,12 @@ class CfgPatches {
         authors[] = {"CSLA Studio"};
         url = ECSTRING(main,URL);
         VERSION_CONFIG;
+        skipWhenMissingDependencies = 1;
```
PabstMirror: I don't think we even need these
can't think of a case where someone will have the map_compat but not the base acre

Mike-MF: Ah, you got me just before I updated requiredAddons.

PabstMirror: we don't really need the mod in requiredAddons
we could just leave it like before and the compats just always load and it shouldn't hurt anything

PabstMirror:
I think we want to drop the meta.cpps?

addons/compat_csla/config.cpp
jonpas: Some compats do more than just add an entry in `CfgAcreWorlds`, like SOG's radio replacement (don't want to run that function if SOG isn't even loaded) and the newly coming SPE, as well as improved compat for SOG currently in PR.

jonpas (approved):
Renaming shouldn't break anything unless someone required our compatibilities, which I highly doubt. If anyone knows of any, let us know.

addons/compat_csla/config.cpp
PabstMirror: my concern is that we have to wait for 2.14 to merge this PR
not sure if we want to wait for that to get SPE compat

we could use a `has_include` setup to get these merged now

jonpas: Nah, we can wait a few weeks. Previous compats have come much later.

jonpas (approved):

PR #1289, opened 2023-09-23
Tools - Fix configs for HEMTT v1.9.0

PabstMirror:
fixes all warnings on hemtt1.9.0rc2
also makes buildable in pboProject


jonpas:
See https://github.com/CBATeam/CBA_A3/pull/1609#pullrequestreview-1640937601

jonpas (approved):

PR #1300, opened 2023-11-04
Cleanup spaces in config macros

PabstMirror:


addons/ace_interact/ace_arsenal_stats.hpp
```diff
@@ -2,7 +2,7 @@
     class statBase;
     class GVAR(frequencyRange): statBase {
         scope = 2;
-        stats[] = {QEGVAR(arsenalStats,frequencyMin), QEGVAR(arsenalStats,frequencyMax)};
+        stats[] = {QEGVAR(arsenalStats,frequencyMin),QEGVAR(arsenalStats,frequencyMax)};
```
jonpas: This space was not in a macro.

BrettMayson:
You can have spaces in the define, those are fine.

You just can't have them in the call, that's where they get kept

jonpas (approved):

jonpas:
@PabstMirror Do you have a shell script or something you did this with? It needs to be applied to more files with newer HEMTT (radio states, menus, etc.).

PabstMirror:
https://github.com/IDI-Systems/acre2/pull/1304 should get it building, macro space should just be a warning

jonpas:
#1317 already made it build, I didn't see that PR of yours. I'd like to cleanup those spaces as well though.

PR #1305, opened 2023-11-29
Zeus - Fix remote control breaking Racks and resetting custom PTTs

### LINKED ISSUE #1267: Zeus - Exiting remote control while the AI character is inside a vehicle breaks racks and intercom

mrschick:
- Fix Racks being broken for the Zeus Player when remote controlling AI in vehicles, which resolves #1267:
It was caused by the `sys_rack/fnc_rackPFH.sqf` PFH being started when entering that AI and not terminated when exiting it, resulting in the PFH constantly disconnecting the player from his Racks. My fix works by limiting Rack initialization and PFH launch to only player units entering a vehicle, effectively preventing use of Racks by RCd AI entirely.
- Fix custom PTT assignments being reset when using a remote controlled unit's radios:
A minor bug I encountered some time ago, happened because there was only a single `DVAR(ACRE_ASSIGNED_PTT_RADIOS)` that stored custom PTT assignments, when controlling a unit and using its radios `sys_core/fnc_handleMultiPttKeyPress.sqf` would check for the custom PTT radios to be present, not find them on the AI and reset the list. Now there are 2 DVARs to remember custom PTT assignments on the player and last controlled AI unit.


Mike-MF (approved):
Works

addons/sys_core/XEH_preInit.sqf
```diff
@@ -66,6 +66,7 @@ DVAR(ACRE_LISTENER_DIVE) = 0;
 
 DVAR(ACRE_PTT_RELEASE_DELAY) = 0.2;
 DVAR(ACRE_ASSIGNED_PTT_RADIOS) = [];
+DVAR(ACRE_ASSIGNED_PTT_RADIOS_REMOTE) = [];
```
jonpas:
```suggestion
DVAR(ACRE_ASSIGNED_PTT_RADIOS_REMOTE) = []; // Remote controlled unit
```

jonpas (commented):
How does this work when controlling a UAV, as that is also remote control?

> My fix works by limiting Rack initialization and PFH launch to only player units entering a vehicle, effectively preventing use of Racks by RCd AI entirely.

It's more of a workaround but it does prevent an issue, so I am fine with accepting this change, but we should add a note in documentation.

jonpas:
@BrettMayson Since you authored all of the Zeus compatibility, mind taking a look at this?

mrschick:
> How does this work when controlling a UAV, as that is also remote control?

I did not think of testing that. Unless it also changes `acre_player` from `player`, it should not be affected.
Should get around to testing it next week.

> It's more of a workaround but it does prevent an issue, so I am fine with accepting this change, but we should add a note in documentation.

Maybe we can figure out an actual fix. Didn't look too much into it due to not seeing a use case in which a RCd AI should have to use Racks.

mrschick:
> How does this work when controlling a UAV, as that is also remote control?

Turns out `addons/sys_rack/fnc_enterVehicle.sqf` is not called when controlling a drone, only when entering a vehicle, so I can't find any cases in which it would create issues.
Unless a UAV has Racks, though it doesn't seem that is supported ATM.

mrschick:
Haven't looked too deep into it, but given the lack of apparent use-cases I don't think implementing Racks being accessible by RCd AI is worth the time.

BrettMayson (approved):

jonpas (approved):

PR #1320, opened 2024-04-01
Main - Replace Fast Hash system with native hashmaps

JonBons:
- Replaces all fast hash based macros with native HashMap methods
- Removes the fast hash monitor and fast hash garbage collector
- Removes ACRE_FastHashNamespaceDummy location type as it's unnecessary now


jonpas (approved):

addons/main/CfgLocationTypes.hpp
```diff
@@ -1,13 +0,0 @@
-class CfgLocationTypes {
-    // For use in Fast Hashes ONLY!
-    class ACRE_FastHashNamespaceDummy {
```
PabstMirror, line 3: we might want to leave this in for one release if it could effect saved games

PabstMirror:
need to check capitalization on all hash keys
e.g. "channelMode" vs "channelmode" on 117/152

["ACRE_PRC117F_ID_1"] call acre_sys_prc117f_fnc_getChannelDataInternal
now returns nil

["ACRE_PRC152", "default"] call acre_api_fnc_getPresetData;
returned location now hashmap

["ACRE_PRC152_ID_1"] call acre_api_fnc_getRadioChannel; // item's config case
returns -1
["acre_prc152_id_1"] call acre_api_fnc_getRadioChannel;
returns expected 6

BrettMayson:
> ["ACRE_PRC152", "default"] call acre_api_fnc_getPresetData;
> returned location now hashmap

This is quite a problem for other mods is it not? There is for example no documented way to get the channel labels except via this function, followed up with `getVariable`, this is going to break as it now will need hashmap functions

veteran29:
Yep, code like [this](https://github.com/ArmaForces/Mods/blob/master/addons/acre/functions/fnc_getRadioDisplayName.sqf#L29) will break... But it never was a public API so I guess fair enough.

BrettMayson:
I mean the example I posted is very much public api

JonBons:
I think there is enough complications from this with mod compat and general API use that it should probably be reverted until a better solution is designed or you're willing to break compat.

I don't use any public mods that interface with ACRE2 so I was unaware people were directly accessing this data and not utilizing some form of API call that could have adjustments made to maintain compat at the tradeoff of slightly worse perf per call (toLowerANSI on keys, etc)

jonpas:
We will revert if necessary (#1322 does not make it).

PR #1325, opened 2024-04-17
PRC152 - Previous/Next Channel Stepping via PRE+/- instead of Knob

mrschick:
- Modify the Previous/Next Channel Keybind logic for the PRC-152, to simulate manipulation of the PRE+/- buttons instead of turning of the knob.
This allows stepping beyond channel 5, which I previously found pretty limiting in many situations. Realistically IMO, one could and would use the buttons blindly without having to stop and pull out the radio.

### DIFF
--- a/addons/sys_core/fnc_switchChannelFast.sqf
+++ b/addons/sys_core/fnc_switchChannelFast.sqf
@@ -35,7 +35,20 @@ if (_isManpack == 0 || {_isRackRadio}) then {
             _channel = ((_channel + _dir) max (_currentBlock*16 + 1)) min ((_currentBlock + 1)*16);
         };
         case "ACRE_PRC152": {
-            _channel = (_channel + _dir) min 5;
+            if (_dir == 1) then {
+                if (_channel < 99) then {
+                    _channel = _channel + 1;
+                } else {
+                    _channel = 0;
+                };
+            } else {
+                if (_channel > 1) then {
+                    _channel = _channel - 1;
+                } else {
+                    // Go to the last preset
+                    _channel = 99;
+                };
+            };
         };
         default {
             _channel = _channel + _dir;
--- a/addons/sys_core/fnc_switchChannelFast.sqf
+++ b/addons/sys_core/fnc_switchChannelFast.sqf
@@ -35,20 +35,7 @@ if (_isManpack == 0 || {_isRackRadio}) then {
             _channel = ((_channel + _dir) max (_currentBlock*16 + 1)) min ((_currentBlock + 1)*16);
         };
         case "ACRE_PRC152": {
-            if (_dir == 1) then {
-                if (_channel < 99) then {
-                    _channel = _channel + 1;
-                } else {
-                    _channel = 0;
-                };
-            } else {
-                if (_channel > 1) then {
-                    _channel = _channel - 1;
-                } else {
-                    // Go to the last preset
-                    _channel = 99;
-                };
-            };
+            _channel = (_channel + _dir + 98) % 99 + 1;
         };
         default {
             _channel = _channel + _dir;


addons/sys_core/fnc_switchChannelFast.sqf
```diff
@@ -35,7 +35,20 @@ if (_isManpack == 0 || {_isRackRadio}) then {
...
+                    _channel = 0;
+                };
+            } else {
+                if (_channel > 1) then {
+                    _channel = _channel - 1;
+                } else {
+                    // Go to the last preset
+                    _channel = 99;
+                };
+            };
```
BrettMayson:
```suggestion
            _channel = (_channel + _dir + 98) % 99 + 1;
```

mrschick: Damn, that's a neat solution 👍🏻

mrschick: Unfortunately it doesn't work when stepping backward over Ch1->Ch99. It just remains on Ch1, because:
`(1 + (-1) + 100) % 100 = 100 % 100 = 0 (expected 99)`

BrettMayson: Oh right, I was thinking there was a channel 0.
updated suggestion

mrschick: Not working on Ch99->Ch1, because:
`(99 + (+1) + 100) % 100 = 200 % 100 = 0`, making the `if` kick in.

BrettMayson: What are you trying? The suggestion was updated, it is now `(_channel + _dir + 98) % 99 + 1` which works for me at [99,+1] and [1,-1]

mrschick: I must have hallucinated seeing the following code in the suggestion.
```sqf
_channel = (_channel + _dir + 100) % 100;
if (_channel == 0) then {
    _channel = 99;
};
```

mrschick: `(_channel + _dir + 98) % 99 + 1` works 👍🏻

BrettMayson: It was there for a short bit, but was replaced not too long after and well before your comment, so I wanted to make sure that's what you were trying

mrschick: Awesome, I kept debugging the first edit until I wrote my comment 😆

jonpas:
Previous/Next Channel keybind is supposed to imitate turning the knob, as you would do that quickly without pulling the radio out.

You need to pull the radio out to manipulate the PRE+/- buttons on the physical radio. I am not sure how you would use those buttons while the radio is inside a pouch of some kind.

BrettMayson:
If there was a delay of some kind, like an animation, I'd say make it take longer when going past 5, as if taking time to feel out the buttons.

Could have this as an option? Personally I don't ever go beyond 5 though, so I don't care either way

mrschick:
There have been several instances, mainly as a pilot, where I only carried a PRC-152 while dismounted and had to switch between channel 5-6-7 to keep contact with ATC and TACP.
I just find the current restriction overly limiting in those situations, even realistically speaking.

> You need to pull the radio out to manipulate the PRE+/- buttons on the physical radio. I am not sure how you would use those buttons while the radio is inside a pouch of some kind.

IRL you could still pull the PRC-152 halfway out of the pouch and press the PRE+/- buttons on the move, without having to stop, pull out the radio and hold in front of your face. That's the main appeal of the keybind after all, it works on the move.
Also, it might be in a normal pocket as a backup for pilots (or covert SOF operatives), in a front-facing pouch that doesn't block the buttons, or a claw pouch:
claw-pouch

> If there was a delay of some kind, like an animation, I'd say make it take longer when going past 5, as if taking time to feel out the buttons.

IMO, that would just be a needlessly convoluted abstraction of tactile input.
I also need to feel for my "change channel" keybind or avert my attention from the screen to find it immediately.

jonpas (approved):

PR #1326, opened 2024-04-25
IO - Fix non-ASCII characters breaking channel switching

jokoho48:
- Add Sanitization of serverName in sqf to prevent ts3ChannelMove to fail

### DIFF
--- a/addons/sys_io/fnc_ts3ChannelMove.sqf
+++ b/addons/sys_io/fnc_ts3ChannelMove.sqf
@@ -17,7 +17,7 @@
 
 private _serverName = toString ((toArray serverName) select {_x < 127});
 if (_serverName == "") then {
-	_serverName = "ServerName Fully Sanitized away";
+    _serverName = "ServerName Fully Sanitized away";
 };
 private _ts3ChannelDetails = format ["%1,%2,%3", EGVAR(sys_core,ts3ChannelName), EGVAR(sys_core,ts3ChannelPassword), _serverName];
 TRACE_1("Moving TS3 Channel",_ts3ChannelDetails);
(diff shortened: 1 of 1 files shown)


jonpas:
Does this fix #579?

jokoho48:
> Does this fix #579?

Yes, it should fix that issue.

addons/sys_io/fnc_ts3ChannelMove.sqf
```diff
@@ -15,6 +15,10 @@
  * Public: No
  */
 
-private _ts3ChannelDetails = format ["%1,%2,%3", EGVAR(sys_core,ts3ChannelName), EGVAR(sys_core,ts3ChannelPassword), serverName];
+private _serverName = toString ((toArray serverName) select {_x < 127});
+if (_serverName == "") then {
+    _serverName = "ServerName Fully Sanitized away";
```
jonpas: Can we add a warning for RPT for that as well?

jokoho48: done

addons/sys_io/fnc_ts3ChannelMove.sqf
```diff
@@ -15,6 +15,11 @@
...
+    WARNING_1("Server name '%1' did not include any ASCII Characters and got fully Sanaitzed.", serverName)
```
jonpas:
```suggestion
    _serverName = "Unsupported Server Name";
    WARNING_1("Server name '%1' did not include any valid (ASCII) characters and got fully sanitized!",serverName)
```

jonpas (approved):

PR #1334, opened 2024-07-25
Rack - API and Usability Improvements

### LINKED ISSUE #1311: Text goes out of interface
All steam version
- **Arma 3:** `2.14 stable`
- **CBA:** `3.16.1 stable`
- **ACRE2:** `2.12.0 stable`

**Description:**
In some languages the text for vehicles racks and intercom goes out of the predefined space for the interface (screenshot for example)

**Steps to reproduce:**
- Set the Arma language to, in this case, `Italian`

**Screenshots:**
WhatsApp Image 2023-12-26 at 22 24 12
WhatsApp Image 2023-12-26 at 22 24 11

### LINKED ISSUE #1312: API manipulation of vehicle racks on dedicated servers (removeRackfromVehicle not functioning as expected)
**Mods:**
- **Arma 3:** `2.14.150957`
- **CBA:** `CBA_A3 v3.16.1.231025`
- **ACRE2:** `2.12.0.1056`
- **Unsung Redux Stable** `Update: 27 Sep, 2023 @ 7:32pm`
**Description:**

A CBA "InitPost" ClassEventHandler is added in init.sqf to a parent class of vehicle (uns_plane). 

The config for the test vehicle is such that two rack mounted radios are added.

The intention is for this event handler to remove the vehicle racks on vehicles of this type and replace them with other vehicle racks, as the vehicles in question, when placed in Eden Editor or via Zeus and entered by a player, have a type of rack and mounted radio that is not desired (117F instead of 77).

The event handler calls a CBA function using RemoteExec to (per ACRE2 functions documentation) run only on the server, and as the event handler has been added fires on all connected clients, it also only calls RemoteExec if the event is firing on the server itself. An alternative would be to implement the event handler on the server only via initserver.sqf and avoid the need for remote execution of a function.

```
["uns_plane", "InitPost", {
params ["_vehicle"];
if (isServer) then {	
_vehicle remoteExec ["TACD_adjustRadioRacks", 2];
};
}, nil, nil, true] call CBA_fnc_addClassEventHandler;
```
As no player has entered the vehicle at the point at which the event handler fires, it first initialises the vehicle racks and waits until this is done. Then it attempts to delete the racks which have been initialised and add the new ones, forcing initialisation of the new racks via the flag in the addRackToVehicle function.

```
_target = _this;
[_target, {}] call acre_api_fnc_initVehicleRacks;
waitUntil { _target call acre_api_fnc_areVehicleRacksInitialized };
_racks = _target call acre_api_fnc_getVehicleRacks;
{[_target, _x] call acre_api_fnc_removeRackFromVehicle} forEach _racks;
[_target, ["ACRE_VRC64", "Panel 1", "Dash", false, ["crew"], [], "ACRE_PRC77", [], []], true] call acre_api_fnc_addRackToVehicle;
```
When this mission is run "multiplayer but locally" (ie, using the 'play in multiplayer'  feature in Eden editor), the effect is that only one rack is removed, and one is added. This is not as expected, since the construction of forEach _racks would expect that both racks found in _racks would be removed. If the code is adjusted by adding a copy of row 5 (ie, just calling the rack remove function again), it works correctly.

When this mission is run on a dedicated server, the effect is not as desired. The original two racks are not removed at all, but the new rack is added.

If the code is executed manually via the console after a player has joined the dedicated server, rather than via an event handler, then it works as per a local server (with the issue above - only a single rack is removed). 

I cannot find in the documentation references to this. Is it necessary to also waitUntil acre_api_fnc_isInitialized returns true on the first connecting player?

mrschick:
- Fix object variable `acre_sys_rack_initialized` not being a reliable indicator for whether the Racks are fully initialized and added to the vehicle. This seems to have been the main cause behind past Rack-API jankiness.\
Will close #1312.
- Add API function `acre_api_fnc_removeAllRacksFromVehicle`:\
It removes all Racks from a Vehicle after handling the preceding initialization (if necessary).
- Add API function `acre_api_fnc_replaceRacksOnVehicle`:\
It replaces all Config-defined Racks on a vehicle of a specific type with a given other type, while preserving other config properties like Name/Shortname, allowed/denied positions and Wired Intercom.\
Use case: for an old-school scenario, put just this function into your APC/Helicopters' Init fields in 3DEN to replace all of its default VRC-103 Racks with VRC-64s. Eliminates the need to initialize, delete all Racks and add new VRC-64s with a custom script.
- Add documentation for new API functions.
- Improve German and Italian translations of short "Dash", to reduce overflow in the rack/intercom info dialog.
- Fix text wrapping of the rack/intercom info dialog in order to prevent content overflows.\
Will close #1311.
- Improve handling of a player's connection to Racks when exiting and re-entering a vehicle:
  - Don't reset the custom PTT order of Rack Radios;
  - Automatically Re-"Use" Racks;
- Fix Rack-mounted AN/PRC-77 interaction showing "Channel: 1" instead of frequency in MHz.


mrschick:
Preserving PTTs when re-entering vehicles works, so does re-using Racks, but the latter doesn't properly handle seat switching yet.

mrschick:
Managed to fix the vehicle info text overflowing, was caused by `addons/sys_gui/RscTitles.hpp:53` multiplying Rsc width by 10, more than what would have been visible anyway, which prevented the native structured text wrapping from working properly.
I also reduced the divisor at `addons/sys_gui/fnc_updateVehicleInfo.sqf:33` a bit to get some extra lower padding.

mrschick:
> the latter doesn't properly handle seat switching yet

I thought a bit more about how the rack-reusing feature should work in relation to how seat switching influences Rack usage.
Currently, a used Rack remains used when switching seat, the RackPFH just stops using a given Rack if it is inaccessible from the (new) current seat.

My re-use implementation works as follows:
When starting to use a Rack, that state is stored in a local variable for the current seat. When entering the vehicle again, the variable for the selected seat is read and any racks in that array are re-used.
If the seat is switched within the vehicle, stopping usage of the Rack won't update its stored state for the seat it was added for, only for the one currently occupied.

In the end, I think that this implementation is ok, with storage per-seat being the best option, avoiding situations in which stopping usage from any seat would remove its stored state from every other seat.

jonpas (approved):

PR #1336, opened 2024-08-12
GUI - Allow Volume Control during ACE Progressbar Dialog

mrschick:
- Dynamically add the `"MouseZChanged"` `displayEventHandler` to an opened ACE Progressbar, to allow Volume Control while performing an Interaction.

This is the simplest way I figured out to implement this.
I guess a cleaner and slightly more efficient implementation would be an XEH that fires upon creation of the progressbar dialog, but that appears to require implementing it into ACE first.

### DIFF
--- a/addons/sys_gui/fnc_volumeKeyDown.sqf
+++ b/addons/sys_gui/fnc_volumeKeyDown.sqf
@@ -20,11 +20,8 @@ if (!alive acre_player || ACRE_IS_SPECTATOR || GVAR(volumeOpen)) exitWith {false
 // Abort on open Dialogs that aren't an ACE Progressbar
 private _aceProgressBar = displayNull;
 if (
-    dialog && 
-    {
-        !EGVAR(sys_core,aceLoaded) || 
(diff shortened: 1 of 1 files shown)


addons/sys_gui/fnc_volumeKeyDown.sqf
```diff
@@ -15,7 +15,23 @@
...
  */
 
-if (!alive acre_player || dialog || ACRE_IS_SPECTATOR || GVAR(volumeOpen)) exitWith {false};
+if (!alive acre_player || ACRE_IS_SPECTATOR || GVAR(volumeOpen)) exitWith {false};
+
+// Abort on open Dialogs that aren't an ACE Progressbar
+private _aceProgressBar = displayNull;
+if (
+    dialog && 
+    {
```
rautamiekka: The next check is a boolean, so lazy eval can't be useful here.

rautamiekka (changes requested):

addons/sys_gui/fnc_volumeKeyDown.sqf
mrschick: True that

jonpas (approved):

PR #1353, opened 2024-10-10
PRC117F/PRC152 - Fix undefined macro

PabstMirror:
this fixes being able to use other cba keybinds while the radio ui is open


PabstMirror:
this also fixes `Variable should not be all caps: ACRE_DEBUG_NAMESPACE` warnings

jonpas (approved):

PR #1357, opened 2024-11-30
Core - Fix Fast Channel Switch of VRC-64

mrschick:
- Fix the fast channel switch keybind for rack mounted PRC-77s. Now it will have the same behaviour as turning the MHz knob in the UI, when stepping below/above the range, it will jump to the last/first valid frequency respectively.

Previously, due to `sys_prc77/radio/fnc_getCurrentChannel.sqf` always returning channel index `0`, the fast channel switch keybind would only switch between channels of index `-1` (breaks the radio UI) and `+1` (31MHz).

### DIFF
--- a/addons/api/fnc_setRadioChannel.sqf
+++ b/addons/api/fnc_setRadioChannel.sqf
@@ -24,10 +24,10 @@ params [
 if !(_radioId isEqualType "") exitWith { false };
 
 if (_channelNumber isEqualType 0) then {
-    private _eventData = if (([_radioId] call FUNC(getBaseRadio)) == "ACRE_PRC77") then {
+    private _eventData = _channelNumber - 1;
+    if (([_radioId] call FUNC(getBaseRadio)) == "ACRE_PRC77") then {
         [_channelNumber, 0] // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
-    } else {
-        _channelNumber - 1
+    }
     };
     [_radioId, "setCurrentChannel", _eventData] call EFUNC(sys_data,dataEvent);
 } else {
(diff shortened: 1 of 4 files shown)


addons/api/fnc_setRadioChannel.sqf
```diff
@@ -24,9 +24,10 @@ params [
 if !(_radioId isEqualType "") exitWith { false };
 
 if (_channelNumber isEqualType 0) then {
-    private _eventData = _channelNumber - 1;
-    if (([_radioId] call FUNC(getBaseRadio)) == "ACRE_PRC77") then {
-        _eventData = [_eventData, 0]; // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
+    private _eventData = if (([_radioId] call FUNC(getBaseRadio)) == "ACRE_PRC77") then {
+        [_channelNumber, 0] // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
+    } else {
+        _channelNumber - 1
```
jonpas:
```suggestion
    private _eventData = _channelNumber - 1;
    if (([_radioId] call FUNC(getBaseRadio)) == "ACRE_PRC77") then {
        [_channelNumber, 0] // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
    }
```
Would prefer to keep it this way.

addons/api/fnc_setRadioChannel.sqf
```diff
@@ -26,7 +26,8 @@ if !(_radioId isEqualType "") exitWith { false };
 if (_channelNumber isEqualType 0) then {
     private _eventData = _channelNumber - 1;
     if (([_radioId] call FUNC(getBaseRadio)) == "ACRE_PRC77") then {
-        _eventData = [_eventData, 0]; // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
+        [_channelNumber, 0] // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
```
jonpas:
```suggestion
        _eventData = [_channelNumber, 0] // ACRE_PRC77 expects an array of [mhzKnob, khzKnob]
```

jonpas (approved):

PR #1372, opened 2025-01-28
Core - Use `fnc_getDescriptiveName` for all ACE Action Names

mrschick:
- Make all ACE Interaction Names use `sys_core_fnc_getDescriptiveName` in favour of `format ["%1 Chn: %2", _radioName, api_fnc_getRadioChannel]`.
I noticed yesterday that the GSA action to connect a PRC-77 displayed the radio's frequency as `[MHzKnobPos, KHzKnobPos]`, since `api_fnc_getRadioChannel` returns that raw data in its case. `getDescriptiveName`'s return is IMO better in all situations, as it will properly format channel data for all radios and also display useful info like the PTT number.
- Improve the PRC-343 description to show `Bk x Ch y` instead of the absolute (overflowing) channel number.
- Improve the SEM52SL description (and `switchChannelFast` hint) to show H and P channels as letters instead of numbers.
- Improve the SEM70 (and SEM90 Rack) description to show `x MHz` in HW mode and `Chn x NetID y` in AKW mode.
- Always pad out the decimal positions of frequency-based radios, i.e: `30.00 MHz` for the PRC-77 and `30.000 MHz` for the SEM 70 (in HW mode).

### DIFF
--- a/addons/ace_interact/stringtable.xml
+++ b/addons/ace_interact/stringtable.xml
@@ -38,6 +38,11 @@
             <German>%1 Bk %2 Kn %3</German>
             <Italian>%1 Bc %2 Cn %3</Italian>
         </Key>
+        <Key ID="STR_ACRE_ace_interact_channelNetIDShort">
+            <English>Chn %1 NetID %2</English>
+            <German>SP %1 FK-Nr %2</German>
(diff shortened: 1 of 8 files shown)


jonpas:
Agreed!

jonpas (approved):

PR #1374, opened 2025-02-04
Translation - Fix "Team" swapped between Japanese and Italian

Apricot-ale:
- Fix "Team" was swapped between Japanese and Italian
- Do not make pasta sushi


jonpas (approved):
Haha.

PR #1377, opened 2025-02-05
Extensions - Fix acre2.ini Configuration Path

Apricot-ale:
- Edit CTS3Client::getConfigFilePath to Fix config path not loaded, error causing by double slash.

Example of Error:
```
Configuration Path: {C:\Users\%USERNAME%\AppData\Roaming\TS3Client\\acre\acre2.ini}
Failed to load ACRE ini file. Using defaults...
```
ts3Functions.getConfigPath(`tempFolder`) is maybe return `AppData\Roaming\TS3Client\`
`\\acre2` is return `\acre2`
so `tempFolder += "\\acre";` is returns `AppData\Roaming\TS3Client\\acre`

I'm not familiar with C, so please correct me if this is a problem.
Also, I haven't tried this fix yet because I don't know how to build and test the plugin...

### DIFF
--- a/extensions/src/ACRE2TS/TS3Client.cpp
+++ b/extensions/src/ACRE2TS/TS3Client.cpp
@@ -456,7 +456,7 @@ std::string CTS3Client::getTempFilePath( void ) {
     char tempPath[MAX_PATH - 14];
     GetTempPathA(sizeof(tempPath), tempPath);
     std::string tempFolder = std::string(tempPath);
-    tempFolder += "acre";
+    tempFolder += "//acre";
     if (!PathFileExistsA(tempFolder.c_str())) {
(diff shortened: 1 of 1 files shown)


jonpas:
Double slashes shouldn't be a problem.

Apricot-ale:
> Double slashes shouldn't be a problem.

Apricot-ale:
> Double slashes shouldn't be a problem.

Yes, this isn't a big problem, and the necessary data is included by the fallback.
However, an error is logged every time, and I'm sure acre2.ini cannot be load.

Well, since acre2.ini is no longer needed by anyone, and this PR is mainly to eliminate the error, we might as just remove the  entirely acre2.ini things instead.

jonpas:
Removing it would be ideal. I am not sure if its really not being used by something though.

PR #1380, opened 2025-02-06
General - Add icons to ACE actions

Timi007:
- Add icons to ACE actions to radios
- Add icons to intercom
- Add icons to GSA
- Cleanup code (use `LLSTRING`)

Example:  
image

## TODO

- [x] Add icons to self-interaction
- [x] Add icons to intercom
- [x] Add icons to shared radios
- [x] Add icons for intercom broadcast
- [x] Add icons for intercom accent


addons/ace_interact/README.md
```diff
@@ -0,0 +1,16 @@
+### Attribution
+
+Some icons where taken (and modified) from Font Awesome Free 6.7.2 by https://fontawesome.com.
```
rautamiekka:
```suggestion
Some icons were taken (and modified) from Font Awesome Free 6.7.2 by https://fontawesome.com.
```

rautamiekka (changes requested):

Drofseh:
I'm not sure icons with letters should be used (L, R, PTT, etc) as they won't be translated if the user selects a non-English language,

Timi007:
> I'm not sure icons with letters should be used (L, R, PTT, etc) as they won't be translated if the user selects a non-English language,

I also am not sure how to depict those actions. I thought that L and R would be ok because most headphone makers also label it as such. In addition to that I added the sound direction. Would arrows be better or (double) chevrons? What about center, should it be left and right arrows?

Is a microphone with a number better suited for PTT?

Drofseh:
> > I'm not sure icons with letters should be used (L, R, PTT, etc) as they won't be translated if the user selects a non-English language,
> 
> I also am not sure how to depict those actions. I thought that L and R would be ok because most headphone makers also label it as such. In addition to that I added the sound direction. Would arrows be better or (double) chevrons? What about center, should it be left and right arrows?
> 
> Is a microphone with a number better suited for PTT?

I think probably just the sound wave icons, `(((`, `)))`, `((( )))`, are fine for L, R, and C
You could do waves and an arrow `((( <-` but I'm not sure if it'd be better or not.

Mic icon by itself (no number) is probably fine for the PTT icons. The number is presumably always in the text right under the icon.

Timi007:
I think this is ready for review.

jonpas:
Are we happy with these wave icons? I personally have no clue which is Left and Right from the icon itself, but I also don't think it really matters.

If we want them clearer, maybe some ear on the side or something?

jonpas (approved):
If we want to change icons in the future, we can, I am merging this as it is an improvement for sure.

Issue #1329, opened 2024-06-24
Option to turn off default distortion filter on radio audio


LuckyOwl1364:
**Is your enhancement related to a problem?**
The baseline distortion filter applied to radio communication makes it difficult-to-impossible for me to understand what is being said; even with the speaker ~10m away from me with no obstructions, the audio has hisses & pops & crackles applied to it that distort speech.

**Solution you'd like:**
If the filter could be turned off, so the only distortion to speech comes from obstructions, that would make operations go from frustrating & exhausting to perfectly fine.

**Alternatives you've considered:**
I do not think there are alternatives that will work for me: it is not a volume problem, but the hiss and the pops and the crackles applied to all audio played through radios that makes speech unparsable.

**Additional context:**
Example recordings of the audio (first clip the speaker is ~10m away, second clip the speaker is <100m away)

https://github.com/IDI-Systems/acre2/assets/24234179/a92a3876-a1f8-44c6-a1b5-419b203af192

https://github.com/IDI-Systems/acre2/assets/24234179/412f5222-c058-4e09-855d-dce14c40efc5

LuckyOwl1364:
Oh interesting: in the file `FilterRadio.cpp` is a check on a getter `getDisableRadioFilter()` that skips the audio-filtering process:
https://github.com/IDI-Systems/acre2/blob/e5582cd7716f620ed38d719ff283b2cf45f8d3f5/extensions/src/ACRE2Core/FilterRadio.cpp#L15
that is set to `false` by default in `AcreSettings.cpp`:
https://github.com/IDI-Systems/acre2/blob/e5582cd7716f620ed38d719ff283b2cf45f8d3f5/extensions/src/ACRE2Core/AcreSettings.cpp#L66

but does not look like it checks for a config change when fetching all the other configs higher up in the file. Is there a way to change `m_DisableRadioFilter`'s value other than this setting of a default? Ideally I want to do this client-side not server-side

LuckyOwl1364:
Further investigations: it seems like the main culprits for speech being unintelligible to me may be the white noise and pink noise mixed in. I see in `FilterRadio.cpp`
https://github.com/IDI-Systems/acre2/blob/e5582cd7716f620ed38d719ff283b2cf45f8d3f5/extensions/src/ACRE2Core/FilterRadio.cpp#L32
they can be toggled with the `noise` parameter, which comes from `RadioEffect.cpp`: 
[...]

PR #1383, opened 2025-02-14
General - Add setting to remove white/purple noise from TX

EmilyWinjum:
- Add "Radio FX Noise" toggle in CBA Addon Options
- Add new behavior removing pink/white noise when option is unchecked
- Remove stubbed "disableRadioFilter" config option that lacked implementation

This PR is proposed as a solution to #1329.

### DIFF
--- a/extensions/src/ACRE2Core/FilterRadio.cpp
+++ b/extensions/src/ACRE2Core/FilterRadio.cpp
@@ -27,7 +27,7 @@ acre::Result CFilterRadio::process(short* samples, int sampleCount, int channels
         }
 
         // Add noise unless setting disabled
-        if (noise && CAcreSettings::getInstance()->getDisableRadioNoise() != 1) {
+        if (noise && (CAcreSettings::getInstance()->getDisableRadioNoise() != 1)) {
             this->mixPinkNoise(floatPointer[0], sampleCount*channels, value);
             this->mixWhiteNoise(floatPointer[0], sampleCount*channels, value);
         }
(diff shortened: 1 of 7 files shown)


EmilyWinjum:
Hello, I'm not sure who to ping but I've tested my change successfully and would like to get it reviewed!

EidolonForgery:
Is this just being ignored? Two of my unit members have a lot of difficulty with the noise filtering and it's negatively impacting their gameplay experience.

jonpas (approved):
Looks good. @PabstMirror ?

PabstMirror (approved):

PabstMirror:
I guess we could make it a scalar setting from 0-1 if that would be preferred by anyone
But I think this should be good as is

extensions/src/ACRE2Core/FilterRadio.cpp
```diff
@@ -24,12 +21,13 @@ acre::Result CFilterRadio::process(short* samples, int sampleCount, int channels
...
 
-
         // Boost it
         for (int i = 0; i < sampleCount*channels; i++) {
             floatPointer[0][i] = floatPointer[0][i] * 3.0f;
         }
-        if (noise) {
+
+        // Add noise unless setting disabled
+        if (noise && CAcreSettings::getInstance()->getDisableRadioNoise() != 1) {
```
TheMagnetar:
```suggestion
        if (noise && (CAcreSettings::getInstance()->getDisableRadioNoise() != 1)) {
```

TheMagnetar (changes requested):

TheMagnetar (approved):

PR #1384, opened 2025-02-14
Intercom - Add keybinds for fast switch of FFCS Work Knob

Timi007:
- Add keybind for fast switching of working radio
- Categorize all intercom keybinds into its own subcategory

### TODO

- [x] MP Test

### Discussion

I have not found any better keybinds for this. Suggestions are welcome. It should ideally be easy to press with the left hand.


jonpas (approved):
I would prefer to see `switchWorkRadioFast`, as we are talking about the "Work" knob being turned basically.

Otherwise I have no issue with the keybind. I didn't check if it might conflict with something else important out there.

jonpas:
@Timi007 Check if my changes are fine with you.

German translations need to still be changed.

Timi007:
I've updated the translation

PR #1398, opened 2025-07-07
Extensions - Fix Longley-Rice segfault

PabstMirror:
3 Crashdumps from dedman
Different maps, but all using long rice, crashing in same spot

Untitled

in all 3 cases elev is only size 3 and `i` is going off the end

This PR should fix the overrun (and keep the average correct)
But I couldn't replicate issue locally, so I can't be positive it will fully solve root issue

### DIFF
--- a/extensions/src/ACRE2Arma/signal/models/longleyRice_itm.cpp
+++ b/extensions/src/ACRE2Arma/signal/models/longleyRice_itm.cpp
@@ -1550,7 +1550,7 @@ namespace acre {
 #endif
 
                     ja = (long) (3.0 + 0.1 * elev[0]);
-                    jb = np - ja + 6;
+                    jb = std::min<long>(np - ja + 6, np + 1);
                     for (i = ja - 1; i < jb; ++i)
                         zsys += elev[i];
                     zsys /= (jb - ja + 1);


PabstMirror:
Tested in-game with  extensions\src\ACRE2Arma\signal\models\longleyRice.cpp:75
```
    itmElevations.reserve(3);
    itmElevations.push_back(0.0);                                             // ITM: Number of points in the data - 1
    itmElevations.push_back(static_cast<float64_t>(sampleSize));           // ITM: Distance between points
    itmElevations.push_back(10.2);      // dummy
    //_map->terrain_profile(tx_pos, rx_pos, sampleSize, itmElevations);      // Get Terrain elevations
    itmElevations[0] = static_cast<float64_t>(itmElevations.size() - 1u);  // ITM: Update the number of points
```
to manually get a 3 capacity array
gave no crash and reasonable rx values

jonpas (approved):

TheMagnetar:
Hmm. While this would indeed fix the crash, this is not fixing the root problem. Why does it happen in the first place? LR is one of the must fucked up codes i have ever seen, partly due to being a port of FORTRAN. So any code modification must be tested with different combination of paths and compare it with the original implementation.

jonpas:
Replaced by #1407.

PR #1402, opened 2025-09-08
Attenuate - Add `forceSoundAttenuation` to vehicle config

OverlordZorn:
This provides an option to ignore the "disableSoundAttenuation" check per vehicle.

`disableSoundAttenuation = 1` is inherited from `CargoTurret`. This causes all FFV Seats to ignore any configured sound attenuations.

`forceSoundAttenuation` has been added as a shortcut, to avoid having to edit each individual turret manually.


OverlordZorn:
@Timi007 Hi. Sorry for the bother but i'd like to ask you regarding some acre attenuation stuff.

https://github.com/IDI-Systems/acre2/blob/master/addons/sys_attenuate/fnc_getAttenuationEffectType.sqf#L30

you're checking the vanilla disableSoundAttenuation and based of this, apply the attenuation effect or not.

Do you by chance remember why you added this check?

Timi007:
> @Timi007 Hi. Sorry for the bother but i'd like to ask you regarding some acre attenuation stuff.
> 
> https://github.com/IDI-Systems/acre2/blob/master/addons/sys_attenuate/fnc_getAttenuationEffectType.sqf#L30
> 
> you're checking the vanilla disableSoundAttenuation and based of this, apply the attenuation effect or not.
> 
> Do you by chance remember why you added this check?

I didn't add this check. It was already there in the original code. All I did is move the code into its own function and refactor it a bit. So, I don't know either why this check is there. ¯\_(ツ)_/¯

jonpas:
"ignore disable", maybe a better name would be something like `enableSoundAttenuation`or `forceSoundAttenuation`?

Documentation would be nice for this too.

jonpas (approved):

PR #1407, opened 2025-12-11
Extensions - Fix Longley-Rice segfault (alt)

PabstMirror:
Alt to #1398

doesn't touch longleyRice_itm.cpp
just make the vector big so it can't segfault


TheMagnetar (approved):
I have never been able to understand why. ITM is a FORTRAN port so yeah...

jonpas (approved):

jonpas:
@PabstMirror Undraft?

PR #1424, opened 2026-02-17
Translation - Improve Spanish 

regiregi22:
Spanish translation

Traslate all remaining english strings to spanish
- Respect the [SQF](http://acre2.idi-systems.com/wiki/development/coding-guidelines-sqf) and [C++](http://acre2.idi-systems.com/wiki/development/coding-guidelines-cpp) Coding Guidelines
- Follow title standard `System - Add|Fix|Improve|Change|Make|Remove radios`


addons/compat_spe/stringtable.xml
```diff
@@ -216,6 +239,7 @@
             <German>B-Kanäle</German>
             <Japanese>Bセット</Japanese>
             <Chinesesimp>电台B</Chinesesimp>
+            <Spanish>Equipo B</Spanish>
```
BrettMayson, line 242: All of these translations (not just the Spanish) are a little all over the place, compared to the original English

Would `Conjunto` be better on these?

TheMagnetar: Nah, not really. It is one of those words that is gramatically correct but no one would ever use

TheMagnetar (approved):

PR #1431, opened 2026-08-12
API/Various - Fix headers, match expected return value

PabstMirror:


jonpas:
PR is fine, I just want to check how we write multiple argument types in the header `<NUMBER><ARRAY>` or `<NUMBER>/<ARRAY>` or `<NUMBER> or <ARRAY>`.

jonpas:
`docs/tools/document_functions.py` needs to be updated to support `<BOOL>/<ARRAY>` syntax: https://github.com/acemod/ACE3/pull/10681

